PR #29, opened 2024-01-29
Use multi-thread for cloud experiments and mutli-process for local ones

DonggeLiu:
With clould experiments enabled, local program is left with no computationally heavy tasks. Most of the time, it only waits for cloud build results and writes them to a file.
using multi-thread can achieve higher parallelism (e.g., avoid overhead in creation, management, context switching) and require less resource (e.g., reduce memory usage).

### DIFF
--- a/run_all_experiments.py
+++ b/run_all_experiments.py
@@ -53,7 +53,7 @@ class Result:
     self.result = result
 
 
-def get_experiment_configs(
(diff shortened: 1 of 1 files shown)


DonggeLiu:
Not high priority, converting it to a draft for now and will come back to this later.

PR #46, opened 2024-01-31
Pull latest repo in exp

### LINKED ISSUE #42: Pull repo for daily experiments
Some places to change:
1. No need to [`COPY`](https://github.com/google/oss-fuzz-gen/blob/main/Dockerfile#L65)
2. Need to pull from this repo in https://github.com/google/oss-fuzz-gen/blob/main/report/docker_run.sh

cjx10:
Resolves #42

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -62,7 +62,5 @@ RUN apt-get update && \
     docker-compose-plugin
 
 
-COPY . /experiment
-WORKDIR /experiment
-RUN /venv/bin/pip install --disable-pip-version-check -r requirements.txt
-ENTRYPOINT ["./report/docker_run.sh"]
+COPY ./report/docker_run.sh /docker_run.sh
+ENTRYPOINT ["/docker_run.sh"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -62,5 +62,6 @@ RUN apt-get update && \
     docker-compose-plugin
 
 
-COPY ./report/docker_run.sh /docker_run.sh
-ENTRYPOINT ["/docker_run.sh"]
+CMD git clone --depth=1 --branch=main https://github.com/google/oss-fuzz-gen.git /experiment && \
+    cd /experiment && \
+    ./report/docker_run.sh
--- a/report/docker_run.sh
+++ b/report/docker_run.sh
@@ -19,7 +19,7 @@
 ##
 ##   benchmark_set determines the set of benchmarks used for the experiment.
 ##     The value should be the name of `benchmark-sets` directory.
-##     Default: comp_benchmarks
+##     Default: comparison
 ##   frequency_label: Examples: daily, weekly, manual, etc. This is used as part of
 ##     Cloud Build tags and GCS report directory.
 ##     Default: daily
@@ -47,7 +47,7 @@ fi
 
 if [[ $BENCHMARK_SET = '' ]]
 then
-  BENCHMARK_SET='comp_benchmarks'
+  BENCHMARK_SET='comparison'
   echo "Benchmark set was not specified as the first argument. Defaulting to ${BENCHMARK_SET:?}."
 fi
 
--- a/report/docker_run.sh
+++ b/report/docker_run.sh
@@ -26,8 +26,7 @@
 ##   run_timeout: Timeout in seconds for each fuzzing target.
 ##     Default: 300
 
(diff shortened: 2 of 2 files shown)


DonggeLiu:
Thanks, LGTM!
I will update GKE with it and test running it there this afternoon.
We can merge after that : )

Dockerfile
```diff
@@ -62,7 +62,5 @@ RUN apt-get update && \
     docker-compose-plugin
 
 
-COPY . /experiment
-WORKDIR /experiment
-RUN /venv/bin/pip install --disable-pip-version-check -r requirements.txt
-ENTRYPOINT ["./report/docker_run.sh"]
+COPY ./report/docker_run.sh /docker_run.sh
```
DonggeLiu, line 65: Since we purely rely on pulling the repo from GitHub, we won't have `./report/docker_run.sh` before cloning right? : )

cjx10: We need to put this script into the docker image so that it can run in docker before we can clone the repo when running.
This way every time we run the image it will have the latest repo (just not running the latest docker_run.sh).

DonggeLiu: How about `clone` in the Dockerfile (so that have access to the script) and `pull` in `docker_run.sh`?

DonggeLiu: Alternatively, we can have:
```
CMD git clone <this repo> && \
    cd <this repo> && \
    docker_run.sh
```

cjx10: Ah yes we can do this

DonggeLiu: Oh speaking of this, I have an additional feature request:
Print the hash value of the HEAD commit before experiment starts. For example:
```
# At the beginning of docker_run.sh:
echo "Using OSS-Fuzz-Gen version: $(git rev-parse HEAD)"
```
This helps us to pinpoint the range of commits that are responsible for improvement/regression.

DonggeLiu: Let's hold this for a bit because @erfanio is working on a more automated solution at #49

report/docker_run.sh
```diff
@@ -19,13 +19,16 @@
 ##
 ##   benchmark_set determines the set of benchmarks used for the experiment.
 ##     The value should be the name of `benchmark-sets` directory.
-##     Default: comp_benchmarks
+##     Default: comparison
```
DonggeLiu, line 22: @erfanio could you please also apply this change to your PR?
Thanks!

report/docker_run.sh
```diff
@@ -43,7 +46,7 @@ fi
 
 if [[ $BENCHMARK_SET = '' ]]
 then
-  BENCHMARK_SET='comp_benchmarks'
+  BENCHMARK_SET='comparison'
```
DonggeLiu, line 49: And this one : ) @erfanio

Dockerfile
cjx10: Sound good

Dockerfile
```diff
@@ -62,7 +62,6 @@ RUN apt-get update && \
     docker-compose-plugin
 
 
-COPY . /experiment
-WORKDIR /experiment
-RUN /venv/bin/pip install --disable-pip-version-check -r requirements.txt
-ENTRYPOINT ["./report/docker_run.sh"]
+CMD git clone --depth=1 --branch=main https://github.com/google/oss-fuzz-gen.git /experiment && \
```
erfanio, line 65: I don't think we should be going down this path tbh... Having the clone at on docker run startup is fragile and creates weird issues when we want to customize the entry point with different run timeout or experiment types.

I had a chat with Dongge about why he thinks https://github.com/google/oss-fuzz-gen/issues/42 is necessary and it comes down to making sure the images for daily experiments are built from a clean state.

I'm adding a github action to build a new image after every commit on main (https://github.com/google/oss-fuzz-gen/pull/49) and I think that's a better path to make sure the daily experiments are representative of the repo at HEAD.

report/docker_run.sh
erfanio: I created #51

report/docker_run.sh
erfanio: I created #51

Dockerfile
cjx10: Agreed, github action feels more elegant than what we proposed here. I'll close this pr.

PR #52, opened 2024-01-31
Add anon client

### LINKED ISSUE #21: Use anonymous client to retrieve AST from the public bucket
This requires replacing [`gsutil download`](https://github.com/google/oss-fuzz-gen/blob/main/data_prep/project_context/context_retriever.py#L256-L264) with code similar to [this example solution](https://github.com/google/oss-fuzz-gen/blob/b5f67787035d0c79c0ccd0481f22c6427ff3cb21/data_prep/project_targets.py#L49-L52).

Again, this requires a temporary fix before #10 is ready so that all people can access ASTs. We will switch to something different once #10 is on.

Thanks, @trashvisor : )

trashvisor:
Add anon client to download ASTs from gcs bucket.
Also some path changes (more usage of os.path).

Fixes #21

### DIFF
--- a/data_prep/project_context/context_retriever.py
+++ b/data_prep/project_context/context_retriever.py
@@ -6,6 +6,7 @@ import subprocess
 import uuid
 from collections import defaultdict
 from typing import List, Tuple
+
 from google.cloud import storage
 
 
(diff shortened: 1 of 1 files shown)


data_prep/project_context/context_retriever.py
```diff
@@ -253,17 +258,18 @@ def _get_header_from_file(self, fully_qualified_path: str) -> str:
...
-        check=True,
-        stdout=subprocess.PIPE,
-        stderr=subprocess.PIPE,
-    )
+    storage_client = storage.Client.create_anonymous_client()
+    bucket = storage_client.bucket(self.OSS_FUZZ_EXP_BUCKET)
+    project_prefix = os.path.join('project_asts', self._project_name)
+    blobs = bucket.list_blobs(prefix=project_prefix)
+    ast_dir = os.path.abspath(
+        os.path.join(os.path.dirname(__file__), '..', '..', self._ast_path))
```
DonggeLiu, line 266: nit: Consider using `os.path.dirname()` (rather than `..`) to get parent path.

DonggeLiu (approved):

PR #64, opened 2024-02-01
Restructure benchmark YAML

### LINKED ISSUE #54: Convert functions in `benchmark.yaml` into a more structured way.
For example, 
from:
```
functions:
  - int main2(int argc, char **argv)
  - int scanopt_usage(scanopt_t *scanner, FILE *fp, const char *usage)
project: flex
target_path: /src/fuzz-main.c
```

to:
```
"functions":
- "name": "main2"
  "param_names":
  - "argc"
  - "argv"
  "param_types":
  - "int"
  - "char **"
  "return_type": "int"
  "signature": "int main2(int argc, char ** argv)"
- "name": "scanopt_usage"
  "param_names":
  - "scanner"
  - "fp"
  - "usage"
  "param_types":
  - "char **"
  - "struct._IO_FILE *"
  - "char *"
  "return_type": "int"
  "signature": "int scanopt_usage(char ** scanner, struct _IO_FILE * fp, char * usage)"

project: flex
target_path: /src/fuzz-main.c

```

DonggeLiu:
Fixes #54

Done ~~Will also push the new benchmarks in this PR when they are ready.~~


DonggeLiu:
CC @trashvisor @cjx10 We now have more structured benchmarks as discussed.
Please adapt the code generation / fixing infrastructure to this.

You can base your PR on this and merge your PR to here once finished.
We will test our code together before merging into `main`.

benchmark-sets/structured/abseil-cpp.yaml
```diff
@@ -0,0 +1,69 @@
...
+  - "int"
+  - "char *"
+  "return_type": "bool"
+  "signature": "bool absl::str_format_internal::FormatArgImpl::Dispatch<long long>(char * arg, size_t spec, int out, char * )"
+- "name": "boolabsl::str_format_internal::FormatArgImpl::Dispatch<long>(absl::str_format_internal::FormatArgImpl::Data,absl::str_format_internal::FormatConversionSpecImpl,void*)"
+  "param_names":
+  - "arg"
+  - "spec"
+  - "out"
+  - ""
```
DonggeLiu, line 20: Note that some `param_name` can be empty.

DonggeLiu: Asking to avoid this in FI for better consistency.

DonggeLiu:
Close and reopen to enable building image for this PR.

benchmark-sets/structured/abseil-cpp.yaml
```diff
@@ -0,0 +1,69 @@
+"functions":
+- "name": "boolabsl::str_format_internal::FormatArgImpl::Dispatch<longlong>(absl::str_format_internal::FormatArgImpl::Data,absl::str_format_internal::FormatConversionSpecImpl,void*)"
```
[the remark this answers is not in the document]
DavidKorczynski: I think this depends a bit on what we're going for.

Is the function what one reads or what the compiler uses as the unique identifier for the function? If it's the latter, then perhaps using the mangled version is the right choice, which will include e.g. types etc.

Alternatively, is the function name:

- `absl::str_format_internal::FormatArgImpl::Dispatch<longlong>`
- `Dispatch`
- `Dispatch<longlong>`
...

?

I think likely the mangled is the right choice for uniquely identifying CPP functions, and what is used atm is the demangled version stripped from whitespace.

From a program analysis perspective, I think we will end up using the mangled version as the function name (this is what you'll see in the debugger if gdb'ing a C++ program), as this is the unique name identifier. However, it likely means we need some heuristics when queries are done based on the demangled name.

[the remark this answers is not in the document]
DavidKorczynski: For reference,

The two `rfind` functions defined here: https://github.com/abseil/abseil-cpp/blob/ddcf8be90575d494e40fcd4e0c408f0237efe0da/absl/strings/string_view.h#L485-L492 provide the following debug information:

```
{
    "name": "rfind",
    "source": {
        "source_file": "/proc/self/cwd/external/com_google_absl/absl/strings/string_view.h",
        "source_line": "485"
    },
    "args": [
        "const string_view *",
        "char",
        "size_type"
    ],
    "return_type": "size_type"
},
{
    "name": "rfind",
    "source": {
        "source_file": "/proc/self/cwd/external/com_google_absl/absl/strings/string_view.h",
        "source_line": "489"
    },
    "args": [
        "const string_view *",
        "Nonnull<const char *>",
        "size_type",
        "size_type"
    ],
    "return_type": "size_type"
},
```

Perhaps what we need is two keys per function: `fuction_identifier`, `function_name`?

benchmark-sets/structured/avahi.yaml
```diff
@@ -0,0 +1,51 @@
+"functions":
+- "name": "avahi_dns_packet_new_reply"
+  "param_names":
+  - "p"
+  - "mtu"
+  - "copy_queries"
+  - "aa"
+  "param_types":
+  - "struct.AvahiDnsPacket *"
```
trashvisor, line 9: would it make sense to break out tags from types? something like this?

```
"param_types":
- "AvahiDnsPacket *"
- "int"
- "int"
- "int"
"param_tags":
- "struct"
- ""
- ""
- ""

Comment (DonggeLiu) [reply benchmark-sets/structured/avahi.yaml:9]: Sure! I can implement a temp solution for now and request FI to separate this later.
Ideally we want to keep `introspector.py` as lightweight and transparent as possible.

File: benchmark-sets/structured/abseil-cpp.yaml [hunk 4]
```diff
@@ -0,0 +1,69 @@
...
+  "param_names":
```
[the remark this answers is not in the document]
DonggeLiu: Sure!

benchmark-sets/structured/avahi.yaml
DonggeLiu: @DavidKorczynski could you please look into this later?
I will have a temporary solution for now so no rush.
But in the long-term, I suppose we do want to avoid having tags in types because they normally do not appear in function signatures?

benchmark-sets/structured/abseil-cpp.yaml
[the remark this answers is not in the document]
DonggeLiu: Done.
Will test more and re-generate the benchmarks once FI is back online.

benchmark-sets/structured/avahi.yaml
DonggeLiu: The temp fix is done.
@trashvisor: I discarded the tag (e.g., `struct`) because I presume you don't require it. Please let me know if that can be useful. Thanks!

benchmark-sets/structured/abseil-cpp.yaml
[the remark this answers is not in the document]
DonggeLiu: Based on our Slack channel discussion, I suppose we will use the function signature from the developer's view (i.e., no `bool` before function name) and use function signature as the unique identifier?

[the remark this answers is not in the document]
DonggeLiu: Added [a temp fix](https://github.com/google/oss-fuzz-gen/pull/64/commits/5947c70eff1bd3df8934e8fd2e850523fb1adaf8#diff-2988c9b2abbf23c03fa8ed236ed0f62fb249e13ef31c1509b7fef0f6d78dd1beL201-R221) for now, but ideally, this can be handled by FI : )

experiment/benchmark.py
```diff
@@ -99,20 +106,14 @@ def from_yaml(cls, benchmark_path: str) -> List:
...
-        print(f'WARNING: Invalid benchmark config: {benchmark_path}')
-        traceback.print_exc()
-        continue
-
-      # Prevent ':' from causing issues as it propagates to other places.
-      function_name = function_name.replace('::', '-')
+    for function in functions:
       benchmarks.append(
-          cls(f'{benchmark_name}-{function_name}'.lower(),
+          cls(f'{benchmark_name}-{function.get("function_name")}'.lower(),
```
trashvisor, line 111: `function_name` -> `name` (i guess no 'function_name' under function in the yaml file (?))

DonggeLiu: Fixed, nice catche.

DonggeLiu: > i guess no 'function_name' under function in the yaml file (?)

Yeah, I used `name` instead of `function_name` to avoid stuttering.

benchmark-sets/structured/avahi.yaml
trashvisor: that should be good for now. thanks!

DavidKorczynski: Isn't `struct` part of the type?

```c
struct AvahiDnsPacket {
  int a1;
} defedAvahiDnsPacket;
```

Declares two types:

```
struct AvahiDnsPacket
defedAvahiDnsPacket
```

You'll always need to provide `struct` when creating a variable of `struct AvahiDnsPacket`?


I guess my question is what do we get from separating tag? What other tags are there, is it e.g. `const` you have in mind? If so, then there is a difference in that you can have `const struct AvahiDnsPacket` and `struct AvahiDnsPacket`, but, you can't use it in a similar manner with structs, e.g. `AvahiDnsPacket` and `struct AvahiDnsPacket` -- only the later is correct use.

> I discarded the tag (e.g., struct) because I presume you don't require it.

Not sure I understand the context, but I wouldn't have thought `AvahiDnsPacket` be used without `struct`?

DonggeLiu: I think @trashvisor was looking for a symbol name that is consistent with the one used in the signature.
In this case, the signature is [`AvahiDnsPacket* avahi_dns_packet_new_reply(AvahiDnsPacket* p, unsigned mtu, int copy_queries, int aa);`](https://github.com/avahi/avahi/blob/99eae8dd997233d3e7cff7254275d5abcaae53a9/avahi-core/dns.h#L44), which does not have the `struct` tag.

Hence from a user perspective, by looking at the function signature, the user cannot map `AvahiDnsPacket` to `struct.AvahiDnsPacket`.

DonggeLiu: > You'll always need to provide struct when creating a variable of struct AvahiDnsPacket?

Please correct me if I am wrong, but I thought we can omit the `struct`, for example:
https://github.com/avahi/avahi/blob/99eae8dd997233d3e7cff7254275d5abcaae53a9/avahi-core/dns.c#L858

Because the author used the same name:
https://github.com/avahi/avahi/blob/99eae8dd997233d3e7cff7254275d5abcaae53a9/avahi-core/dns.h#L32C1-L36C18

DonggeLiu: > So, in this case I think it's correct to omit struct from the type/tag, since it's a typedef

Could FI omit the tag in this and similar cases? It allows better consistency with the function from developers' view (e.g., the return/param types are `AvahiDnsPacket`, not `struct AvahiDnsPacket`) and can avoid confusing users/LLM.

>  in the event that there are no typedef then I think struct should be part of the type and not a tag

Would it make sense for FI to output `struct AvahiDnsPacket` (i.e., replace dot with a space) in those cases, because that's the exact words that users/LLMs will use in fuzz targets?

benchmark-sets/comparison/abseil-cpp.yaml
```diff
@@ -0,0 +1,16 @@
+"functions":
+- "name": "absl::str_format_internal::FormatArgImpl::Dispatch<long long>(absl::str_format_internal::FormatArgImpl::Data, absl::str_format_internal::FormatConversionSpecImpl, void*)"
```
cjx10, line 2: Should params be removed from name?
```suggestion
- "name": "absl::str_format_internal::FormatArgImpl::Dispatch<long long>"
```

DonggeLiu: Not for now if we want to omit the return type in names.
Please correct me if I am wrong: In C++, two functions can have the same name (i.e., `Dispatch `) and template specialization (i.e., `<long long>`) but different return types and params.
Happy to omit the params if I am wrong, otherwise we have to have either the return types or params.

BTW, may I ask why you would prefer to omit the parameters?

cjx10: Assuming we don't intend to use function name as the unique identifier, we already have separated params info below, and params seems unrelated to the name of the function. Function signature can be used to identify a specific overloaded function, or by reconstructing `"params"`.
Since function params are removed in some projects (e.g. [avahi](https://github.com/google/oss-fuzz-gen/pull/64/files#diff-80970d1beea041fc1ab131007204780451bf90ede303cc4fc2c2dad89e7c8568)), it'd be nice to keep with a same format (with or without params)?

DonggeLiu: 1. Yes, you are right: we will use signature as the unique identifier in APIs later.
2. However, the function name also needs to be unique for now because they will be used as [workdir names](https://github.com/google/oss-fuzz-gen/blob/1a0c8323b3b00701ca0910ea749fecc8857fb791/run_all_experiments.py#L81).
3. Later, we can resolve this by separating three things:
  a) benchmark id, 
  b) benchmark directory name (which needs to be unique), and 
  c) benchmark docker image (which needs to be unique and should not contain special chars). 
  But I reckon this is not related to the primary intention of this PR (making the benchmark more structured) or our priority. So far, we only have ugly names in the report, and need a bit parsing when checking if function is in fuzz target (Please do let me know if [this regex](https://github.com/google/oss-fuzz-gen/pull/64/commits/d813d36d38d479a9e291041f1f34de623010bcec#diff-2988c9b2abbf23c03fa8ed236ed0f62fb249e13ef31c1509b7fef0f6d78dd1beR127) does not work).
4. Yes, params are removed from some projects. That's mainly because they are C programs, which do not allow overloading.

DonggeLiu:
@oliverchang  This PR is ready to merge.
If you are happy with [this report](https://llm-exp.oss-fuzz.com/Result-reports/ofg-pr/2024-02-11-64-dg-comparison/sort.html), could you please approve this PR? I need at least one approve to merge it : )

Follow-up work are:
1. Reduce the size of the `comparison` benchmark set to make the experiment quicker (#90).
2. Capture more corner cases in function name parsing and fix them (#89).

They are lower priorities for now and can be done later.

benchmark-sets/comparison/abseil-cpp.yaml
DonggeLiu: I got some time to extend this PR a bit more based on your suggestions:
1. Beautify benchmark names. Now benchmark directory name, which is still `becnhmark.id` uses `<return_type><function_name>< Template Specialization>`. This removes params, making the directory shorter and more readable while maintaining uniqueness.
2. Separate the docker image name from `benchmark.id` by [removing the special chars](https://github.com/google/oss-fuzz-gen/pull/64/commits/770cafdf133df5b7adc6c3847326341d345ea097) in them.
3. Ensure special chars are encoded in the HTML report so that we can download them via `wget2` and visit them.

Here is the [final view](https://github.com/google/oss-fuzz-gen/pull/64/commits/770cafdf133df5b7adc6c3847326341d345ea097).

DonggeLiu: I might use the function signature as benchmark id later for simplicity, as we will use that as the unique identifier in FI anyway.

benchmark-sets/comparison/abseil-cpp.yaml
```diff
@@ -0,0 +1,65 @@
+"functions":
+- "name": "bool absl::str_format_internal::FormatArgImpl::Dispatch<long long>(absl::str_format_internal::FormatArgImpl::Data, absl::str_format_internal::FormatConversionSpecImpl, void*)"
```
oliverchang, line 2: What's the difference between name and signature now? Should we just make this all simpler and have just a single field?

DonggeLiu: Yes, I created #92 for this.

I prefer to replace all function name usages with signatures in another PR, given this mainly focuses on adding more details to benchmark YAML.
Would this be reasonable or would you prefer me to fix that in this PR?

oliverchang: SGTM. please add more text/details to the bug to better explain the context :)

oliverchang (approved):

benchmark-sets/comparison/abseil-cpp.yaml
DonggeLiu: Thanks, done!

PR #73, opened 2024-02-02
Fix web

DonggeLiu:
Fix[ error](https://pantheon.corp.google.com/logs/query;query=resource.type%3D%22k8s_container%22%0Aresource.labels.project_id%3D%22oss-fuzz%22%0Aresource.labels.location%3D%22us-central1%22%0Aresource.labels.cluster_name%3D%22llm-experiment%22%0Aresource.labels.namespace_name%3D%22default%22%0Aresource.labels.pod_name%3D%22daily-manual-6-qgv8f%22%20severity%3E%3DDEFAULT;cursorTimestamp=2024-02-02T07:59:28.341953276Z;aroundTime=2024-02-02T07:58:31.072Z;duration=PT10M?e=-13802955&mods=logs_tg_prod&project=oss-fuzz) and linting in `web.py`.

### DIFF
--- a/report/web.py
+++ b/report/web.py
@@ -249,7 +249,7 @@ def benchmark(bnmk):
                          prompt=get_prompt(bnmk))
 
 
-@app.route('/sample/<benchmark>/<sample>')
+@app.route('/sample/<bnmk>/<smp>')
 def sample(bnmk, smp):
   return render_template('sample.html',
                          bnmk=bnmk,
--- a/report/web.py
+++ b/report/web.py
@@ -108,6 +108,20 @@ def get_generated_targets(bnmk: str) -> list[str]:
   return targets
 
 
+def _is_valid_benchmark_dir(cur_dir: str) -> bool:
+  """Checks if |cur_dir| is a valid benchmark directory (e.g., no lost+found)"""
+  # Check prefix.
+  if not cur_dir.startswith('output-'):
+    return False
+  # Check sub-directories.
+  expected_dirs = ['raw_targets', 'status', 'fixed_targets', 'logs', 'corpora']
+  if not all(
+      os.path.isdir(os.path.join(RESULTS_DIR, cur_dir, expected_dir))
+      for expected_dir in expected_dirs):
+    return False
+  return True
+
+
 def list_benchmarks() -> List[Benchmark]:
   """Lists benchmarks in the result directory."""
   benchmarks = []
@@ -116,9 +130,9 @@ def list_benchmarks() -> List[Benchmark]:
   # failures in get_generated_targets().
   # Maybe it is from mounting?
   # TODO(erfan): Check if not mounting to the root dir can solve this.
-  benchmark_names = sorted(
-      [dir for dir in os.listdir(RESULTS_DIR) if dir != 'lost+found'])
   for bnmk in benchmark_names:
+    if not _is_valid_benchmark_dir(bnmk):
+      continue
     results, targets = get_results(bnmk)
     status = 'Done' if all(r for r in results) and results else 'Running'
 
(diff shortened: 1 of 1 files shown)


report/web.py
```diff
@@ -106,10 +109,17 @@ def get_generated_targets(benchmark: str) -> list[str]:
...
   benchmarks = []
   benchmark_names = sorted(os.listdir(RESULTS_DIR))
-  for benchmark in benchmark_names:
-    results, targets = get_results(benchmark)
+  # Not sure why there is a `lost+found` dir in |RESULTS_DIR|, which caused
+  # failures in get_generated_targets().
+  # Maybe it is from mounting?
+  # TODO(erfan): Check if not mounting to the root dir can solve this.
+  benchmark_names = sorted(
+      [dir for dir in os.listdir(RESULTS_DIR) if dir != 'lost+found'])
```
DonggeLiu, line 120: @erfanio : I started to observe this error after mounting `results-volume`.
Could we avoid this error by mounting a different directory (or other ways)?
Thanks :)

report/web.py
```diff
@@ -85,8 +88,8 @@ def get_results(benchmark) -> tuple[list[evaluator.Result], list[str]]:
   return results, targets
 
 
-def get_prompt(benchmark) -> Optional[str]:
-  root_dir = os.path.join(RESULTS_DIR, benchmark)
+def get_prompt(bnmk) -> Optional[str]:
```
oliverchang, line 91: Why is this rename required?

report/web.py
oliverchang: Looks like this is usually due to some disk corruption. 

Is there some other way we can filter this out, without resorting to hardcoding this directory name? 

e.g. ensuring that the directory contains some file/structure that matches what we expect from a benchmark, or just continuing when we hit an exception below (with a log).

report/web.py
DonggeLiu: To avoid linting warnings:
```
Redefining name 'benchmark' from outer scope
```
This is becase we have a function named `benchmark`:
https://github.com/google/oss-fuzz-gen/blob/main/report/web.py#L232

report/web.py
DonggeLiu: > just continuing when we hit an exception below (with a log)

Yep, I can do that!

> ensuring that the directory contains some file/structure that matches what we expect from a benchmark

I am not quite sure what the content of `lost+found` is; hence, I cannot tell what must be in a benchmark dir and not in it.

But continuing with a log is a quick fix I can make.

report/web.py
oliverchang: Can we instead rename that function? `bnmk` is a bit inconsistent with all of our other naming.

DonggeLiu: OK!

DonggeLiu: I will do the same with `sample`

report/web.py
DonggeLiu: Fixed by adding an overall validation, because there are too many contents in the benchmark dir to check individually.

report/web.py
DonggeLiu: Fixed

report/web.py
```diff
@@ -105,10 +108,31 @@ def get_generated_targets(benchmark: str) -> list[str]:
...
 
 
+def _is_valid_benchmark_dir(cur_dir: str) -> bool:
+  """Checks if |cur_dir| is a valid benchmark directory (e.g., no lost+found)"""
+  # Check prefix.
+  if not cur_dir.startswith('output-'):
+    return False
+  # Check sub-directories.
+  expected_dirs = ['raw_targets', 'status', 'fixed_targets', 'logs', 'corpora']
+  if not all(
```
oliverchang, line 118: Can just do `return all(...)`.

oliverchang (approved):

report/web.py
DonggeLiu: Thanks! Fixed in #83 : )

PR #78, opened 2024-02-05
Use sub-directories to categorize reports

DonggeLiu:
As we start to have many reports (from daily, one-off, pr), I wanted to use sub-directory as an easy way to visually categorize reports to keep the directory more organized.

@erfanio also proposed using tags in the report, which will provide more dimensionality in classification later.

### DIFF
--- a/report/docker_run.sh
+++ b/report/docker_run.sh
@@ -29,6 +29,7 @@
 BENCHMARK_SET=$1
 FREQUENCY_LABEL=$2
 RUN_TIMEOUT=$3
+SUB_DIR=$4
 # Uses python3 by default and /venv/bin/python3 for Docker containers.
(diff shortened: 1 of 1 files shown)


report/docker_run.sh
```diff
@@ -59,17 +60,23 @@ then
   echo "Run timeout was not specified as the third argument. Defaulting to ${RUN_TIMEOUT:?}."
 fi
 
+# The subdirectory for the generated report in GCS.
+if [[ $SUB_DIR = '' ]]
+then
+  SUB_DIR='others'
```
oliverchang, line 66: How about "default" instead? "others" seems a bit potentially confusing.

DonggeLiu: Sure!

DonggeLiu: Fixed

oliverchang (approved):

PR #86, opened 2024-02-09
Header for cpp

trashvisor:
Add support to retrieve header files for c++ projects

### DIFF
--- a/data_prep/project_context/context_retriever.py
+++ b/data_prep/project_context/context_retriever.py
@@ -9,6 +9,8 @@ from typing import List, Tuple
 
 from google.cloud import storage
 
+from experiment import benchmark as benchmarklib
+
 
 class ContextRetriever:
   """ContextRetriever attempts to retrieve context for a certain project/function from ASTs."""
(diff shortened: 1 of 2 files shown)


trashvisor:
doesnt handle templates

data_prep/project_context/context_retriever.py
```diff
@@ -231,28 +228,76 @@ def _get_dequal_type(self, fully_qualified_type: str) -> str:
...
+      if current_depth == len(self._namespace_tokens):
+        if node.get('name') == self._function_name:
+          return True
+
+      if node.get('kind') == 'NamespaceDecl' and current_depth < len(
+          self._namespace_tokens):
+        if not self._get_namespace_node(node, current_depth):
+          continue
+        else:
+          return True
```
DonggeLiu, line 248: nit: I suppose this is:
```python
        if (node.get('kind') == 'NamespaceDecl' and current_depth < len(self._namespace_tokens) and self._get_namespace_node(node, current_depth)):
          return True
```
Using extra `()` here because the conditions are too long and need linebreaks.
Autoformat **should** be able to add the linebreaks for you.

data_prep/project_context/context_retriever.py
```diff
@@ -231,28 +228,76 @@ def _get_dequal_type(self, fully_qualified_type: str) -> str:
...
+
+    return False
+
+  def _search_backwards_for_header(self, search_index: int, ast_nodes) -> str:
+    """Iterates backwards over a node to get the file location.
+    This is required because the AST stores the file location only on the
+    top most node of the file which can contain a loc object (?)."""
+    current_index = search_index
```
DonggeLiu, line 256: nit: Do we need to declare `current_index`?

DonggeLiu (approved):
Some nits.
Feel free to ignore them if that takes too long for you to fix this time: We will replace AST parsing with API shortly.

DonggeLiu:
Ops there are conflicts.
Is this important to add, or will be replaced by APIs next weeek?

trashvisor:
should mostly be replaced by APIs, i think keeping headers around somewhere is worthwhile (though im not sure if thats possible via debug information, see the chat with mihai) as there are situations where embedding type definitions might be too explosive(?) if there are conflicts i guess it doesnt make sense to merge since we are looking to deprecate this.

DonggeLiu:
> should mostly be replaced by APIs, i think keeping headers around somewhere is worthwhile (though im not sure if thats possible via debug information, see the chat with mihai) as there are situations where embedding type definitions might be too explosive(?) if there are conflicts i guess it doesnt make sense to merge since we are looking to deprecate this.

Thanks for the explanation.
I do expect that info from FI (generated from debug or other methods).
It's not a bad idea to have the PR anyway, so that we have a permanent copy of this part of code, in case we need to use it on some projects later.
If resolving the conflict is too much effort, no problem dropping this for now.

trashvisor:
> > should mostly be replaced by APIs, i think keeping headers around somewhere is worthwhile (though im not sure if thats possible via debug information, see the chat with mihai) as there are situations where embedding type definitions might be too explosive(?) if there are conflicts i guess it doesnt make sense to merge since we are looking to deprecate this.
> 
> Thanks for the explanation. I do expect that info from FI (generated from debug or other methods). It's not a bad idea to have the PR anyway, so that we have a permanent copy of this part of code, in case we need to use it on some projects later. If resolving the conflict is too much effort, no problem dropping this for now.

ACK, actually the delta is about 5 lines, so resolving the merge conflict should not be a problem

trashvisor:
i think ill wait until #64 is finished as changes are still iterative.

data_prep/project_context/context_retriever.py
```diff
@@ -240,8 +240,10 @@ def _get_namespace_node(self, namespace_node, current_depth: int) -> bool:
         if node.get('name') == self._function_name:
           return True
 
-      if node.get('kind') == 'NamespaceDecl' and current_depth < len(self._namespace_tokens) and self._get_namespace_node(node, current_depth):
-          return True
+      if node.get('kind') == 'NamespaceDecl' and current_depth < len(
+          self._namespace_tokens) and self._get_namespace_node(
+              node, current_depth):
```
DonggeLiu, line 245: Add `()` around conditions because we cannot fit them in one line. For example:
https://github.com/google/oss-fuzz-gen/pull/86#discussion_r1483929208

DonggeLiu:
#64 is merged, hope it won't block you anymore : )

But given we will start using FI APIs this week, feel free to drop this if you see fit.

DonggeLiu:
Closing this PR as we will use FI APIs now.

PR #87, opened 2024-02-09
two minor changes for running all experiments

sallywang147:
Hi,

PR for two minor changes: 

- removing `| str` at line 47 in run_all_experiments.py, because of the compilation error below: 
Screen Shot 2024-02-09 at 2 39 24 PM

-cleaning the benchmark yaml file by removing the first string, because of the error below:
Screen Shot 2024-02-09 at 2 47 10 PM

 
Users can run the added file: `python clean.py` at the root of the repository and retrieve all experiments without the issue above. 

Happy to refactor any of the submitted code if/when needed.

### DIFF
--- a/benchmark-sets/all/brpc.yaml
+++ b/benchmark-sets/all/brpc.yaml
@@ -1,4 +1,3 @@
-Retrieving human-written fuzz targets of brpc from local Docker build.
 functions:
 - int brpc::policy::ConsulNamingService::RunNamingService(brpc::policy::ConsulNamingService
   * , char * , brpc::ProgressiveReader * )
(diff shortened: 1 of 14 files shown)


DonggeLiu:
Happy new year, @sallywang147! Thanks for the PR!

> removing | str at line 47 in run_all_experiments.py, because of the compilation error below:

This appears to be caused by different `Python` versions:
[PEP 604 – Allow writing union types as X | Y](https://peps.python.org/pep-0604/) was introduced in Python 3.10, hence the error from your `Python 3.9`.

We recommend [using `Python3.11` for this repo](https://github.com/google/oss-fuzz-gen/blob/main/USAGE.md), which is the same version we used for development : )

> [delete top string](https://github.com/google/oss-fuzz-gen/pull/87/commits/08c73b49e1c71e833aad128e98af708652be5494)

Nice catch : )
I am updating the benchmarks in this #64, which will replace the `benchmark-sets/all` once ready.
I will double-check and ensure this won't happen in the new version.

Thanks again!

sallywang147:
Ah I see. Thank you Dongge about the python version suggestion!

PR #91, opened 2024-02-11
Using function signature as unique function identifier

DonggeLiu:
Testing using function signature to replace function name.
Given FI APIs will use a function signature as the unique identifier; there is no need to use the function name in the benchmark yaml, and it is more consistent to for us to uniquely identify function with their signatures.

I've handled special chars in benchmark signatures when generating reports (web.py) and building docker images in the base PR, so this should work directly.

This also fixed #91.


DonggeLiu:
Result:
https://llm-exp.oss-fuzz.com/Result-reports/ofg-pr/2024-02-12-91-dg-comparison/sort.html

oliverchang:
Can you update the PR title  to be more descriptive?

DonggeLiu:
> Can you update the PR title to be more descriptive?

Done

DonggeLiu:
TBR.

PR #94, opened 2024-02-12
Catch the error in preparing OSS-Fuzz for benchmark generation.

DonggeLiu:
Seeing this error occurred on some projects when generating benchmarks.
Log the project name for this error in case this happens again so that we can re-generate the benchmark.yaml for that project.


cjx10 (approved):
lgtm

PR #96, opened 2024-02-12
Hide the error message when deleting temp OSS-Fuzz dir

DonggeLiu:
Hiding the error message details because:
1. The error is expected in all local experiments: OSS-Fuzz dir (`/tmp/<tmp_dir>/build/work`) contains the files copied from the docker container, which requires root privilege to delete.
2. The error is benign: Temp dir is in `/tmp` and will be automatically deleted by OS.
3. The error is hard to fix: Fixing this requires using `sudo` or docker, which adds too much unnecessary effort.

We did not have this problem before because we used `pass`:
https://github.com/google/oss-fuzz-gen/commit/9885ae16b9f12686b504eecd94b35641f27c608a#diff-c265089165acff162907dbef5c22b1f966e638b1e750bee28c5a548262464fc3L40

Now, we also log a warning in case this happens elsewhere.

### DIFF
--- a/experiment/oss_fuzz_checkout.py
+++ b/experiment/oss_fuzz_checkout.py
@@ -41,7 +41,7 @@ def _remove_temp_oss_fuzz_repo():
   try:
     shutil.rmtree(OSS_FUZZ_DIR)
   except PermissionError as e:
-    logging.warning('No permission to remove %s: %s', OSS_FUZZ_DIR, e)
+    logging.warning('No permission to remove %s', OSS_FUZZ_DIR)
   except FileNotFoundError as e:
     logging.warning('No OSS-Fuzz directory %s: %s', OSS_FUZZ_DIR, e)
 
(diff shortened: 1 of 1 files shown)


jonathanmetzman (approved):
I'm not sure why it's worth removing the exception but this change dosn't seem to break anything to me.

DonggeLiu:
> I'm not sure why it's worth removing the exception but this change dosn't seem to break anything to me.

Sorry, I should provide a bit more context:
Previously, we held the temp dir of OSS-Fuzz in a local variable of a function. Hence, the dir may be recycled during the experiment. The good thing is that the long and annoying error message will not pop at the end of experiments, where we print out the results. The bad thing is that some trials failed to find the OSS-Fuzz dir during the experiment and break.

I fixed it by [holding it in a global variable](https://github.com/google/oss-fuzz-gen/pull/64/commits/89a45986c02967191be0742b4abfd885e8f3c981). This makes the temp last till the end, but the long error messages will also show up at the end, which becomes annoying.

I should probably also remove the other `e`, for the same reason.

PR #97, opened 2024-02-12
Add raw function name to benchmark YAML

DonggeLiu:
Add `raw_function_name` into benchmark.yaml, because they will be used as the unique function identifier.
Use `raw_function_name` as the benchmark's local directory name for their succinctness and uniqueness.
Replace other usages of `function_name` with `function_signature`.


DonggeLiu:
JOB: https://pantheon.corp.google.com/kubernetes/job/us-central1/llm-experiment/default/ofg-pr-97-dg
REPORT: https://llm-exp.oss-fuzz.com/Result-reports/ofg-pr/2024-02-13-97-dg-comparison/index.html

oliverchang (approved):

DonggeLiu:
Some recent changes (from us or GKE) seem to increase GKE memory consumption and break uploading reports.
Fixed in the other repo.

PR #98, opened 2024-02-12
Truncate ids to fit within max OS file name length

trashvisor:
C++ projects in particular with longer names in benchmark.yaml cause an error to be raised in workdir generation due to long file names


trashvisor:
JOB: https://pantheon.corp.google.com/kubernetes/job/us-central1/llm-experiment/default/ofg-pr-98-ggryan
REPORT: https://llm-exp.oss-fuzz.com/Result-reports/ofg-pr/2024-02-12-98-ggryan-comparison/index.html

experiment/benchmark.py
```diff
@@ -73,16 +73,18 @@ def from_yaml(cls, benchmark_path: str) -> List:
...
-    benchmark_name = os.path.splitext(os.path.basename(benchmark_path))[0]
+    project_name = os.path.splitext(os.path.basename(benchmark_path))[0]
 
     use_context = data.get('use_context', False)
     use_project_examples = data.get('use_project_examples', True)
     cppify_headers = data.get('cppify_headers', False)
     commit = data.get('commit')
     functions = data.get('functions', [])
     for function in functions:
+      max_len = os.pathconf('/', 'PC_NAME_MAX') - len('output-')
```
oliverchang, line 84: Can you add a comment explaining that function signatures can be very long and exceed filesystem limits?

oliverchang (approved):
Thanks @trashvisor !

@DonggeLiu we might want to see if we can come up with some better way to generate this, in case we introduce clashes by truncating.

experiment/benchmark.py
trashvisor: added, thanks!

DonggeLiu:
> @DonggeLiu we might want to see if we can come up with some better way to generate this, in case we introduce clashes by truncating.

Yep, this is a temp solution.

@trashvisor and I discussed using `raw_function_name` as the directory name.
However, our current directory name is tied with the name in [the report](https://llm-exp.oss-fuzz.com/Result-reports/ofg-pr/2024-02-12-93-dg-comparison/index.html), and having benchmark name like `"_ZN4absl19str_format_internal13FormatArgImpl8DispatchIxEEbNS1_4DataENS0_24FormatConversionSpecImplEPv"` is not user friendly.

We can mitigate this by matching `raw_function_name` with the function signature during report generation (e.g., in `web.py`), but that takes a bit more extra work and can wait until this finishes.

DonggeLiu:
> JOB: https://pantheon.corp.google.com/kubernetes/job/us-central1/llm-experiment/default/ofg-pr-98-ggryan REPORT: https://llm-exp.oss-fuzz.com/Result-reports/ofg-pr/2024-02-12-98-ggryan-comparison/index.html

Killing this as its code is outdated : )

DonggeLiu:
Seeing this error:
https://pantheon.corp.google.com/logs/query;cursorTimestamp=2024-02-12T13:20:40.335827856Z;duration=PT12H;query=resource.type%3D%22k8s_container%22%0Aresource.labels.project_id%3D%22oss-fuzz%22%0Aresource.labels.location%3D%22us-central1%22%0Aresource.labels.cluster_name%3D%22llm-experiment%22%0Aresource.labels.namespace_name%3D%22default%22%0Alabels.k8s-pod%2Fbatch_kubernetes_io%2Fcontroller-uid%3D%22f475de05-eccc-4e51-8a9b-a15efc705afa%22%20severity%3E%3DDEFAULT%0Atimestamp%3D%222024-02-12T13:20:40.335827856Z%22%0AinsertId%3D%22egk006exzw7vw80x%22?project=oss-fuzz

No concrete proof yet, but since it happened after the 'Uploading the report.', I guess it relates to uploading directories with truncated name.

trashvisor:
double pointers sorry, see https://cloud.google.com/storage/docs/wildcards

```
Match zero or more characters across directory boundaries. When used as part of a local file path, the ** wildcard should always be immediately preceded by a directory delimiter. For example, my-directory/**.txt is valid, but my-directory/abc** is not.
```

DonggeLiu:
> double pointers sorry, see https://cloud.google.com/storage/docs/wildcards
> 
> ```
> Match zero or more characters across directory boundaries. When used as part of a local file path, the ** wildcard should always be immediately preceded by a directory delimiter. For example, my-directory/**.txt is valid, but my-directory/abc** is not.
> ```

Yes, it could be.
As a temp mitigation, I will switch to `raw_function_name`, given we have that info now.
Let me test the change in my PR.

DonggeLiu:
Would you mind if I steal your code to mine?
I want to merge, but copying is probably simpler than resolving conflicts in this case : (

trashvisor:
haha go for it!!

DonggeLiu:
Manually 'merged' into #97

PR #110, opened 2024-02-16
Fix docker tag pattern

DonggeLiu:
Docker appears to have an undocumented name rule that disallows dashes (`-`) followed by underscores (`_`) (and `-`s after `_`), which caused some of our projects to fail to build.

This PR fixes this issue by replacing these patterns (`[_-]{2, }`) with `-`.


DonggeLiu:
Replaced by #111

PR #121, opened 2024-02-22
Fix priming for fixing on OpenAI models.

mihaimaruseac:
There was a confusion between the priming text and the priming prompt. It is not a problem for text prompts where the role is ignored, but for structured prompts (OpenAI), we got into a state where the structured prompt was a nested structure:

```json
{
  "role": "system",
  "content": [
    {
      "role": "system",
      "content": "Given the following C++ program and its build error message, fix the code to achieve the original functionality.\n\nIf a function is missing, fix it by including the related libraries.\n\nIf the error contains \"undefined reference to\", this means one or more of the header <code>#include</code> directives need to be wrapped with <code>extern \"C\" {}</code> for the object linking to succeed. You must fix this by rewriting the relevant <code>#include</code> and wrapping it with <code>extern \"C\" {}</code>.\n"
    }
  ]
}
```

### DIFF
--- a/llm_toolkit/prompt_builder.py
+++ b/llm_toolkit/prompt_builder.py
@@ -239,19 +239,21 @@ class DefaultTemplateBuilder(PromptBuilder):
   def build_fixer_prompt(self, benchmark: Benchmark, raw_code: str,
                          errors: list[str]) -> prompts.Prompt:
     """Prepares the code-fixing prompt."""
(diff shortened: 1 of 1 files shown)


DonggeLiu (approved):

DonggeLiu:
Please ignore the the Build docker image CI, it is trivial for this PR.

PR #100, opened 2024-02-13
retry on gcloud crashed (OSError)

cjx10:
Fixes #85

PR #124, opened 2024-02-27
Capture `RESOURCE_EXHAUSTED` error and retry with jitter

cjx10:
Mentioned in pr #100


experiment/builder_runner.py
```diff
@@ -328,7 +328,9 @@ class CloudBuilderRunner(BuilderRunner):
       # Temp workaround for issue #12.
       'You do not currently have an active account selected',
       # Workaround for issue #85.
-      'gcloud crashed (OSError): unexpected end of data'
+      'gcloud crashed (OSError): unexpected end of data',
+      # As mentioned in pr #100.
+      'RESOURCE_EXHAUSTED'
```
DonggeLiu, line 333: nit: Append a comma (i.e., `'RESOURCE_EXHAUSTED',`) so that we don't have to change this line if later we find more errors to handle.

experiment/builder_runner.py
```diff
@@ -376,12 +378,19 @@ def build_and_run(self, generated_project: str, target_path: str,
...
         break
       except sp.CalledProcessError as e:
-        # Remove \n for single log entry on cloud.
+        # Replace \n for single log entry on cloud.
         stdout = e.stdout.decode('utf-8').replace('\n', '\t')
         stderr = e.stderr.decode('utf-8').replace('\n', '\t')
-        if (any(error in stdout + stderr for error in self._RETRYABLE_ERRORS)
-            and attempt_id < CLOUD_EXP_MAX_ATTEMPT):
+
+        captured_error = next(
```
DonggeLiu, line 385: A silly question: Why do we use `next` instead of `any`?

experiment/builder_runner.py
```diff
@@ -376,12 +378,19 @@ def build_and_run(self, generated_project: str, target_path: str,
...
+            (err for err in self._RETRYABLE_ERRORS if err in stdout + stderr),
+            '')
+        if captured_error and attempt_id < CLOUD_EXP_MAX_ATTEMPT:
           delay = 5 * 2**attempt_id
+          if captured_error == 'RESOURCE_EXHAUSTED':
+            # Add random jitter in case of exceeding request per minute quota.
+            delay += random.randint(1, 10)
```
DonggeLiu, line 392: Given the quota is `per minute`, would it be better to wait at least 60 secs?

experiment/builder_runner.py
cjx10: Previously we just need to check if any error in the list presents, so using `any` was fine. 
Now we also want to check for one specific case to add jitter, hence changing to `next` to get which error we are catching.
If we want to capture more errors with different backoff time in the future, would probably just refactor this in a separate function to simplify the code.

experiment/builder_runner.py
cjx10: If long waiting time is acceptable in this case. The rate is calculated in a 60 sec moving window per [Compute Engine rate quotas](https://cloud.google.com/compute/api-quota#:~:text=Google%20enforces%20rate%20quotas%20at%20a%20per%2Dminute%20(60%20seconds)%20interval%20for%20each%20group.).

experiment/builder_runner.py
DonggeLiu: > Now we also want to check for one specific case to add jitter, hence changing to next to get which error we are catching.

Should we iterate through items instead of only getting the first one via `next`?

> If we want to capture more errors with different backoff time in the future, would probably just refactor this in a separate function to simplify the code.

Yes, that's better.

experiment/builder_runner.py
DonggeLiu: Yes, waiting a few minutes is acceptable because this error happens when we exceed the quota limit, which only happens in large-scale experiments.

experiment/builder_runner.py
cjx10: > Should we iterate through items instead of only getting the first one via `next`?

I think getting the first one is sufficient, but probably better to check the one with the longest backoff first

experiment/builder_runner.py
DonggeLiu: > The rate is calculated in a 60 sec moving window per [Compute Engine rate quotas](https://cloud.google.com/compute/api-quota#:~:text=Google%20enforces%20rate%20quotas%20at%20a%20per%2Dminute%20(60%20seconds)%20interval%20for%20each%20group.).

IIUC, given `attempt_id` won't be more than 5, the current delay will be at most 2.83 minutes (=5*2**5+10 seconds).
In weekly experiments with the `all` benchmark set, we have ~1500 benchmarks * 10 fuzz targets per benchmark.
More than 3 minutes is likely required for this experiment size.

cjx10:
PR: https://github.com/google/oss-fuzz-gen/pull/124
JOB: https://pantheon.corp.google.com/kubernetes/job/us-central1/llm-experiment-large/default/ofg-pr-124-dg-large
REPORT: https://llm-exp.oss-fuzz.com/Result-reports/ofg-pr/2024-02-27-124-dg-large-comparison/index.html
BUCKET: https://pantheon.corp.google.com/storage/browser/oss-fuzz-gcb-experiment-run-logs/Result-reports/ofg-pr/2024-02-27-124-dg-large-comparison
BUCKET GS: `gs://oss-fuzz-gcb-experiment-run-logs/Result-reports/ofg-pr/2024-02-27-124-dg-large-comparison`

cjx10:
JOB: https://pantheon.corp.google.com/kubernetes/job/us-central1/llm-experiment-large/default/ofg-pr-124-dg-large
REPORT: https://llm-exp.oss-fuzz.com/Result-reports/ofg-pr/2024-02-27-124-dg-large-all/index.html
BUCKET: https://pantheon.corp.google.com/storage/browser/oss-fuzz-gcb-experiment-run-logs/Result-reports/ofg-pr/2024-02-27-124-dg-large-all
BUCKET GS: `gs://oss-fuzz-gcb-experiment-run-logs/Result-reports/ofg-pr/2024-02-27-124-dg-large-all`

data_prep/introspector.py
```diff
@@ -42,7 +42,11 @@
 def _query_introspector(api: str, params: dict) -> dict:
   """Queries FuzzIntrospector API and return data specified by |key|,
   returns None if unable to get the value."""
-  resp = requests.get(api, params, timeout=TIMEOUT)
+  try:
```
DonggeLiu, line 45: Thanks for addressing this so quickly, @cjx10!

Just a minor tip: It would be nicer to have this in a different PR, though : ) This is different from the `RESOURCE_EXHAUSTED` error in the title.
But no need to change it this time.

data_prep/introspector.py
```diff
@@ -42,7 +42,11 @@
...
+    resp = requests.get(api, params, timeout=TIMEOUT)
+  except requests.exceptions.Timeout as err:
+    logging.error('Failed to get data from FI due to timeout\n%s', err)
+    return {}
```
DonggeLiu, line 49: 1. It would be better to use the 'delay and retry' trick again here, just in case this is caused by too frequent API access.
2. Could you please log more details about the request to help FI debug. For example, `api`, `params`.

cjx10: Done. Moving this change to new pr #128 as it's more than a few lines

DonggeLiu (approved):

cjx10:
JOB: https://pantheon.corp.google.com/kubernetes/job/us-central1/llm-experiment/default/ofg-pr-124-jim
REPORT: https://llm-exp.oss-fuzz.com/Result-reports/ofg-pr/2024-02-29-124-jim-comparison/index.html
BUCKET: https://pantheon.corp.google.com/storage/browser/oss-fuzz-gcb-experiment-run-logs/Result-reports/ofg-pr/2024-02-29-124-jim-comparison
BUCKET GS: `gs://oss-fuzz-gcb-experiment-run-logs/Result-reports/ofg-pr/2024-02-29-124-jim-comparison`

PR #136, opened 2024-03-05
add helper script to search string in result

### LINKED ISSUE #45: A helper script to find all fuzz targets that share the same error in one experiment

cjx10:
Resolve #45

### DIFF
--- a/helper/result_string_search.py
+++ b/helper/result_string_search.py
@@ -13,7 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 """
-Search for the benchmarks that contains a specific string in result.
+Search for benchmark results that contain a specific string.
 
 Example usage:
 To search for benchmarks that failed to parse error:
@@ -30,8 +30,8 @@ import logging
 import os
 
 
-def parse_args() -> argparse.Namespace:
-  """parse arguments"""
+def _parse_args() -> argparse.Namespace:
+  """Parses arguments."""
   parser = argparse.ArgumentParser(
       description=
       'Search for all benchmark that contains the <string> in <result>')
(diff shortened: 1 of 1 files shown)


helper/result_string_search.py
```diff
@@ -0,0 +1,137 @@
...
+      '-u',
+      '--url',
+      type=str,
+      default='',
+      help='Optional url template to web report for easy access.')
+
+  args = parser.parse_args()
+  assert os.path.isdir(args.result), '--result must be an existing directory.'
+
+  output_dir = next(os.walk(args.result))[1][0]
```
DonggeLiu, line 65: If `output_dir` is expected to be a direct child dir (e.g., benchmark's output directory), could we use `os.listdir(args.result)`?

Correct me if I am wrong: `next` merely gets the first element from the iterator, which is the same as `[0]`. In that case, could we use `[0]` for clarity?

Taking one step further, I guess the `assert` below:
1. Does not directly check if `args.sub` is one of `[fixed_targets, logs, raw_targets, ..., status]` for flexibility, e.g., some benchmarks do not have them due to exp failures, or we added more sub dirs later.
2. Checks if `args.sub` in the *first* (not other) benchmark result dir for simplicity, assuming all benchmark result dirs have the same sub dirs.
In this case, would it be better to assert *at least one of the benchmark result dirs contain `sub`* for better flexibility without losing simplicity?  E.g., use `any`.

helper/result_string_search.py
```diff
@@ -0,0 +1,137 @@
...
+  assert os.path.isdir(os.path.join(args.result, output_dir, args.sub)), (
+      '--sub must be a directory in output-* directories under <result>:\n'
+      'e.g. fixed_targets, logs, raw_targets, status.')
+
+  if args.url:
+    assert '[benchmark]' in args.url, (
+        "--url must contain '[benchmark]'\n"
```
DonggeLiu, line 72: nit: '--url must contain "[benchmark]"\n'
Same below for consistency.

helper/result_string_search.py
```diff
@@ -0,0 +1,137 @@
...
+  return args
+
+
+def find_in_dir(search_lines: list[str], file_paths: list[str]) -> bool:
+  """Returns True if any file in |file_paths| contains |search_lines|."""
+  # Caveat: With support for multiline search in potentially large files
+  # (e.g. log files), this function does not search for the exact substring
+  # in the file containing the new line char. Instead, it returns True when:
+  # something <search line 1> something
+  # some <search line 2> some
```
DonggeLiu, line 84: typo?

helper/result_string_search.py
cjx10: Using `next` on `os.walk()` because it returns a generator instead of an iterator, so we can't get the element at a certain index. We can even modify the result on the fly, e.g. [ignore corpora](https://github.com/google/oss-fuzz-gen/pull/136/files#diff-74f98584d5ba085369bd1b5bca0319fa225807042fc53f1acd67732aa7acd2ccR117)

Both assumptions are true. Will use any, thanks :)

PR #141, opened 2024-03-06
data_prep: use argparse for introspector

DavidKorczynski:
Use argparse for introspector's module. This fixes a todo and will also make it easier to add more feature to the introspector module.

### DIFF
--- a/data_prep/introspector.py
+++ b/data_prep/introspector.py
@@ -456,19 +456,19 @@ def get_project_funcs(project_name: str) -> Dict[str, List[Dict]]:
 def _parse_arguments() -> argparse.Namespace:
   """Parses command line args."""
   parser = argparse.ArgumentParser(
-      description='Parse arguments to generate benchmarks')
+      description='Parse arguments to generate benchmarks.')
 
-  parser.add_argument('project', help='Name of the project', type=str)
+  parser.add_argument('project', help='Name of the project.', type=str)
   parser.add_argument('-m',
                       '--max-functions',
                       type=int,
                       default=3,
-                      help='Number of benchmarks to generate')
+                      help='Number of benchmarks to generate.')
   parser.add_argument('-o',
                       '--out',
                       type=str,
                       default='',
-                      help='Output directory')
+                      help='Output directory.')
 
   args = parser.parse_args()
   return args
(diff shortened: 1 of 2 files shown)


DonggeLiu:
> seems to be unrelated to the changes?

Yes, it is unrelated.
That CI builds a docker image based on the code in the PR and pushes it to an artifact registry so that we can run experiments with it.

data_prep/introspector.py
```diff
@@ -452,32 +453,46 @@ def get_project_funcs(project_name: str) -> Dict[str, List[Dict]]:
   return fuzz_target_funcs
 
 
+def _parse_arguments() -> argparse.Namespace:
+  """Parses command line args."""
+  parser = argparse.ArgumentParser(
+      description='Parse arguments to generate benchmarks')
```
DonggeLiu, line 459: Nitpicking: For better consistency, could you please end the description with a period (e.g., `.`)? Thanks!
Same below.

data_prep/introspector.py
```diff
@@ -452,32 +453,46 @@ def get_project_funcs(project_name: str) -> Dict[str, List[Dict]]:
...
-  max_num_function = 3
-  if len(sys.argv) > 2:
-    max_num_function = int(sys.argv[2])
-  if len(sys.argv) > 3:
-    outdir = sys.argv[3]
-    os.makedirs(outdir, exist_ok=True)
-  else:
-    outdir = ''
+  args = _parse_arguments()
+  if args.out != '':
```
DonggeLiu, line 481: Nitpicking: How about `if args.out:` as we have specified its type to `str` above?

DonggeLiu (approved):
Thanks so much for the PR, @DavidKorczynski !
I left some minor nitpicking comments, hope you would not mind :)

data_prep/introspector.py
DavidKorczynski: Sounds good, done!

data_prep/introspector.py
DavidKorczynski: Sounds good, done!

DonggeLiu (approved):
Thanks again, @DavidKorczynski !

PR #144, opened 2024-03-07
Use introspector to retrieve context.

trashvisor:
Port context retrieval to use introspector APIs (moving from ASTs to DWARF debug information).


trashvisor:
There are a few issues preventing this from working at the moment.

Some include...
Enums not being a part of the information we receive from FI. Enums are occasionally used in types.
FI only returning a single element for a type, when multiple are typically required to be embedded to get a syntactically sound program (.e.g. typedef+struct). Fix is already in place from FIs end but not currently live (I think).

DavidKorczynski:
Would it be smarter to wait for an API that gives types recursively in FI? I think we should be able to this ready for Monday 11th

trashvisor:
hey @DavidKorczynski that was something we were going to ask about in Monday actually, it would definitely be strongly preferred

DavidKorczynski:
Sounds good -- I'll aim to have something ready on FIs side for then.

data_prep/project_context/context_introspector.py
```diff
@@ -0,0 +1,203 @@
...
+      type_tokens.remove('volatile')
+
+    # If there is more than a single token
+    # we probably do not care about querying for the type (?)
+    # E.g. unsigned [...], long [...], short [...], ...
+    # as they're most likely builtin.
+    if len(type_tokens) > 1:
+      return ''
+
+    return type_tokens[0]
```
DonggeLiu, line 194: I presume this will be delegated to FI later?

prompts/template_xml/context.txt
```diff
@@ -1,8 +1,4 @@
-The function is defined in
-<include_file>
-{CONTEXT_HEADER}
-</include_file>
-
+You must insert this code before the function being tested:
```
DonggeLiu, line 1: 1. How about other context info, such as [function implementation](https://github.com/google/oss-fuzz-gen/blob/main/data_prep/introspector.py#L135) and [cross references](INTROSPECTOR_XREF)?

2. LLM does not have to include all context info, such as the two above.

llm_toolkit/prompt_builder.py
```diff
@@ -211,26 +208,20 @@ def _add_examples(self,
...
-      example_pair: list[list[str]],
-      project_example_content: Optional[list[list[str]]] = None,
-      project_context_content: Optional[Tuple[str,
-                                              str]] = None) -> prompts.Prompt:
+  def build(self,
+            function_signature: str,
+            target_file_type: FileType,
+            example_pair: list[list[str]],
+            project_example_content: Optional[list[list[str]]] = None,
+            project_context_content: Optional[str] = None) -> prompts.Prompt:
```
DonggeLiu, line 216: I presume `Optional` is not required for `project_context_content`, because it will be `''` if `use_context` is false? E.g.,
```python
project_context_content: str = ''
```

DonggeLiu (changes requested):
Thanks for the PR, @trashvisor !

Left some nit comments, the main one is missing [function implementation](https://github.com/google/oss-fuzz-gen/blob/main/data_prep/introspector.py#L135) and [cross references](https://github.com/google/oss-fuzz-gen/pull/144/INTROSPECTOR_XREF).

Could we have a flag for each, so that we can test their effect on the performance?
Thanks!

data_prep/project_context/context_introspector.py
```diff
@@ -0,0 +1,202 @@
...
+
+class ContextRetriever:
+  """Class to retrieve context from introspector for
+  better prompt generation."""
+
+  def __init__(self, benchmark: benchmarklib.Benchmark):
+    """Constructor."""
+    self._benchmark = benchmark
+
+  def get_embeddable_declaration(self) -> str:
```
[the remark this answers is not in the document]
trashvisor: i was thinking of making those public functions, in case some consumer of the api wanted to get the type-list / declaration separately. i cant think of a situation where they might want to be separated at the moment though. i can move these to private

data_prep/project_context/context_introspector.py
trashvisor: ideally this isn't required yup. currently parameter types in the benchmark yaml files look like `struct x` though and only `x` is expected during the FI lookup.

data_prep/project_context/context_introspector.py
```diff
@@ -0,0 +1,203 @@
...
+      # Then we can add it ourselves. This would cause problems
+      # when typedef and struct definitions are combined.
+      # The alternative is to query for source code lines until a '}' is found.
+      if curr_line == end_line_elem:
+        if '}' not in source_line:
+          reconstructed_type += '};\n'
+
+      curr_line += 1
+
+      if not newly_seen_type or newly_seen_type in seen_types:
```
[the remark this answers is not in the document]
trashvisor: sorry, newly_seen_type is badly named. it should be `potential_new_type`, as it could be an old type. in the case where it was a previously seen type, it could be added to types_to_get and be looked up again redundantly.

data_prep/project_context/context_introspector.py
```diff
@@ -0,0 +1,203 @@
...
+          self._extract_type_from_source_line(
+              source_line, elements[curr_line - begin_line_elem]))
+
+      # If we do not see a '}' in the final element's source line
+      # Then we can add it ourselves. This would cause problems
+      # when typedef and struct definitions are combined.
+      # The alternative is to query for source code lines until a '}' is found.
+      if curr_line == end_line_elem:
+        if '}' not in source_line:
+          reconstructed_type += '};\n'
```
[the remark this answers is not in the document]
trashvisor: i originally did it like this as it made sense to me, but the prepush hook complained about source_line being potentially unassigned. should i just assign it to an empty string before the loop and move it outside the loop so the hook doesnt complain?

llm_toolkit/prompt_builder.py
trashvisor: that sounds right. fixing

data_prep/project_context/context_introspector.py
```diff
@@ -0,0 +1,203 @@
...
+
+    types_to_get.add(self._clean_type(self._benchmark.return_type))
+
+    params = self._benchmark.params
+
+    for param in params:
+      cleaned_type = self._clean_type(param['type'])
+      if not cleaned_type:
+        continue
+      types_to_get.add(cleaned_type)
```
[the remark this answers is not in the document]
trashvisor: ack

data_prep/project_context/context_introspector.py
```diff
@@ -0,0 +1,202 @@
...
+
+  def get_embeddable_declaration(self) -> str:
+    """Retrieve declaration by language. Attach extern C to C projects."""
+    lang = self._benchmark.language.lower()
+    sig = self._benchmark.function_signature
+
+    if lang == 'c++':
+      return sig + ';'
+    if lang == 'c':
+      return 'extern "C" ' + sig + ';'
```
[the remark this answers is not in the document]
trashvisor: ack

data_prep/project_context/context_introspector.py
```diff
@@ -0,0 +1,202 @@
...
+class ContextRetriever:
+  """Class to retrieve context from introspector for
+  better prompt generation."""
+
+  def __init__(self, benchmark: benchmarklib.Benchmark):
+    """Constructor."""
+    self._benchmark = benchmark
+
+  def get_embeddable_declaration(self) -> str:
+    """Retrieve declaration by language. Attach extern C to C projects."""
```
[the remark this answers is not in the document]
trashvisor: ack

data_prep/project_context/context_introspector.py
```diff
@@ -0,0 +1,202 @@
...
+    lang = self._benchmark.language.lower()
+    sig = self._benchmark.function_signature
+
+    if lang == 'c++':
+      return sig + ';'
+    if lang == 'c':
+      return 'extern "C" ' + sig + ';'
+
+    print('Unsupported declaration requested')
+    return ''
```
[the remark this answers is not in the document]
trashvisor: ack, thanks!

data_prep/project_context/context_introspector.py
```diff
@@ -0,0 +1,202 @@
...
+    """Retrieve declaration by language. Attach extern C to C projects."""
+    lang = self._benchmark.language.lower()
+    sig = self._benchmark.function_signature
+
+    if lang == 'c++':
+      return sig + ';'
+    if lang == 'c':
+      return 'extern "C" ' + sig + ';'
+
+    print('Unsupported declaration requested')
```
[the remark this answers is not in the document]
trashvisor: ack - done

data_prep/project_context/context_introspector.py
```diff
@@ -0,0 +1,203 @@
...
+      if not newly_seen_type or newly_seen_type in seen_types:
+        continue
+
+      seen_types.add(newly_seen_type)
+      types_to_get.add(newly_seen_type)
+
+    return reconstructed_type
+
+  def get_embeddable_types(self) -> list[str]:
+    """Retrieve types from FI."""
```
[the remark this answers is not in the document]
trashvisor: ack

data_prep/project_context/context_introspector.py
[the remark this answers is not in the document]
DonggeLiu: Yes, assigning it to an empty string is the simplest fix. Alternatively, we can check if the last line in `reconstructed_type` contains `}`, which is a bit more complicated.

The linter in pre-push hook is correct about this: if the loop was never executed, `source_line` will never be defined.

data_prep/project_context/context_introspector.py
```diff
@@ -0,0 +1,203 @@
...
+
+    # The reason we need to iteratively query FI is because
+    # we have to parse each individual source line and
+    # extract a possible type which we will recursively
+    # query the definition for.
+    while curr_line < begin_line_elem:
+      source_line = introspector.query_introspector_source_code(
+          self._benchmark.project, file_name, curr_line, curr_line)
+      curr_line += 1
+      reconstructed_type += source_line
```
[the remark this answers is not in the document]
trashvisor: thanks a lot, this helps and i understand why its important!

i moved it over to two for loops

data_prep/project_context/context_introspector.py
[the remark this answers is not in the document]
trashvisor: done! thanks

data_prep/project_context/context_introspector.py
```diff
@@ -0,0 +1,197 @@
...
+        types.add(reconstructed_type)
+
+    return list(types)
+
+  def _extract_type_from_source_line(self, source_line: str,
+                                     type_element: dict) -> str:
+    """Attempts to extract a type from a source line.
+    Do so by attempting to match for name and extracting
+    everything before it."""
+    start = source_line.find(next(filter(str.isalpha, source_line)))
```
oliverchang, line 147: nit: 

```
next(i for i, c in enumerate(source_line) if str.isalpha(c))
```

might be slightly easier to understand here, as it's less of a roundabout way of achieving this.

data_prep/project_context/context_introspector.py
```diff
@@ -0,0 +1,197 @@
...
+      type_tokens.remove('const')
+
+    if 'volatile' in type_tokens:
+      type_tokens.remove('volatile')
+
+    # If there is more than a single token
+    # we probably do not care about querying for the type (?)
+    # E.g. unsigned [...], long [...], short [...], ...
+    # as they're most likely builtin.
+    if len(type_tokens) > 1:
```
oliverchang, line 185: Can we log this with logging.debug just in case we're missing legitimate types here somehow?

data_prep/project_context/context_introspector.py
```diff
@@ -0,0 +1,197 @@
...
+      info = introspector.query_introspector_type_info(self._benchmark.project,
+                                                       current_type)
+      if not info:
+        continue
+
+      reconstructed_type = ''
+      # For now, info is a single element and not a list.
+      # Requested it be changed to a list, so that type
+      # elements with the same name will be returned.
+      if info['type'] == 'struct':
```
DonggeLiu, line 131: Got this error from GKE experiment:
https://pantheon.corp.google.com/kubernetes/job/us-central1/llm-experiment-large/default/weekly-manual-mar-10/logs?project=oss-fuzz

image

DonggeLiu:
GKE experiment (added `--context` to `run_all_experiments.py` in `docker_run.sh`).
https://pantheon.corp.google.com/kubernetes/job/us-central1/llm-experiment-large/default/weekly-manual-mar-10/logs?project=oss-fuzz

data_prep/project_context/context_introspector.py
DavidKorczynski: Likely because `type-info` now returns a list of types with the given name as opposed to a single type.

DonggeLiu:
JOB: https://pantheon.corp.google.com/kubernetes/job/us-central1-c/llm-experiment/default/ofg-pr-144-dg
REPORT: https://llm-exp.oss-fuzz.com/Result-reports/ofg-pr/2024-03-12-144-dg-comparison/index.html
BUCKET: https://pantheon.corp.google.com/storage/browser/oss-fuzz-gcb-experiment-run-logs/Result-reports/ofg-pr/2024-03-12-144-dg-comparison
BUCKET GS: `gs://oss-fuzz-gcb-experiment-run-logs/Result-reports/ofg-pr/2024-03-12-144-dg-comparison`

oliverchang:
Looking at https://llm-exp.oss-fuzz.com/Result-reports/ofg-pr/2024-03-12-150-dg-comparison/benchmark/output-avahi-avahi_string_list_add_pair,

I'm seeing :


```
typedef __uint8_t uint8_t;

typedef struct AvahiStringList {
    struct AvahiStringList *next; /**< Pointer to the next linked list element */
    size_t size;  /**< Size of text[] */
    uint8_t text[1]; /**< Character data */
};
typedef struct AvahiStringList {
    struct AvahiStringList *next; /**< Pointer to the next linked list element */
    size_t size;  /**< Size of text[] */
    uint8_t text[1]; /**< Character data */
};
typedef struct AvahiStringList {
    struct AvahiStringList *next; /**< Pointer to the next linked list element */
    size_t size;  /**< Size of text[] */
    uint8_t text[1]; /**< Character data */
};
AvahiStringList * avahi_string_list_add_pair(AvahiStringList *, const char *, const char *);
```

Is there some logic we need to add to avoid duplicating the types? @DavidKorczynski

trashvisor:
> Looking at https://llm-exp.oss-fuzz.com/Result-reports/ofg-pr/2024-03-12-150-dg-comparison/benchmark/output-avahi-avahi_string_list_add_pair,
> 
> I'm seeing :
> 
> ```
> typedef __uint8_t uint8_t;
> 
> typedef struct AvahiStringList {
>     struct AvahiStringList *next; /**< Pointer to the next linked list element */
>     size_t size;  /**< Size of text[] */
>     uint8_t text[1]; /**< Character data */
> };
> typedef struct AvahiStringList {
>     struct AvahiStringList *next; /**< Pointer to the next linked list element */
>     size_t size;  /**< Size of text[] */
>     uint8_t text[1]; /**< Character data */
> };
> typedef struct AvahiStringList {
>     struct AvahiStringList *next; /**< Pointer to the next linked list element */
>     size_t size;  /**< Size of text[] */
>     uint8_t text[1]; /**< Character data */
> };
> AvahiStringList * avahi_string_list_add_pair(AvahiStringList *, const char *, const char *);
> ```
> 
> Is there some logic we need to add to avoid duplicating the types? @DavidKorczynski

This is probably an issue on my end. The API we are going to use to query types is going to be changed very soon(there will be an endpoint which supports recursively retrieving type information). The reason this PR was proposed before that change was made was mostly because we initially wanted something out asap to do some fine-tune testing but I don't think that is required at the moment (if my understanding is correct).

oliverchang:
> This is probably an issue on my end. The API we are going to use to query types is going to be changed very soon(there will be an endpoint which supports recursively retrieving type information). The reason this PR was proposed before that change was made was mostly because we initially wanted something out asap to do some fine-tune testing but I don't think that is required at the moment (if my understanding is correct).

Thanks for the explanation! Sorry I haven't been following this too closely lately. I have another question. 

Is the rationale behind the recursive types essentially so we can include all the self contained definitions inline in the targets?

the current prompt structure:

```
You must insert this code before the function being tested:
<code>
... defs ...
</code>
```

seems to indicate so. 

Should we instead see if we can just include the correct header files instead? It seems like trying to recursively expand/inline the necessary definitions can get hairy very quickly. 

IIRC, the header file for a function prototype is difficult to get accurately, but the header file for types are not? Is this true @DavidKorczynski  ? 

If this is true, it may be better to instead do something like:

```
You must include these header files in the solution:
<code>
#include "foo.h"
#include "Bar.h"
</code>

And also include the following function signature:
<code>
void foo(int *blah);
</code>
```

Even if the signature is already defined in one of the headers, I believe it's perfectly valid to repeat function signatures.

trashvisor:
> Should we instead see if we can just include the correct header files instead? It seems like trying to recursively expand/inline the necessary definitions can get hairy very quickly.
> 
> IIRC, the header file for a function prototype is difficult to get accurately, but the header file for types are not? Is this true @DavidKorczynski ?
> 
> If this is true, it may be better to instead do something like:
> 
> ```
> You must include these header files in the solution:
> <code>
> #include "foo.h"
> #include "Bar.h"
> </code>
> 
> And also include the following function signature:
> <code>
> void foo(int *blah);
> </code>
> ```
> 
> Even if the signature is already defined in one of the headers, I believe it's perfectly valid to repeat function signatures.

Really good idea, can totally do that. In fact file information for a type is already exposed. I'll mess around with that idea.

trashvisor:
Upon review, it appears to be a problem with FI.

Multiple items are returned for the same type in a "duplicated" manner.  E.g. Slight variations in source files which normalise to the same path.

```
  1 {"type": "typedef", "name": "AvahiStringList", "source": {"source_file": "/src/avahi/./avahi-common/strlst.h", "source_line": "44"}},
  2 {"type": "struct", "name": "AvahiStringList", "source": {"source_file": "/src/avahi/./avahi-common/strlst.h", "source_line": "40"},
  3  {"type": "typedef", "name": "AvahiStringList", "source": {"source_file": "/src/avahi/avahi-common/./strlst.h", "source_line": "44"}},
  4 {"type": "struct", "name": "AvahiStringList", "source": {"source_file": "/src/avahi/avahi-common/./strlst.h", "source_line": "40"}
  ```

**Comment by DavidKorczynski** [comment]
> Should we instead see if we can just include the correct header files instead? It seems like trying to recursively expand/inline the necessary definitions can get hairy very quickly.

We've pushed some changes to make looking up type information easier in FI based on raw DWARF info in https://github.com/ossf/fuzz-introspector/pull/1471 -- there is also an API "addr-to-recursive-dwarf-info" that can be used to extract full type hierarchy for a given type, e.g. the types used in a function signature. There will still be some work on the users end, for example deciding how deep in the type tree one is interested in looking.

> IIRC, the header file for a function prototype is difficult to get accurately, but the header file for types are not? Is this true @DavidKorczynski ?

Correct, we don't have the location for a given function prototype defined in a header file, but we have locations for all types (caveat: elements in an enum are not considered types, the enum is a type in and of itself, so we need to have heuristics here).

File: prompts/template_xml/context.txt (hunk 2 above)
Comment (trashvisor) [reply prompts/template_xml/context.txt]: added

File: data_prep/project_context/context_introspector.py (hunk 14 above)
Comment (trashvisor) [reply data_prep/project_context/context_introspector.py]: done thanks

File: data_prep/project_context/context_introspector.py (hunk 15 above)
Comment (trashvisor) [reply data_prep/project_context/context_introspector.py:340]: added, yup.

File: data_prep/project_context/context_introspector.py (hunk 16 above)
Comment (trashvisor) [reply data_prep/project_context/context_introspector.py]: yeah, this has been addressed. we will likely move away from type-info for now and go with `addr-to-recursive-dwarf-info`

File: data_prep/project_context/context_introspector.py [hunk 17]
```diff
@@ -0,0 +1,202 @@
...
+    if lang == 'c':
+      return 'extern "C" ' + sig + ';'
+
+    print('Unsupported declaration requested')
+    return ''
+
+  def _get_typedef_type(self, info: dict) -> str:
+    """Reconstructs type definition for a typedef element.
+    Simply queries for a single line in source code."""
+    file_name = os.path.normpath(info['source']['source_file'])
```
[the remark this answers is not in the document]
trashvisor: done

DonggeLiu:
@trashvisor I updated a new feature to main, so I rebased this PR, fixed the merge conflict, and made a minor naming modification, hope you would not mind : )

data_prep/project_context/context_introspector.py
```diff
@@ -283,22 +283,29 @@ def _get_files_to_include(self) -> list[str]:
...
           continue
-        # Get the character after the '.', ensure it is a h
-        if source_base.split('.')[1][0] != 'h':
+        # Get the character after the '.', ensure it is a h.
+        if source_base.split('.')[-1][0] != 'h':
           logging.debug('Non .hxx file found as a header file %s', source_file)
           continue
+        # Remove "system" header files.
+        # Just do this by seeing if the source_file begins with a /usr/
+        if source_file[:5] == '/usr/':
```
DonggeLiu, line 305: Could you please add a `TODO`:
```python
# TODO: Dynamically adjust path prefixes (e.g., based on existing fuzz targets).
```

data_prep/project_context/context_introspector.py
```diff
@@ -283,22 +283,29 @@ def _get_files_to_include(self) -> list[str]:
...
-        # Ensure source_base is a file
-        if len(source_base) == 0 or '.' not in source_base:
+
+        # Ensure source_base is a file.
+        if not source_base or '.' not in source_base:
           logging.debug('File %s found as a source path', source_file)
           continue
-        # Get the character after the '.', ensure it is a h
-        if source_base.split('.')[1][0] != 'h':
+        # Get the character after the '.', ensure it is a h.
```
DonggeLiu, line 299: nit: 
```python
# Ensure it is a header file (suffix starting with an h).
```
BTW, I presume this is a header file, not a source file?

data_prep/project_context/context_introspector.py
```diff
@@ -283,22 +283,29 @@ def _get_files_to_include(self) -> list[str]:
...
+        if source_base.split('.')[-1][0] != 'h':
           logging.debug('Non .hxx file found as a header file %s', source_file)
```
DonggeLiu, line 301: Do we always expect it to be `.hxx` or can it be `.h` too?

DonggeLiu:
BTW, I suppose we should merge this with the `--context` flag disabled.

It's better to merge it to avoid potential merge conflicts.
Better to disable the flag as the performance is not 100% ready.

data_prep/project_context/context_introspector.py
trashvisor: Can be a .h, .hxx, .hpp or whatever other variant. I'm mostly looking at eliminating .c[variant] files that make it into there.

data_prep/project_context/context_introspector.py
trashvisor: yup, source[file/line] in this case refers to the file or line which defines the type.

data_prep/project_context/context_introspector.py
```diff
@@ -283,22 +283,29 @@ def _get_files_to_include(self) -> list[str]:
     for current_type in types:
       info_list = introspector.query_introspector_type_info(
           self._benchmark.project, current_type)
+      print("INFO LIST: {}".format(info_list))
```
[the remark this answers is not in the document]
trashvisor: oh yup oops, that was for debugging locally

data_prep/project_context/context_introspector.py
trashvisor: ack

trashvisor:
> BTW, I suppose we should merge this with the `--context` flag disabled.
> 
> It's better to merge it to avoid potential merge conflicts. Better to disable the flag as the performance is not 100% ready.

that sounds good, i'll clean the code up then

data_prep/project_context/context_introspector.py
DonggeLiu: Shall we change the `.hxx` in the log to be more accurate?

data_prep/project_context/context_introspector.py
DonggeLiu: Shall we change the var name to reflect that?

data_prep/project_context/context_introspector.py
trashvisor: sounds good! ack

data_prep/project_context/context_introspector.py
trashvisor: done!

data_prep/project_context/context_introspector.py
```diff
@@ -283,46 +69,35 @@ def _get_files_to_include(self) -> list[str]:
     for current_type in types:
       info_list = introspector.query_introspector_type_info(
           self._benchmark.project, current_type)
-      print("INFO LIST: {}".format(info_list))
       if not info_list:
+        logging.debug('Could not retrieve info for type: %s', current_type)
```
DonggeLiu, line 73: Could you please also log the project name?
This will help identify which project to debug :)

Also, `logging.warning` is more suitable here because this is unexpected and may indicate an error elsewhere.

data_prep/project_context/context_introspector.py
```diff
@@ -283,46 +69,35 @@ def _get_files_to_include(self) -> list[str]:
...
+        include_file = self._get_source_file(info)
+        include_file = os.path.normpath(include_file)
+        include_base = os.path.basename(include_file)
 
-        # Ensure source_base is a file.
-        if not source_base or '.' not in source_base:
-          logging.debug('File %s found as a source path', source_file)
+        # Ensure include_file is a file.
+        if not include_base or '.' not in include_base:
+          logging.debug('File %s found as a source path', include_file)
```
DonggeLiu, line 83: Same as above:
1. Could we log the project name to assist in debugging?
2. Please use `warning` to surface this.

data_prep/project_context/context_introspector.py
```diff
@@ -283,46 +69,35 @@ def _get_files_to_include(self) -> list[str]:
...
           continue
-        # Get the character after the '.', ensure it is a h.
-        if source_base.split('.')[-1][0] != 'h':
-          logging.debug('Non .hxx file found as a header file %s', source_file)
+        # Ensure it is a header file (suffix starting with .h).
+        if include_base.split('.')[-1][0] != 'h':
```
DonggeLiu, line 86: Could we use [`endswith('.h')`](https://docs.python.org/3/library/stdtypes.html#str.endswith) for simplicity?

data_prep/project_context/context_introspector.py
```diff
@@ -283,46 +69,35 @@ def _get_files_to_include(self) -> list[str]:
...
+          logging.debug('File found with unexpected suffix %s', include_file)
```
DonggeLiu, line 87: Same here : ) Log project name use `warning` for debugging.

data_prep/project_context/context_introspector.py
```diff
@@ -283,46 +69,35 @@ def _get_files_to_include(self) -> list[str]:
...
           continue
         # Remove "system" header files.
-        # Just do this by seeing if the source_file begins with a /usr/
-        if source_file[:5] == '/usr/':
-          logging.debug('Header file removed: %s', source_file)
+        # Just do this by seeing if the full path begins with a /usr/.
+        if include_file[:5] == '/usr/':
```
DonggeLiu, line 91: How about `startswith()`?

data_prep/project_context/context_introspector.py
```diff
@@ -283,46 +69,35 @@ def _get_files_to_include(self) -> list[str]:
...
-          logging.debug('Non .hxx file found as a header file %s', source_file)
+        # Ensure it is a header file (suffix starting with .h).
+        if include_base.split('.')[-1][0] != 'h':
+          logging.debug('File found with unexpected suffix %s', include_file)
           continue
         # Remove "system" header files.
-        # Just do this by seeing if the source_file begins with a /usr/
-        if source_file[:5] == '/usr/':
-          logging.debug('Header file removed: %s', source_file)
+        # Just do this by seeing if the full path begins with a /usr/.
```
DonggeLiu, line 90: A more descriptive alternative:
```
# Assuming header files under /usr/ are irrelevant.
```

DonggeLiu (commented):
Thanks @trashvisor!
Some nit comments.

prompts/template_xml/context.txt
```diff
@@ -1,8 +1,21 @@
...
+<code>
+{CONTEXT_MUST_INSERT}
+</code>
+
+Here is the source code of the function being tested:
+<code>
+{CONTEXT_FUNC_SOURCE}
+</code>
 
+Here is the source code for functions which reference the function being tested:
```
oliverchang, line 18: Can we omit these sections if any of them are empty? Instead of just having an empty code block? 

I wonder if that's impacting the efficacy of this.

oliverchang: I did this in https://github.com/google/oss-fuzz-gen/pull/144/commits/5ac4932e7924a049c0b6b6a48ff32c2577f0bb9f

oliverchang:
@trashvisor I pushed a small change in https://github.com/google/oss-fuzz-gen/pull/144/commits/5ac4932e7924a049c0b6b6a48ff32c2577f0bb9f to:

-  use jinja2 for the template formatting. this enables conditional insertion of context only when they exist. 
- add a `--dry-run` flag to run_all_experiments.py which only generates prompts. this can be used to easily look at the results of context-enhanced prompts locally without actually running the full experiment.

prompts/template_xml/context.txt
```diff
@@ -1,8 +1,30 @@
...
+Here is the source code of the function being tested:
+<code>
+{{ func_source }}
+</code>
+{% endif %}
+{% if xrefs %}
+
+Here is the source code for functions which reference the function being tested:
+<code>
+{{ xrefs }}
```
oliverchang, line 28: When this works, it seems like we are including the entire source code of the functions that call the function-to-fuzz. e.g. on https://llm-exp.oss-fuzz.com/Result-reports/ofg-pr/2024-05-03-192-test-244-comparison/benchmark/output-libvpx-parse_block/index.html 

This is **a lot** of context. Would there be an easy way to instead say, include up to 10 lines before and 10 lines after the relevant function references? 

I.e.

```
... 10 previous lines ...
<function call>
... 10 lines after ...
```

Rather than the entire source code?

@DavidKorczynski for thoughts too.

data_prep/project_context/context_introspector.py
trashvisor: ACK done

data_prep/project_context/context_introspector.py
trashvisor: ACK

data_prep/project_context/context_introspector.py
trashvisor: ACK

data_prep/project_context/context_introspector.py
trashvisor: didnt know about these functions! done

data_prep/project_context/context_introspector.py
trashvisor: added!

trashvisor:
> @trashvisor I pushed a small change in [5ac4932](https://github.com/google/oss-fuzz-gen/commit/5ac4932e7924a049c0b6b6a48ff32c2577f0bb9f) to:
> 
> * use jinja2 for the template formatting. this enables conditional insertion of context only when they exist.
> * add a `--dry-run` flag to run_all_experiments.py which only generates prompts. this can be used to easily look at the results of context-enhanced prompts locally without actually running the full experiment.

thanks for this. will be ultra ultra helpful

prompts/template_xml/context.txt
trashvisor: i could the split returned code into lines and then do a string match and truncate based on that, though there might be complications with c++.

DonggeLiu (approved):
Thanks @trashvisor!
I will merge this now : )

PR #145, opened 2024-03-07
Make it possible to point to a local deployment of Fuzz Introspector

DavidKorczynski:
--> This builds on top of https://github.com/google/oss-fuzz-gen/pull/141 so will rebase once that has landed.

Adds ability to use a local deployment of Fuzz Introspector's web api instead of using https://introspector.oss-fuzz.com/

This has several benefits. Deploying a local version and pointing oss-fuzz-gen to this will:
- For larger experiments avoid any timeout issues incurred by too many requests to `introspector.oss-fuzz.com`. Local will have no issues responding to requests.
- Testing new things will go faster as it's not necessary to deploy a new version of the API by way of  `introspector.oss-fuzz.com` before oss-fuzz-gen can use it.
- If for some reason some limitations or regressions are landed on `introspector.oss-fuzz.com` then oss-fuzz-gen testing against introspector isn't blocked because using the local version (where the regression can be fixed) gives the exact same features as `introspector.oss-fuzz.com`.
- it's easier to test against data from different dates. `introspector.oss-fuzz.com` is updated once a day against the latest data from oss-fuzz. However, a local version can create a DB against oss-fuzz's state at any given date.

To test this:

1. Follow the instructions here to set up a local webapp: https://github.com/ossf/fuzz-introspector/tree/main/tools/web-fuzzing-introspection#launching-a-local-version

2. Run in oss-fuzz-gen: `python3 -m data_prep.introspector tinyxml2 -m 5 -o test5 -l`

### DIFF
--- a/data_prep/introspector.py
+++ b/data_prep/introspector.py
@@ -46,14 +46,15 @@ INTROSPECTOR_TYPE = ''
 INTROSPECTOR_FUNC_SIG = ''
 
 
-def _set_introspector_endpoints(is_local):
+def _set_introspector_endpoints(is_local: bool):
+  """Sets URLs for Fuzz Introspector endpoints to local or remote endpoints."""
   global INTROSPECTOR_ENDPOINT, INTROSPECTOR_CFG, INTROSPECTOR_FUNCTION, INTROSPECTOR_SOURCE, INTROSPECTOR_XREF, INTROSPECTOR_TYPE, INTROSPECTOR_FUNC_SIG
 
   if is_local:
-    logging.info('Setting Fuzz Introspector endpoint to local')
+    logging.info('Setting Fuzz Introspector endpoint to local.')
     INTROSPECTOR_ENDPOINT = 'http://127.0.0.1:8080/api'
   else:
-    logging.info('Setting Fuzz Introspector endpoint to remote')
+    logging.info('Setting Fuzz Introspector endpoint to remote.')
     INTROSPECTOR_ENDPOINT = 'https://introspector.oss-fuzz.com/api'
 
   INTROSPECTOR_CFG = f'{INTROSPECTOR_ENDPOINT}/annotated-cfg'
--- a/data_prep/introspector.py
+++ b/data_prep/introspector.py
@@ -456,19 +456,19 @@ def get_project_funcs(project_name: str) -> Dict[str, List[Dict]]:
 def _parse_arguments() -> argparse.Namespace:
   """Parses command line args."""
   parser = argparse.ArgumentParser(
-      description='Parse arguments to generate benchmarks')
+      description='Parse arguments to generate benchmarks.')
 
-  parser.add_argument('project', help='Name of the project', type=str)
+  parser.add_argument('project', help='Name of the project.', type=str)
   parser.add_argument('-m',
                       '--max-functions',
                       type=int,
                       default=3,
-                      help='Number of benchmarks to generate')
+                      help='Number of benchmarks to generate.')
   parser.add_argument('-o',
                       '--out',
                       type=str,
                       default='',
-                      help='Output directory')
+                      help='Output directory.')
 
   args = parser.parse_args()
   return args
@@ -478,7 +478,7 @@ if __name__ == '__main__':
   logging.basicConfig(level=logging.INFO)
 
   args = _parse_arguments()
-  if args.out != '':
+  if args.out:
     os.makedirs(args.out, exist_ok=True)
 
   try:
--- a/data_prep/introspector.py
+++ b/data_prep/introspector.py
@@ -46,16 +46,12 @@ INTROSPECTOR_TYPE = ''
 INTROSPECTOR_FUNC_SIG = ''
 
 
-def _set_introspector_endpoints(is_local: bool):
+def _set_introspector_endpoints(endpoint):
   """Sets URLs for Fuzz Introspector endpoints to local or remote endpoints."""
   global INTROSPECTOR_ENDPOINT, INTROSPECTOR_CFG, INTROSPECTOR_FUNCTION, INTROSPECTOR_SOURCE, INTROSPECTOR_XREF, INTROSPECTOR_TYPE, INTROSPECTOR_FUNC_SIG
 
-  if is_local:
-    logging.info('Setting Fuzz Introspector endpoint to local.')
-    INTROSPECTOR_ENDPOINT = 'http://127.0.0.1:8080/api'
-  else:
-    logging.info('Setting Fuzz Introspector endpoint to remote.')
-    INTROSPECTOR_ENDPOINT = 'https://introspector.oss-fuzz.com/api'
+  INTROSPECTOR_ENDPOINT = endpoint
+  logging.info(f'Fuzz Introspector endpoint set to {INTROSPECTOR_ENDPOINT}')
 
   INTROSPECTOR_CFG = f'{INTROSPECTOR_ENDPOINT}/annotated-cfg'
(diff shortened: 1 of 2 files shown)


DavidKorczynski:
Merged on top of https://github.com/google/oss-fuzz-gen/pull/141 -- this is ready for review

data_prep/introspector.py
```diff
@@ -37,13 +37,31 @@
...
+INTROSPECTOR_XREF = ''
+INTROSPECTOR_TYPE = ''
+INTROSPECTOR_FUNC_SIG = ''
+
+
+def _set_introspector_endpoints(is_local):
+  global INTROSPECTOR_ENDPOINT, INTROSPECTOR_CFG, INTROSPECTOR_FUNCTION, INTROSPECTOR_SOURCE, INTROSPECTOR_XREF, INTROSPECTOR_TYPE, INTROSPECTOR_FUNC_SIG
+
+  if is_local:
+    logging.info('Setting Fuzz Introspector endpoint to local')
```
DonggeLiu, line 53: nit: End this with a period.

data_prep/introspector.py
```diff
@@ -37,13 +37,31 @@
...
+    INTROSPECTOR_ENDPOINT = 'http://127.0.0.1:8080/api'
+  else:
+    logging.info('Setting Fuzz Introspector endpoint to remote')
```
DonggeLiu, line 56: nit: Append a period, thanks: )

DonggeLiu (approved):
Some nits, thanks @DavidKorczynski !

DonggeLiu:
> For larger experiments avoid any timeout issues incurred by too many requests to introspector.oss-fuzz.com. Local will have no issues responding to requests.

That's a very good point, we did see timeouts in past experiments.
Would it make sense to automatically start a localhost in [`docker_run.sh`](https://github.com/google/oss-fuzz-gen/blob/main/report/docker_run.sh) so that all cloud experiments (large and small) use this?

DavidKorczynski:
> 
> Would it make sense to automatically start a localhost in [`docker_run.sh`](https://github.com/google/oss-fuzz-gen/blob/main/report/docker_run.sh) so that all cloud experiments (large and small) use this?

I think we should strive for this but not do this by default yet.

A caveat is that it takes some time to create the DB. This document [here](https://github.com/ossf/fuzz-introspector/blob/8431f7459376cf663df666d8bd77c8b67f9c5f27/tools/web-fuzzing-introspection/app/static/assets/db/README.md) (doc is waiting to be merged in) gives a bit of context.

A couple of factors play a role in how long it takes, but most importantly is how many projects to include in the DB. If you are going to traverse the entire OSS-Fuzz project list then it can take an hour or more. If you're doing a handful of projects then it will take a minute.

However, there are some factors we can do to speed things up (such as always exclude projects in languages not supported by oss-fuzz-gen), but this wasn't a major concern so far. If you think this would be nice to have by default then we could likely get it to satisfactory speed (10-20 min perhaps) even when doing large full-scale OSS-Fuzz analysis.

That said, the DB just need to be created once, assuming no new features in a certain part of the pipeline are added, and this can then be used for multiple runs.

If you try the local version then it would be great to hear from your experience.

DavidKorczynski:
Just thinking more about it -- `docker_run.sh` does it launch a lot of cloud VMs where each VM is responsible for generating harnesses for a given project? If so, then it might make sense to do it such that each worker that generates for a given OSS-Fuzz project will create a local instance of `introspector.oss-fuzz.com` but with only a database for the target project -- that would go fast for each worker since the DB creation will only be for a single project. If so, then there aren't performance concerns.

data_prep/introspector.py
DavidKorczynski: done

data_prep/introspector.py
DavidKorczynski: done

data_prep/introspector.py
```diff
@@ -36,13 +37,31 @@
...
+INTROSPECTOR_ENDPOINT = ''
+INTROSPECTOR_CFG = ''
+INTROSPECTOR_FUNCTION = ''
+INTROSPECTOR_SOURCE = ''
+INTROSPECTOR_XREF = ''
+INTROSPECTOR_TYPE = ''
+INTROSPECTOR_FUNC_SIG = ''
+
+
+def _set_introspector_endpoints(is_local):
```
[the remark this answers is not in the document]
DavidKorczynski: done, I didn't add `-> None` as return value since this is not done in other void functions throughout the module.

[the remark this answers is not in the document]
DonggeLiu: Yes, good point.
We were also planning to add that, but there are too many for now.

DonggeLiu:
Thanks for the detailed response, @DavidKorczynski !

> `docker_run.sh` does it launch a lot of cloud VMs where each VM is responsible for generating harnesses for a given project? i.e. 1 worker for 1 OSS-Fuzz project?

Nope, your previous understanding is correct: It launches an experiment with a set of benchmarks.
We can control which benchmarks/projects to use, but it is not designed for 1 project per experiment.


> If you are going to traverse the entire OSS-Fuzz project list then it can take an hour or more. If you're doing a handful of projects then it will take a minute.

I see. Currently, we run three main kinds of experiments:
1. PR experiments: An experiment using the `comparison` benchmark set to check for regressions/improvements introduced by a PR. 
    * When we add the PR to use localhost, I suggest to use PR experiment to check for potential issues before merging it.
    * But we don't use the local host FI on other PR experiments until the localhost FI is stable. This helps debug other PR's issues.
2. Daily experiment: An experiment using the `comparison` benchmark set runs every day with the most recent commit on GitHub.
    * After we merge the PR, we can enable it in the daily exp for a few days to continuously monitor its behaviour.
3. Weekly exp: An experiment using the `all` benchmark set runs every week.
    * After daily experiment shows localhost FI is stable, we start using it on weekly exp.
    * This exp normally takes days, so spending 1-2 hours to setup the DB at the beginning is fine.

Would this sound good to you? Please let me know if I misunderstood anything.

DavidKorczynski:
> Would this sound good to you? Please let me know if I misunderstood anything.

I think the right way to use it now is for development purposes, as with the PR the FI changes can be tested against oss-fuzz-gen instantly as opposed to after a bump.

I think the options you mention sounds reasonable, but am not strongly opinionated on those -- I can also see the benefit of using introspector.oss-fuzz.com in those cases.

DonggeLiu:
> I think the options you mention sounds reasonable, but am not strongly opinionated on those -- I can also see the benefit of using introspector.oss-fuzz.com in those cases.

I see, I proposed it mainly because I wanted to test if your local host method can avoid `timeout` errors in large-scale experiments that send high-frequency requests to introspector.oss-fuzz.com. If so, it may help boost the performance of many benchmarks.

Other than that, I am also happy with using introspector.oss-fuzz.com, which works for all small-scale experiments.

DavidKorczynski:
I can make it a goal in the near term to apply it in the three 3 cases -- is there anything holding the PR back for now, I would be quite happy to have this through while doing changes in the FI environment?

data_prep/introspector.py
```diff
@@ -469,6 +488,10 @@ def _parse_arguments() -> argparse.Namespace:
                       type=str,
                       default='',
                       help='Output directory.')
+  parser.add_argument('-l',
+                      '--local',
```
oliverchang, line 492: Could we generalize this as `--endpoint`, with a default value of introspector.oss-fuzz.com/api ?

DavidKorczynski: Yeah, done

DonggeLiu:
Seems working alright, let's merge this.

PR #151, opened 2024-03-12
Allow generating training data with old experiment results.

DonggeLiu:
The current script assumes the result directory is available locally, which does not support old cloud experiment results.
The code change downloads necessary files from old experiment bucket directories and uses them to generate training data.

Note: Future experiments will automatically generate training data, thanks to #148. This is only used to generate training data from past experiments.

PR #154, opened 2024-03-13
Retry on broken pipe and refactor error handling

cjx10:
[#151 comment](https://github.com/google/oss-fuzz-gen/pull/151#issuecomment-1992869816)

### DIFF
--- a/experiment/builder_runner.py
+++ b/experiment/builder_runner.py
@@ -336,7 +336,7 @@ class CloudBuilderRunner(BuilderRunner):
     retryable_errors = [
         # As mentioned in pr #100.
         ('RESOURCE_EXHAUSTED', lambda x: 5 * 2**x + random.randint(50, 90)),
-        # As mentioned in pr $151.
+        # As mentioned in pr #151.
         ('BrokenPipeError: [Errno 32] Broken pipe',
          lambda x: 5 * 2**x + random.randint(1, 5)),
         # Temp workaround for issue #12.
--- a/experiment/builder_runner.py
+++ b/experiment/builder_runner.py
@@ -356,22 +356,22 @@ class CloudBuilderRunner(BuilderRunner):
         stdout = e.stdout.decode('utf-8').replace('\n', '\t')
         stderr = e.stderr.decode('utf-8').replace('\n', '\t')
 
-        delay = 0
-        for err, delay_f in retryable_errors:
-          if attempt_id < CLOUD_EXP_MAX_ATTEMPT and err in stdout + stderr:
-            delay = delay_f(attempt_id)
-            break
+        delay = next((delay_f(attempt_id)
+                      for err, delay_f in retryable_errors
+                      if err in stdout + stderr), 0)
 
-        if delay:
-          logging.warning(
-              'Failed to evaluate %s on cloud, attempt %d, retry in %ds:\n'
-              '%s\n%s', os.path.realpath(target_path), attempt_id, delay,
-              stdout, stderr)
-          time.sleep(delay)
-          continue
-        logging.error('Failed to evaluate %s on cloud, attempt %d:\n%s\n%s',
-                      os.path.realpath(target_path), attempt_id, stdout, stderr)
-        break
+        if not delay or attempt_id == CLOUD_EXP_MAX_ATTEMPT:
+          logging.error('Failed to evaluate %s on cloud, attempt %d:\n%s\n%s',
+                        os.path.realpath(target_path), attempt_id, stdout,
+                        stderr)
+          break
+
+        logging.warning(
+            'Failed to evaluate %s on cloud, attempt %d, retry in %ds:\n'
+            '%s\n%s', os.path.realpath(target_path), attempt_id, delay, stdout,
+            stderr)
+        time.sleep(delay)
+
     return False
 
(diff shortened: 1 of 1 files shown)


cjx10:
JOB: https://pantheon.corp.google.com/kubernetes/job/us-central1-c/llm-experiment/default/ofg-pr-154-jim
REPORT: https://llm-exp.oss-fuzz.com/Result-reports/ofg-pr/2024-03-13-154-jim-comparison/index.html
BUCKET: https://pantheon.corp.google.com/storage/browser/oss-fuzz-gcb-experiment-run-logs/Result-reports/ofg-pr/2024-03-13-154-jim-comparison
BUCKET GS: `gs://oss-fuzz-gcb-experiment-run-logs/Result-reports/ofg-pr/2024-03-13-154-jim-comparison`

experiment/builder_runner.py
```diff
@@ -322,21 +322,60 @@ def get_coverage_local(
...
+        if delay:
+          logging.warning(
+              'Failed to evaluate %s on cloud, attempt %d, retry in %ds:\n'
+              '%s\n%s', os.path.realpath(target_path), attempt_id, delay,
+              stdout, stderr)
+          time.sleep(delay)
+          continue
+        logging.error('Failed to evaluate %s on cloud, attempt %d:\n%s\n%s',
+                      os.path.realpath(target_path), attempt_id, stdout, stderr)
+        raise e
```
DonggeLiu, line 377: I guess you added `if attempt_id == CLOUD_EXP_MAX_ATTEMPT:` because you'd like to identify the final failure after `CLOUD_EXP_MAX_ATTEMPT` attempts.
Would it be clearer to move the final failure logging outside the `except`?

For example, here we have:
```python
for attempt_id in range():
  try:
    sp.run()
    return
  except sp.CalledProcessError as e:
  ... decode stdout/stderr...
  delay = 0
  for err, delay_f in retryable_errors:
    if err in stdout + stderr:
      delay = delay_f(attempt_id)
      break
    logging.warning(...)
    time.sleep(delay)
logging.error(... )
raise e
```

experiment/builder_runner.py
```diff
@@ -358,47 +397,23 @@ def build_and_run(self, generated_project: str, target_path: str,
...
+              f'--target={self.benchmark.target_name}',
+              f'--upload_build_log={build_log_path}',
+              f'--upload_output_log={run_log_path}',
+              f'--upload_corpus={corpus_path}',
+              f'--upload_coverage={coverage_path}',
+              f'--experiment_name={self.experiment_name}', '--'
+          ] + self._libfuzzer_args(),
+          cwd=oss_fuzz_checkout.OSS_FUZZ_DIR)
+    except sp.CalledProcessError:
+      return build_result, None
```
DonggeLiu, line 416: nit: You don't have to `raise e` in the function and capture it here, if you don't use `e`.
A simple `return True/False` could do the same trick.

DonggeLiu (approved):

experiment/builder_runner.py
cjx10: We'd still want to log stdout and stderr from the exception.
The last error logging will handle two situations:
1. Exceeding max attempt;
2. Non retry-able error occurred.
Logging the attempt_id can also help distinguish between the two.

What we can do alternatively might be to remove `check=True` and handle process error checking manually without a try-except block

experiment/builder_runner.py
cjx10: Done

experiment/builder_runner.py
```diff
@@ -322,21 +322,60 @@ def get_coverage_local(
...
     super().__init__(*args, **kwargs)
 
+  @staticmethod
+  def _run_with_retry_control(target_path: str, *args, **kwargs) -> bool:
+    """sp.run() with controllable retry and customized exponential backoff."""
+    # List of (error_str, exp_backoff_func).
+    retryable_errors = [
+        # As mentioned in pr #100.
+        ('RESOURCE_EXHAUSTED', lambda x: 5 * 2**x + random.randint(50, 90)),
+        # As mentioned in pr $151.
```
DonggeLiu, line 339: nit: #151

experiment/builder_runner.py
DonggeLiu: > Non retry-able error occurred.

I see. Should we have a `break` after `logging.error()` to avoid going through another iteration?

And I presume `if attempt_id == CLOUD_EXP_MAX_ATTEMPT` is not needed either way, because you have initialised `delay = 0` before that?

DonggeLiu: Also you can access `attempt_id` after the for loop terminates.
It will be the value used in the last iteration.

cjx10: `break` added and max attempt checking merged with error checking

> access `attempt_id` after the for loop terminates

Ah good to know, confusing [scope in python](https://mail.python.org/pipermail/python-ideas/2008-October/002109.html) :(

cjx10: Should look better now with [814580f](https://github.com/google/oss-fuzz-gen/pull/154/commits/814580fd49b09696352dfeb5e037c6a63cc9f303) I think

DonggeLiu:
Thanks, @cjx10 !

PR #155, opened 2024-03-14
Update README.md by fixing an invalid link instance.

MarkLee131:
The original example has been removed and its link is invalid, and an existing case [`capnproto.yaml`](../benchmark-sets/all/capnproto.yaml#L44) is now updated.


DonggeLiu:
Good catch, thank you!

DonggeLiu (approved):

PR #176, opened 2024-03-22
Host Fuzz Introspector locally in cloud experiments

DonggeLiu:
Now: Replicates #145 to run PR experiments.
Later: Complete the TODOs to host FI locally in full experiments, which often have `timeout` error that are likely caused by too frequent access to public `introspector.oss-fuzz.com`.

This PR is ready except these TODOs:
1. Rebase this PR after merging #145.
2. Add a flag to `run_all_experiments.py` to use local FI after merging #144.
3. Pass the flag to `ContextRetriever` so that it can call `_set_introspector_endpoints` with it to use local FI.

PR #178, opened 2024-03-23
introspector: enable additional target selectors

DavidKorczynski:
The current oracle for finding interesting targets for harnesses is by way of `far-reach-low-coverage` from Fuzz Introspector. However, the logic behind this API only represents a subset of ways commonly used to find interesting targets. For example, Fuzz Introspector itself provides multiple oracles for findings targets, some of which are displayed here https://introspector.oss-fuzz.com/target_oracle (it will load, but may take some seconds to do so).

This PR extends the data prep logic to enable one more heuristic from fuzz introspector, that's based on finding functions with interesting keywords in them (`serialize`, `parse`, `read` and such things) in combination with low coverage and decent cyclomatic complexity.

We can extend this into a direction where we can have X target oracles (there are a handful of oracles in FI already) as well as attach the oracle IDs to the benchmarks which can be useful for large-scale experiments to understand how various ways of finding interesting targets perform.

FYI this only works using a local deployment of FI -- a bump is needed on OSS-Fuzz to enable this using introspector.oss-fuzz.com


DonggeLiu:
Thanks @DavidKorczynski!
I am very glad that you are working on this; selecting benchmarks is a great idea.

A similar idea is selecting public functions that take untrusted user inputs.
This will reduce false positive bugs (i.e., bugs are infeasible in real-world usage) by avoiding private functions or functions that do not take user inputs.

We have some preliminary heuristics and code. Would you be interested in considering adding them to FI?
We would really appreciate that.

data_prep/introspector.py
```diff
@@ -40,6 +40,7 @@
 INTROSPECTOR_ENDPOINT = ''
 INTROSPECTOR_CFG = ''
 INTROSPECTOR_FUNCTION = ''
+INTROSPECTOR_ORACLE_1 = ''
```
DonggeLiu, line 43: Nitpicking: Could you please make this name more self-explanatory? E.g., `INTROSPECTOR_ORACLE_KEYWORD`, like the API name.
Thanks!

data_prep/introspector.py
```diff
@@ -48,13 +49,14 @@
 
 def _set_introspector_endpoints(endpoint):
   """Sets URLs for Fuzz Introspector endpoints to local or remote endpoints."""
-  global INTROSPECTOR_ENDPOINT, INTROSPECTOR_CFG, INTROSPECTOR_FUNCTION, INTROSPECTOR_SOURCE, INTROSPECTOR_XREF, INTROSPECTOR_TYPE, INTROSPECTOR_FUNC_SIG
+  global INTROSPECTOR_ENDPOINT, INTROSPECTOR_CFG, INTROSPECTOR_FUNCTION, INTROSPECTOR_SOURCE, INTROSPECTOR_XREF, INTROSPECTOR_TYPE, INTROSPECTOR_FUNC_SIG, INTROSPECTOR_ORACLE_1
```
DonggeLiu, line 52: nit: Since we are changing this, would you please use line breaks to avoid overlong lines?
E.g., https://github.com/google/oss-fuzz-gen/pull/176/files#diff-2988c9b2abbf23c03fa8ed236ed0f62fb249e13ef31c1509b7fef0f6d78dd1beR51-R53

data_prep/introspector.py
```diff
@@ -308,9 +337,9 @@ def _group_function_params(param_types: list[str],
 
 
 def populate_benchmarks_using_introspector(project: str, language: str,
-                                           limit: int):
+                                           limit: int, target_oracle: str):
   """Populates benchmark YAML files from the data from FuzzIntrospector."""
-  functions = get_unreached_functions(project)
+  functions = query_introspector_for_targets(project, target_oracle)
```
DonggeLiu, line 342: nit: Shall we remove functions `get_unreached_functions()` and `query_introspector_for_unreached_functions()` since anyone will no longer use them after this change?

DonggeLiu (approved):
LGTM with some nitpicking

data_prep/introspector.py
DavidKorczynski: Done, used your suggestion and also change the `INTROSPECTOR_FUNCTION` to `INTROSPECTOR_ORACLE_FAR_REACH`

data_prep/introspector.py
DavidKorczynski: I refined a bit, please check. I kept `get_unreached_functions` because I noticed it had a bit of post-processing in the code which I thought we should keep. That made think in general it's nice to be able to process a bit of the output from the FI APIs so made the oracle dictionary a dictionary to functions rather than APIs, the functions can themselves have the relevant APIs now.

DavidKorczynski:
> A similar idea is selecting public functions that take untrusted user inputs. This will reduce false positive bugs (i.e., bugs are infeasible in real-world usage) by avoiding private functions or functions that do not take user inputs.
> 
> We have some preliminary heuristics and code. Would you be interested in considering adding them to FI? We would really appreciate that.

Yeah, definitely. We can try to characteristic this using the conditions that a function must:

- Not be called by any other others in a given project (unless it's a test), and;
- Have external linkage, and;
- Have at least some arguments

?

data_prep/introspector.py
DonggeLiu: Thanks for refining the code.
There appears to be a conflict (after merging #180); could you please resolve that? Much appreciated.

DonggeLiu:
Thanks for listing the ideas.

> * Not be called by any other others in a given project (unless it's a test) and

Good point, but would this be a hard rule?
It may lead to false negatives (i.e., excluding functions that are sensible to fuzz).

> Have external linkage 

I reckon this implies:
* Excluding Functions without Prototypes in Header Files
* Excluding Functions in Anonymous namespaces
* Excluding Static Functions

Are there other specific rules under this idea?

> Have at least some arguments.

Yep!


Also some idea to prioritize functions, please let me know if they make sense: 
* Prioritise functions in the ‘Include’ Directory
    * [Project Example](https://gitlab.xiph.org/xiph/opusfile/-/tree/master/include?ref_type=heads)
* Prioritise functions with high cyclomatic complexity.
* Prioritise functions with high external usage (e.g., from other projects).

data_prep/introspector.py
DavidKorczynski: Rebasing done

DonggeLiu (approved):

data_prep/introspector.py
DonggeLiu: Amazing, thanks!
Approved with a final nit comment regarding leading/trailing spaces in the latest commit.

data_prep/introspector.py
```diff
@@ -49,15 +50,16 @@
 
 def set_introspector_endpoints(endpoint):
   """Sets URLs for Fuzz Introspector endpoints to local or remote endpoints."""
-  global INTROSPECTOR_ENDPOINT, INTROSPECTOR_CFG, INTROSPECTOR_FUNCTION, \
-      INTROSPECTOR_SOURCE, INTROSPECTOR_XREF, INTROSPECTOR_TYPE, \
-      INTROSPECTOR_FUNC_SIG
+  global INTROSPECTOR_ENDPOINT, INTROSPECTOR_CFG, INTROSPECTOR_FUNC_SIG, \
+      INTROSPECTOR_SOURCE, INTROSPECTOR_XREF, INTROSPECTOR_TYPE,         \
```
DonggeLiu, line 54: Thanks, @DavidKorczynski, this could be an auto-formatting issue:
Are the spaces between the last `,` and `\` added by auto-formatter?
Just want to double-check that our auto-formatting script does not get in your way : )

DavidKorczynski: > Are the spaces between the last , and \ added by auto-formatter?

No, I added them to keep the `\` aligned.

Running prehooks give me:

```sh
(.venv) $ .github/helper/presubmit
PYLINTHOME is now '/home/dav/.cache/pylint' but obsolescent '/home/dav/.pylint.d' is found; you can safely remove the latter
************* Module introspector
data_prep/introspector.py:61:0: C0301: Line too long (87/80) (line-too-long)
data_prep/introspector.py:62:0: C0301: Line too long (89/80) (line-too-long)
data_prep/introspector.py:162:4: W1201: Use lazy % formatting in logging functions (logging-not-lazy)

added 1 package, and audited 2 packages in 2s

found 0 vulnerabilities
0 errors, 0 warnings, 0 informations 
```

DonggeLiu: > No, I added them to keep the \ aligned.

Ah I see, we don't have to keep it aligned. Could you please remove it, if you don't mind? Thanks!

> data_prep/introspector.py:61:0: C0301: Line too long (87/80) (line-too-long)
data_prep/introspector.py:62:0: C0301: Line too long (89/80) (line-too-long)

Would you mind fixing these two, a potential solution is using `()`, e.g.:
```python
INTROSPECTOR_ORACLE_FAR_REACH = (f'{INTROSPECTOR_ENDPOINT}/'
                                 'far-reach-but-low-coverage')
``` 
> logging.error('No such oracle "%s"' % (target_oracle))

I think the formatter was suggesting:
```python
logging.error('No such oracle "%s"', target_oracle)
```

DonggeLiu (approved):

data_prep/introspector.py
DavidKorczynski: Yep, done! Can confirm:

```sh
$ .github/helper/presubmit
0 errors, 0 warnings, 0 informations
```

DonggeLiu: Amazing, thanks!

PR #185, opened 2024-03-27
Automated fuzz target filtering

DonggeLiu:


DonggeLiu:
JOB: https://console.cloud.google.com/kubernetes/job/us-central1-c/llm-experiment/default/ofg-pr-185-dg
REPORT: https://llm-exp.oss-fuzz.com/Result-reports/ofg-pr/2024-03-28-185-dg-comparison/index.html
BUCKET: https://console.cloud.google.com/storage/browseross-fuzz-gcb-experiment-run-logs/Result-reports/ofg-pr/2024-03-28-185-dg-comparison
BUCKET GS: `gs://oss-fuzz-gcb-experiment-run-logs/Result-reports/ofg-pr/2024-03-28-185-dg-comparison`

DonggeLiu:
@happy-qop Starting the experiment here.

DonggeLiu:
> No, I cannot access these crash reports~

What's your preferred email address to access them?
We can test adding you after returning from holiday (the coming Tuesday).

> I would like to do this but not sure what is the preferred implementation way to align with your following workflow such as how your experiment scripts work with the new classification data.

I am thinking to start with a very preliminary changes: 
1. Replace the current `crash` with your `is_driver_fuzz_err`.
2. Show the FP reason in the report HTML.

Given you cannot see the report now, it might be quicker for me to do it.

> Besides, this classification information also highly relates with the discussed fix prompt strategies. If you can hint on this, I can adapt my existing code such as further classification & fix prompts in the next few days.

Are there any other more sophisticated changes you'd like to propose?
I'd love to hear more : )

DonggeLiu:
Push the first version here: #191

I may adjust it a bit and will update you once it is ready : )

> The error filtered here is of driver runtime error, identifying them and proposing corresponding fix prompts is part of the workflow of implementing error type specific fix prompts (FIX_FUZZ_XXX at [here](https://sites.google.com/view/llm4fdg/examples-of-prompt-strategies?authuser=0)).

I see, that seems to be a big change indeed.
Let's do that in a separate PR later, then!

experiment/evaluator.py
```diff
@@ -217,7 +292,33 @@ def _parse_libfuzzer_logs(self, log_handle) -> tuple[int, int, bool]:
...
+      # FP case 2: 1st func of the 1st thread stack is in driver.
+      crash_stacks = self._parse_stacks_from_libfuzzer_logs(lines)
+      for stack_frame in crash_stacks[:1]:
+        if self._stack_func_is_of_testing_project(stack_frame):
+          if 'LLVMFuzzerTestOneInput' in stack_frame:
+            return cov_pcs, total_pcs, True, True, 'FP_CRASH_IN_DRIVER'
+          break
+
+    else:
+      # Another error driver case: no cov increase.
```
oliverchang, line 316: note: I believe we have seen cases where the fuzz target is legitimate even when there is no cov increase. This is often when a new target fuzzes a very buggy function that has never been fuzzed before, with a very shallow bug that is instantly triggered. 

@DonggeLiu to confirm if this is an issue. 

We may want to think about some kind of confidence score here instead of a binary yes/no for filtering crashes/fuzz targets.

DonggeLiu: Sure, I will check with old known issues.

DonggeLiu: BTW, would it be possible to capture `LeakSanitizer: detected memory leaks`?
They are usually not security-related and caused by the fuzz target, it would be useful to filter them (or give them a low confidence score). E.g.,
https://llm-exp.oss-fuzz.com/Result-reports/ofg-pr/2024-04-03-198-dg-comparison/sample/output-hiredis-rediscommand/01

DonggeLiu: That's fantastic, much appreciated!

PR #187, opened 2024-03-27
detect and filter incorrect fuzz drivers

happy-qop:
three types of cases are handled:
- driver that has no coverage increase
- false positives by crashing at INITED or first few rounds
- false positives by crashing at `LLVMFuzzerTestOneInput`

### DIFF
--- a/experiment/evaluator.py
+++ b/experiment/evaluator.py
@@ -32,13 +32,15 @@ from llm_toolkit import code_fixer
 LLM_FIX_LIMIT = 5
 
 LIBFUZZER_MODULES_LOADED_REGEX = re.compile(
-    br'^INFO:\s+Loaded\s+\d+\s+(modules|PC tables)\s+\((\d+)\s+.*\).*')
-LIBFUZZER_COV_REGEX = re.compile(br'.*cov: (\d+) ft:')
-LIBFUZZER_CRASH_TYPE_REGEX = re.compile(br'.*Test unit written to.*')
-CRASH_EXCLUSIONS = re.compile(br'.*(slow-unit-|timeout-|leak-|oom-).*')
+    r'^INFO:\s+Loaded\s+\d+\s+(modules|PC tables)\s+\((\d+)\s+.*\).*')
+LIBFUZZER_COV_REGEX = re.compile(r'.*cov: (\d+) ft:')
+LIBFUZZER_CRASH_TYPE_REGEX = re.compile(r'.*Test unit written to.*')
+CRASH_EXCLUSIONS = re.compile(r'.*(slow-unit-|timeout-|leak-|oom-).*')
+CRASH_STACK_WITH_SOURCE_INFO = re.compile(r'in.*:\d+:\d+$')
 
 OSS_FUZZ_COVERAGE_BUCKET = 'oss-fuzz-coverage'
 
+LLVM_SOURCE_PATH_PREFIX = '/src/llvm-project/compiler-rt'
 
 @dataclasses.dataclass
 class Result:
@@ -49,6 +51,9 @@ class Result:
   line_coverage_diff: float = 0.0
   coverage_report_path: str = ''
   reproducer_path: str = ''
+  # produces fp or no cov increase at all
+  is_driver_fuzz_err: bool = False
+  driver_fuzz_err: str = ''
 
   def dict(self):
     return dataclasses.asdict(self)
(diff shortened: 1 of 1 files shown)


experiment/evaluator.py
```diff
@@ -195,13 +201,81 @@ def check_target(self, ai_binary, target_path: str) -> Optional[Result]:
...
+          stack = []
+
+    if len(stack) > 0:
+      stacks.append(stack)
+
+    return stacks
+
+  def _parse_fuzz_cov_info_from_libfuzzer_logs(
+      self, lines) -> tuple[Optional[int], Optional[int], Optional[int]]:
+    """Parse cov of INITED & DONE, and round number from libFuzzer logs."""
```
DonggeLiu, line 230: nit: Parses

experiment/evaluator.py
```diff
@@ -195,13 +201,81 @@ def check_target(self, ai_binary, target_path: str) -> Optional[Result]:
...
+
+    for line in lines:
+      if line.startswith('#'):
+        # parse line starting with #int to get the round number
+        roundno = None
+        for ch in line:
+          if ch.isspace():
+            roundstr = line[1:line.index(ch)]
+            roundno = int(roundstr) if roundstr.isdigit() else None
+            break
```
DonggeLiu, line 241: Can we simplify this with regex? E.g.:
```python
match = re.search(r"\d+", line.strip())
roundno = int(match.group(0)) if match else None
```

experiment/evaluator.py
```diff
@@ -195,13 +201,81 @@ def check_target(self, ai_binary, target_path: str) -> Optional[Result]:
...
+    return initcov, donecov, lastround
+
+  def _stack_func_is_of_testing_project(self, stack_func: str) -> bool:
+    if not CRASH_STACK_WITH_SOURCE_INFO.match(stack_func):
+      # stack_func of linked binary with no source info
+      # TODO: check this?
+      return False
+    if LLVM_SOURCE_PATH_PREFIX in stack_func:
+      return False
+    return True
```
DonggeLiu, line 259: How about:
```python
return bool(CRASH_STACK_WITH_SOURCE_INFO.match(stack_func)) and LLVM_SOURCE_PATH_PREFIX not in stack_func
```

experiment/evaluator.py
```diff
@@ -195,13 +201,81 @@ def check_target(self, ai_binary, target_path: str) -> Optional[Result]:
...
     """Parse libFuzzer logs."""
+    lines = None
+    try:
+      fuzzlog = log_handle.read(-1)
+      # some driver crashes can mess up the libfuzzer output and raise exp
+      fuzzlog = fuzzlog.decode('utf-8', errors='ignore')
+      lines = fuzzlog.split('\n')
+    except MemoryError as _:
+      # some logs from abnormal drivers are too large to be parsed
+      return 0, 0, False, True, 'LOG_MESS_UP'
```
DonggeLiu, line 272: Could you please log this so that we can be aware of it? E.g.,
```python
    except MemoryError as e:
      # some logs from abnormal drivers are too large to be parsed
     logging.waring('%s is too large to parse: %s', log_handle.name, e)
      return 0, 0, False, True, 'LOG_MESS_UP'
```

experiment/evaluator.py
```diff
@@ -217,7 +291,34 @@ def _parse_libfuzzer_logs(self, log_handle) -> tuple[int, int, bool]:
         crashes = True
         continue
 
-    return cov_pcs, total_pcs, crashes
+    initcov, donecov, lastround = self._parse_fuzz_cov_info_from_libfuzzer_logs(
+        lines)
+
+    # NOTE: false positive crash will not be counted as crash
```
DonggeLiu, line 297: nit: Capitalise the first char after `:`, e.g., `# NOTE: False ...`.
Same below.

experiment/evaluator.py
```diff
@@ -217,7 +291,34 @@ def _parse_libfuzzer_logs(self, log_handle) -> tuple[int, int, bool]:
...
+
+    if crashes:
+      # FP case 1: driver crashes at init or first few rounds
+      if lastround is None or lastround <= 3:
```
DonggeLiu, line 301: nit: Global var for `3`, e.g., `EARLY_FUZZING_ROUND_THRESHOLD`.

DonggeLiu (commented):
Thanks for having this ready so quickly, @happy-qop!
I left a few minor nitpicking comment to help simplify the logic and unify coding style with the rest of the repo,  hope you would not mind : )

Please also correct me if I misinterpreted your code.

Would you prefer me to merge this now to start some quick experiments when you addressing this comments, or merge this after you have done so?

DonggeLiu (approved):
Wow, thanks for fixing this promptly!

experiment/evaluator.py
```diff
@@ -195,13 +205,78 @@ def check_target(self, ai_binary, target_path: str) -> Optional[Result]:
       traceback.print_exc()
       return None
 
-  def _parse_libfuzzer_logs(self, log_handle) -> tuple[int, int, bool]:
-    """Parse libFuzzer logs."""
+  def _parse_stacks_from_libfuzzer_logs(self, lines: list[str]) -> list[str]:
```
DonggeLiu, line 208: This return type should be `list[list[str]]` because `stacks` is a `list` of `stack` while each `stack` is a `list` of `str`.

experiment/evaluator.py
```diff
@@ -195,13 +205,78 @@ def check_target(self, ai_binary, target_path: str) -> Optional[Result]:
...
+          lastround = roundno
+          if 'INITED' in line:
+            initcov = int(line.split('cov: ')[1].split(' ft:')[0])
+          elif 'DONE' in line:
+            donecov = int(line.split('cov: ')[1].split(' ft:')[0])
+
+    return initcov, donecov, lastround
+
+  def _stack_func_is_of_testing_project(self, stack_frame: str) -> bool:
+    return bool(CRASH_STACK_WITH_SOURCE_INFO.match(stack_frame)) and (
```
DonggeLiu, line 258: This incurred the following error for the same reason as [the incorrect type above](https://github.com/google/oss-fuzz-gen/pull/187/files#r1546206543).
```python
    return bool(CRASH_STACK_WITH_SOURCE_INFO.match(stack_frame)) and (
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: expected string or bytes-like object, got 'list'
```

Could you please fix this in a new PR (and branch)? Thanks!

experiment/evaluator.py
```diff
@@ -217,7 +292,33 @@ def _parse_libfuzzer_logs(self, log_handle) -> tuple[int, int, bool]:
...
+        # No cov line has been identified or only INITED round has been passed.
+        # This is very likely the false positive cases.
+        return cov_pcs, total_pcs, False, True, 'FP_CRASH_NEAR_INIT'
+
+      # FP case 2: 1st func of the 1st thread stack is in driver.
+      crash_stacks = self._parse_stacks_from_libfuzzer_logs(lines)
+      for stack_frame in crash_stacks[:1]:
+        if self._stack_func_is_of_testing_project(stack_frame):
+          if 'LLVMFuzzerTestOneInput' in stack_frame:
+            return cov_pcs, total_pcs, False, True, 'FP_CRASH_IN_DRIVER'
```
oliverchang, line 312: Could we replace these with Python enums? https://docs.python.org/3/library/enum.html 

@DonggeLiu

experiment/evaluator.py
```diff
@@ -195,13 +205,78 @@ def check_target(self, ai_binary, target_path: str) -> Optional[Result]:
...
       return None
 
-  def _parse_libfuzzer_logs(self, log_handle) -> tuple[int, int, bool]:
-    """Parse libFuzzer logs."""
+  def _parse_stacks_from_libfuzzer_logs(self, lines: list[str]) -> list[str]:
+    """Parse stack traces from libFuzzer logs."""
+    # There can have over one thread stack in a log.
+    stacks = []
+
+    # A stack -> a sequence of stack frame lines.
```
oliverchang, line 213: @DonggeLiu  we should add a TODO here to use our stack parsing from ClusterFuzz here instead, which is more tested and can do things like skipping inlined/common C lib functions.

oliverchang (commented):
thank you so much for this contribution!

experiment/evaluator.py
DonggeLiu: Ah, yes good point. Thanks!

experiment/evaluator.py
DonggeLiu: Yep! I presume this is the one?
https://github.com/google/clusterfuzz/blob/698d0bc6dba09b7af8d5acf3fe0941e909e43d57/src/clusterfuzz/_internal/crash_analysis/stack_parsing/stack_parser.py

@happy-qop Would you mind helping me add this in your new PR, given it is ready? Thanks!
```
# TODO (dongge): Use stack parsing from ClusterFuzz.
```

DonggeLiu: Amazing, thanks sooooo much!

DonggeLiu: How about addressing this under the same PR #198?
For example, having a new branch of the current [`stacktrace-fixing-exp`](https://github.com/google/oss-fuzz-gen/tree/stacktrace-fixing-exp) and merge it to #198 once it is ready.
Then we can run further experiments from there.

PR #190, opened 2024-03-28
Parse clang diagnostic format text to regex

cjx10:
TODO:

- [x] lint
- [x] docstring
- [ ] use regex to extract error in build log


DonggeLiu (commented):
Closing this as discussed.
This could be useful in the long term if we need very fine-grained and formal error names, but it's a bit too heavy-weighted for us at the moment. Let's come back to this if we need to summarise the formal regexes of each error category to minimize false positives/negatives when capturing errors.

PR #191, opened 2024-03-28
Surface crash category

DonggeLiu:
Show crash type in HTML reports and JSON summary.

### DIFF
--- a/report/templates/benchmark.html
+++ b/report/templates/benchmark.html
@@ -24,6 +24,8 @@
         <th>Status</th>
         <th>Builds</th>
         <th>Crashes</th>
+        <th>Bug</th>
+        <th>Crash reason</th>
         <th>Coverage</th>
         <th>Line coverage diff</th>
     </tr>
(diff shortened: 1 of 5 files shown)


DonggeLiu:
JOB: https://console.cloud.google.com/kubernetes/job/us-central1-c/llm-experiment/default/ofg-pr-191-dg
REPORT: https://llm-exp.oss-fuzz.com/Result-reports/ofg-pr/2024-03-28-191-dg-comparison/index.html
BUCKET: https://console.cloud.google.com/storage/browser/oss-fuzz-gcb-experiment-run-logs/Result-reports/ofg-pr/2024-03-28-191-dg-comparison
BUCKET GS: `gs://oss-fuzz-gcb-experiment-run-logs/Result-reports/ofg-pr/2024-03-28-191-dg-comparison`

DonggeLiu:
Hi @happy-qop, I made a change ([db22ff0](https://github.com/google/oss-fuzz-gen/pull/191/commits/db22ff028e59d53774bfce85a69429fa1338f917)) to keep the def of 'crash' consistent with the past, i.e., crash includes both bug in fuzz target (fp) and bug in the project under test (tp).
The main goal of this is to ensure the report also highlights bugs in fuzz targets so that we can develop new ways to mitigate/avoid them in the future.

I added another column in the report to show the real ones.
Here is the preview of the table before the change:
image

Take the 2nd row as an example, after this change, if the cell in 'Bug' column is zero, but the one under 'Crash' is not, then we know we should improve fuzz target generation.


We will ensure you have access to them as soon as everyone returns from holiday.

PR #196, opened 2024-03-29
Allow using ai-binary for code generation only, not code fixing

DonggeLiu:
Testing a model tuned for code generation.


mihaimaruseac (changes requested):
Should we instead expose this to a CLI argument and thread it from there to here?

DonggeLiu:
> Should we instead expose this to a CLI argument and thread it from there to here?

Yes, that's the 'improve the code later' part : )
I was running some local experiments with this now and forgot to turn this into a draft.

PR #199, opened 2024-04-02
Fix filter bug & update doc

happy-qop:
1. add todo for use clusterfuzz stacktrace parsing;
2. fix bug of filtering;
3. use enum for types of error fuzz drivers;
4. update the doc of NUM_SAMPLES;


DonggeLiu (approved):
Thanks!

run_one_experiment.py
```diff
@@ -40,9 +40,12 @@
...
 # Default LLM hyper-parameters.
-# WARN: Avoid large NUM_SAMPLES in highly parallelized local experiment.
-# NUM_SAMPLES controls the number of LLM responses per query, which may exceed
-# your LLM's limit on query-per-second.
+# - #182 shows Gemini returns NUM_SAMPLES independent responses via repeated
+#  queries, which generally performs better than top-k responses from one
+#  query [1].
+# WARN: please set the value according to your own LLM's query-per-second
+#     limit, especially in highly parallelized local experiment.
+# [1] TODO(@happy-qop): Update the link.
```
DonggeLiu, line 48: Ah, sorry, I should have been clearer. Would it be better to move `[1]` right after your new comment? E.g.,
```python
# #182 shows Gemini returns NUM_SAMPLES independent responses via repeated
#  queries, which generally performs better than top-k responses from one
#  query [1].
# [1] TODO(@happy-qop): Update the link.
# WARN: Avoid large NUM_SAMPLES in highly parallelized local experiments. 
# It controls the number of LLM responses per prompt, which may exceed your 
# LLM's limit on query-per-second.
```

Sorry about the confusion!

DonggeLiu (approved):

run_one_experiment.py
DonggeLiu: Thanks, merged!

PR #198, opened 2024-04-02
[DO NOT MERGE] Experiment fixing stacktrace parsing

DonggeLiu:
Related code fixing and improvements:
1. https://github.com/google/oss-fuzz-gen/pull/187#discussion_r1546206543
2. https://github.com/google/oss-fuzz-gen/pull/187#discussion_r1546208485
3. https://github.com/google/oss-fuzz-gen/pull/187#discussion_r1547062065
4. https://github.com/google/oss-fuzz-gen/pull/187#discussion_r1547074956

PR #204, opened 2024-04-04
The fix loop now include drivers have fuzz err

happy-qop:
# Overview

This is an initial version where only basic information (error type, crash stacks, and crash summarization from ASAN & libfuzzer) about the driver runtime errors are used for fixing drivers with false positives or have 0 coverage increasement. 

Its main purpose is to propose a base implementation which both lets the fix loop handles drivers with fuzz errors (only build errors are fixed in previous version) and lets the fix prompts be updated accordingly.

# About Fix Prompts

Fix prompts of build errors are kept intact. For runtime errors, please see diff in `fixer_problem.txt` and methods (`extract_symptom`, `_get_error_desc` -> `ERROR_SUMMARY`,  `_get_error_detail` -> `ERROR_MESSAGES`) of class `DriverFuzzErr`.

# Main Changes in Implementation

The main changes are listed as following for easier understanding of the code changes:
- rewrite the fix loop in `evaluator.py`
- DriverFuzzErr is no longer an `Enum` but a class with methods preparing fuzz error related information. These info are passed to fix prompts
- update one line of the fix prompt `fixer_problem.txt` to support the fix of fuzz errors
- add iteration suffix to `run log` to handle the case that `fuzz/build err <=> before/after fix`

I've tested the code in small scale locally and fixed some bugs, but more tests are needed.

# Some thoughts/plans

More contextual information for each type of incorrect driver can be added in the future. For example, for a false positive crash, we may highlight the code of exact error line, and provide the stack trace with partial project source code, which is similar as the information provided by a command line debugger. 

Example illustrating richer contextual info (NOT IMPLEMENTED):
```
The above C code can be built successfully but will crash (ASAN-heap-buffer-overflow on address 0x602000000111 at pc 0x0000007024e0 bp 0x7ffec5e4b850 sp 0x7ffec5e4b848).

The crash line is:
       result = JS_Eval(ctx, (const char *)Data, Size, filename, eval_flags);

The frames of the stack are:

#0 0x7024df in next_token /src/quickjs/quickjs.c:20678:9
#1 0x61ba55 in js_parse_program /src/quickjs/quickjs.c:33463:9
#2 0x61ba55 in __JS_EvalInternal /src/quickjs/quickjs.c:33654:11
#3 0x5f7744 in JS_EvalInternal /src/quickjs/quickjs.c:33694:12
#4 0x5f7744 in JS_EvalThis /src/quickjs/quickjs.c:33725:11
#5 0x5f7744 in JS_Eval /src/quickjs/quickjs.c:33733:12
#6 0x56a041 in LLVMFuzzerTestOneInput /root/workspace/jarvis/targets/quickjs/c/para-iter-account-12-0/dummyfuzzer.c:122:18

The frame `next_token` related code:
20677   s->token.ptr = p;
20678   c = *p;
20679   switch(c) {

The frame `js_parse_program` related code:
33462  
33463   if (next_token(s))
33464   return -1;
...
```


experiment/evaluator.py
```diff
@@ -352,61 +427,80 @@ def do_check_target(self, ai_binary: str, target_path: str) -> Result:
...
+
     # TODO: Log build failure.
     # TODO: Log run success/failure.
 
-    # Loop to try and fix the compilation error using the LLM.
-    while not build_result.succeeded and llm_fix_count < LLM_FIX_LIMIT:
+    # Result variables of generation.
+    llm_fix_count = 0
+
+    while True:
```
DonggeLiu, line 437: This is a huge change, thanks!
Would you mind helping reduce the size of `while` loop, given it has many tasks? I like the comments you left, maybe we can separate the functionalities into new functions based on them, e.g.:
```python
while not driver_fuzz_err.has_err and llm_fix_count < LLM_FIX_LIMIT:
  cov_pcs, total_pcs, crashes, driver_fuzz_err = build_and_evaluate_fuzz_target()  # Everything belongs to your first comment.
  if driver_fuzz_err.has_err:
    fix_fuzz_target()  # Everything belongs to your 2nd comment. This fixes the fuzz target, the next iteration builds and evaluates it.


# Outside while
if driver_fuzz_err.has_err:
  log(failure info)
  return ...
log(success info)
return ...
```

Please let me know if I missed anything, or if you like to do this together (this will involve a lot code change indeed).
Thanks :)

llm_toolkit/prompt_builder.py
```diff
@@ -255,12 +261,18 @@ def _format_fixer_priming(self) -> Tuple[str, int]:
...
 
-  def _format_fixer_problem(self, raw_code: str, errors: list[str],
-                            priming_weight: int) -> str:
+  def _format_fixer_problem(self, raw_code: str, error_desc: Optional[str],
+                            errors: list[str], priming_weight: int) -> str:
     """Formats a problem for code fixer based on the template."""
     with open(self.fixer_problem_template_file) as f:
       problem = f.read().strip()
     problem = problem.replace('{CODE_TO_BE_FIXED}', raw_code)
+    if error_desc is None:
```
DonggeLiu, line 270: `if error_desc` and reverse the block content for simplicity.

llm_toolkit/prompt_builder.py
```diff
@@ -39,6 +39,9 @@
     [FDP_EXAMPLE_2_PROBLEM, FDP_EXAMPLE_2_SOLUTION],
 ]
 
+BUILD_ERROR_SUMMARY = 'The code has the following build issues:'
+FUZZ_ERROR_SUMMARY = 'The code can be built successfully but %s.'
```
DonggeLiu, line 43: `FUZZ_ERROR_SUMMARY = 'The code can build successfully but has a runtime issue: '`

llm_toolkit/prompt_builder.py
```diff
@@ -255,12 +261,18 @@ def _format_fixer_priming(self) -> Tuple[str, int]:
...
+                            errors: list[str], priming_weight: int) -> str:
     """Formats a problem for code fixer based on the template."""
     with open(self.fixer_problem_template_file) as f:
       problem = f.read().strip()
     problem = problem.replace('{CODE_TO_BE_FIXED}', raw_code)
+    if error_desc is None:
+      # Build error does not pass error desc.
+      error_summary = BUILD_ERROR_SUMMARY
+    else:
+      error_summary = FUZZ_ERROR_SUMMARY % (error_desc)
```
DonggeLiu, line 274: error_summary = FUZZ_ERROR_SUMMARY + error_desc

llm_toolkit/prompt_builder.py
```diff
@@ -289,6 +301,8 @@ def _format_fixer_problem(self, raw_code: str, errors: list[str],
 
     # Now, compose the problem part of the prompt
     error_message = '\n'.join(selected_errors)
+    if error_message.strip() == '':
```
DonggeLiu, line 304: Two suggestions:
1. I  am a bit inclined to avoid removing `<error>` tag even if `error_message.strip() == ''`, because:
    1. It avoids assuming the prompt must be `'<error>\n{ERROR_MESSAGES}\n</error>\n'`.
    2. `error_message.strip() == ''` is not expected, so we should a) log a warning, and b) leave `<error>\n\n</error>\n` in the report so that we can easily find it and investigate it. It can be found by `helper/result_string_search.py`.
3. If you'd prefer to keep the code, we can simplify it a bit:
```python
if error_message.strip():
  return problem.replace('{ERROR_MESSAGES}', error_message)
return problem.replace('<error>\n{ERROR_MESSAGES}\n</error>\n', '')
```

DonggeLiu: Funny GitHub does not escape strings in our comments, causing `'<error>'` becomes `''` without the `s on this page.

DonggeLiu (commented):
Sorry that it took a bit longer than expected.
Feel free to take your time on this and have a nice weekend!

experiment/fuzz_target_error.py
```diff
@@ -0,0 +1,110 @@
...
+              'by the fuzz target.')
+    if self.type == self.FP_OOM:
+      return ('Out-of-memory error detected, suggesting memory leak in the'
+              ' fuzz target.')
+    if self.type == self.FP_TIMEOUT:
+      return ('Fuzz target timed out at runtime, indicating its usage for '
+              'the function under test is incorrect or unrobust.')
+    if self.type == self.NO_COV_INCREASE:
+      # TODO(dongge): Append the implementation of the function under test.
+      return ('Low code coverage, indicating the fuzz target ineffectively '
```
DonggeLiu, line 87: I'm Sorry I missed this earlier: Would this be 'No code coverage' for better consistency with the case name `NO_COV_INCREASE`?
I presume this only happens if a fuzz target's code coverage is exactly 0 instead of under a threshold.

llm_toolkit/prompt_builder.py
DonggeLiu: Please correct me if I misinterpreted your message. There are two cases where `error_message` can be empty:
1. Expected case: Non-coverage driver, which has no additional error message.
2. Unexpected case: Error log parsing failure, in which case the builder/runner has an error log, but our incorrect error parsing code failed to parse the error message from it.

My understanding is that exposing the 2nd case to us is important for debugging OSS-Fuzz-Gen, particularly when running large-scale experiments.
I can think of two potential solutions for this.
Solution 1:
1. If `error_message` is empty and`ERROR_SUMMARY` is about no coverage, then we also remove the `<error>` tag pair.
2. If `error_message` is empty and`ERROR_SUMMARY` is not about no coverage, we log a warning and leave the `<error>` tag pair.
3. If `error_message` is not empty, we replace `ERROR_MESSAGE` with `error_message` as normal.

Solution 2:
We handcraft an error message specifically for 'no coverage' and use it in step 2.


Re. 1.i, to avoid hard-coding the prompt content, we can remove them separately, e.g.: 
`problem.replace('<error>', '')`
`problem.replace('{ERROR_MESSAGES}', '')`
`problem.replace('<\error>', '')`

Then we remove the trailing empty lines in `problem` to avoid leaving `\n..\n` at the end.

Would this be a good idea?

DonggeLiu: Thanks for addressing this so quickly!

llm_toolkit/prompt_builder.py
```diff
@@ -302,7 +304,17 @@ def _format_fixer_problem(self, raw_code: str, error_desc: Optional[str],
...
     if error_message.strip() == '':
-      return problem.replace('<error>\n{ERROR_MESSAGES}\n</error>\n', '')
+      if SemanticError.is_no_cov_increase_err(error_desc):
+        # NO_COV_INCREASE error has empty error message.
+        return problem.replace('<error>\n', '')\
+                      .replace('{ERROR_MESSAGES}\n', '')\
+                      .replace('</error>\n', '')
+
+      # Log error types that should have error message.
+      logging.info(
```
DonggeLiu, line 314: `logging.warning` because this likely indicates an error in the code.

llm_toolkit/prompt_builder.py
```diff
@@ -302,7 +304,17 @@ def _format_fixer_problem(self, raw_code: str, error_desc: Optional[str],
...
+          'Unexpected empty error message in fix prompt for error_desc: %s',
+          str(error_desc))
+
     return problem.replace('{ERROR_MESSAGES}', error_message)
```
DonggeLiu, line 318: nit: Reordering the logic a bit to reduce indentation levels:
```python
    # Now, compose the problem part of the prompt
    error_message = '\n'.join(selected_errors)
    if error_message.strip():
      return problem.replace('{ERROR_MESSAGES}', error_message)

    # Expecting empty error message for NO_COV_INCREASE.
    if SemanticError.is_no_cov_increase_err(error_desc):
      return problem.replace('<error>\n', '')\
                      .replace('{ERROR_MESSAGES}\n', '')\
                      .replace('</error>\n', '')

    # Log warning for an unexpected empty error message.
    logging.warning(
        'Unexpected empty error message in fix prompt for error_desc: %s',
        str(error_desc))
    return problem.replace('{ERROR_MESSAGES}', error_message)
```

DonggeLiu:
Thanks, @happy-qop, for the prompt responses!
I left a few more notes for discussion; please let me know your opinions, thanks!

experiment/fuzz_target_error.py
```diff
@@ -0,0 +1,117 @@
...
+"""
+Helper class for fuzz target semantic errors.
+"""
+import re
+from typing import Optional
+
+
+class SemanticError:
+  """Fuzz target semantic errors."""
+  NO_SEMANTIC_ERR = 'NO_SEMANTIC_ERR'
```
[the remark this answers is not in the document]
DonggeLiu: Thanks! Happy to discuss more via our meeting.

DonggeLiu (approved):

DonggeLiu:
Thanks again for addressing the comments so quickly!
Given merging `SematicCheckResult` into `RunResult` is the only thing left and may take a few days to review and implement, would you mind me merging this into #198 so that we can run another experiment in the mean time?

Then merging the two `*Result` classes can be done on a new branch based on `stacktrace-fixing-exp` from #198, if that's convenient for you : )

PR #205, opened 2024-04-08
Add sample Python auto generation

DavidKorczynski:
Sample for auto-generating an OSS-Fuzz project for a given Python project.

This differs a bit from the existing set up. The approach in this PR relies on cloning the Python repository within an OSS-Fuzz base-builder image. Within this image Fuzz Introspector is cloned, and a fuzz introspector analysis purely based on static analysis is performed, to extract details about the Python library under analysis.

It's a bit raw at this stage, and can likely be better integrated into OSS-Fuzz-gen. However, I'm not 100% sure what the smartest steps are here, so am sharing in case there are opinions. It may be smarter to operate on two tracks in parallel and merge them later on once it's better known what works/doesn't work.

### DIFF
--- a/experiment/builder_runner.py
+++ b/experiment/builder_runner.py
@@ -119,6 +119,13 @@ class BuilderRunner:
       return False
     return True
 
+  def build_and_run_python(self, generated_project: str, target_path: str):
+    build_result = BuildResult()
+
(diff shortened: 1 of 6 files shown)


defihook (commented):
use fstrings here to be consistent?

PR #206, opened 2024-04-09
Fix missing build results in case of code fixing failure

happy-qop:
Fix the bug where the build succ rate is not correctly collected when build fails.


DonggeLiu (approved):

DonggeLiu:
BTW, it could be clearer if the semantic check reports 'N/A' in case of build failure, because we only check semantic if a fuzz target can run.

We can do this later if we merge `SemanticCheckResult` into `RunResult`.

Issue #143, opened 2024-03-07
Less ambiguous and more informative coverage data in report


DonggeLiu:
Current report (e.g., [this one](https://llm-exp.oss-fuzz.com/Result-reports/scheduled/2024-03-06-weekly-all/sort.html)) has two main ambiguities regarding coverage data:
1. The coverage column is [**'program counter coverage'**](https://github.com/google/oss-fuzz-gen/blob/49f0c9d6d10d620d9a87e40e056ad03c217b2596/experiment/evaluator.py#L285C36-L285C45). Some readers may assume it is line coverage because the column next to it is 'Line Coverage Diff'.
    * Change the title to **'program counter coverage'** (pc coverage is more fine-grained than line coverage).
2. The link in the 'Line Coverage Diff' is [the line coverage of the fuzz target](https://github.com/google/oss-fuzz-gen/blob/main/experiment/builder_runner.py#L451), not the *diff line coverage* against OSS-Fuzz data.
    * Insert another column for fuzz target line coverage with this link.
    * Change the hyperlink of line coverage diff data to the [`subtracted covreport`](https://github.com/google/oss-fuzz-gen/blob/49f0c9d6d10d620d9a87e40e056ad03c217b2596/experiment/evaluator.py#L291), which can be written to the result dir with [`to_file()`](https://github.com/google/oss-fuzz-gen/blob/49f0c9d6d10d620d9a87e40e056ad03c217b2596/experiment/textcov.py#L175C7-L175C14)

DonggeLiu:
Not related to report UI but related to numbers:
On the index page, the line coverage should count the *union* of lines covered by all new fuzz targets, not the maximum of them. That is:
```python
line_coverage = union(target.line_covered for target in all_LLM_generted_fuzz_targets)
```

Same for line coverage diff.

Issue #105, opened 2024-02-13
Add an `information.json` in each benchmark directory


DonggeLiu:
The information file shows various name/representations of the benchmark used in the experiments, e.g.,:
1. Function signature
2. Local directory name
3. Docker image name
4. Could experiment tag/name
5. Temp directories (OSS-Fuzz)

PR #213, opened 2024-04-10
Improve the usability of the experiment reports

erfanio:
* The tables are much prettier
* The index.html table has sortable columns
* The index.html table has an index column
* Fix a bug in the signature matching algorithm where it returned early on prefix matches.
* Prevent the licence to show up in the HTML (often multiple times).

Preview: 
image

### DIFF
--- a/report/web.py
+++ b/report/web.py
@@ -83,7 +83,9 @@ class Benchmark:
           return function_signature
         if function_name.lower().startswith(self.function.lower()):
           if matched_prefix_signature:
-            logging.warning(f'Multiple substring matches found when looking for function name {function_name}')
+            logging.warning(
+                'Multiple substring matches found when looking for function '
+                'name %s', function_name)
           matched_prefix_signature = function_signature
 
     return matched_prefix_signature
--- a/report/web.py
+++ b/report/web.py
@@ -82,6 +82,8 @@ class Benchmark:
         if function_name.lower() == self.function.lower():
           return function_signature
         if function_name.lower().startswith(self.function.lower()):
+          if matched_prefix_signature:
+            logging.warning(f'Multiple substring matches found when looking for function name {function_name}')
           matched_prefix_signature = function_signature
 
     return matched_prefix_signature
--- a/report/web.py
+++ b/report/web.py
@@ -50,32 +50,41 @@ class Benchmark:
   status: str
   result: run_one_experiment.AggregatedResult
   signature: str = ''
+  project: str = ''
+  function: str = ''
 
   def __post_init__(self):
-    self.signature = self.find_signature(self.id) or self.id
+    self.project = '-'.join(self.id.split('-')[1:-1])
+    self.function = self.id.split('-')[-1]
+    self.signature = self.find_signature() or self.id
 
-  @staticmethod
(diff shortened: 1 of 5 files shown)


DonggeLiu:
Luckily, there does not seem to be a merge conflict : )
I hope you don't mind me taking the liberty to rebase this PR. I was trying to show and fix the merge conflict now, but it seems OK.

DonggeLiu:
A related issues: #143
Happy to discuss more on this tmr

DonggeLiu:
Running a PR-experiment to show the new look : )
JOB: https://console.cloud.google.com/kubernetes/job/us-central1-c/llm-experiment/default/ofg-pr-213-dg
REPORT: https://llm-exp.oss-fuzz.com/Result-reports/ofg-pr/2024-04-11-213-dg-comparison/index.html
BUCKET: https://console.cloud.google.com/storage/browser/oss-fuzz-gcb-experiment-run-logs/Result-reports/ofg-pr/2024-04-11-213-dg-comparison
BUCKET GS: `gs://oss-fuzz-gcb-experiment-run-logs/Result-reports/ofg-pr/2024-04-11-213-dg-comparison`

DonggeLiu:
@erfanio Would you mind me merging this, or is it still WIP?
There are two experiments taking very long but they do not affect the new pages looking great.

erfanio:
Feel free to merge. I had forgotten to ask for review.

report/web.py
```diff
@@ -50,32 +50,41 @@ class Benchmark:
...
+
+        # Best match is a full match, but sometimes the result directory only
+        # has the first n characters of a long function name so a full match is
+        # not possible.
+        # To avoid returning early on a prefix match when there is a full match
+        # farther down the list, we only return the prefix match at the end.
+        if function_name.lower() == self.function.lower():
+          return function_signature
+        if function_name.lower().startswith(self.function.lower()):
+          matched_prefix_signature = function_signature
```
DonggeLiu, line 85: Thanks, this is clearer and more robust. 
1. Do you happen to notice any project/function only has a partial match? I don't recall which one it was.
2. Given now we handle this more carefully, shall we log a warning if there are multiple partial matches?

DonggeLiu (commented):
Thanks! Two quick questions:

report/web.py
erfanio: https://github.com/google/oss-fuzz-gen/blob/main/benchmark-sets/all/ada-url.yaml

has both `ada_can_parse` and `ada_can_parse_with_base` . I only noticed it when I was looking for benchmarks with the same signature. These two were showing the same signature in the report and I tracked it down to this.

erfanio: A lot of the FI generated ones like immer are too long so the directories only have a substring.

I think this is the bit where long function names are truncated https://github.com/google/oss-fuzz-gen/blob/main/experiment/benchmark.py#L91

DonggeLiu: Ah yes you are right.

erfanio: I think this is further motivation for #105

DonggeLiu: Yep!

DonggeLiu (approved):

PR #216, opened 2024-04-11
Add flow for generating OSS-Fuzz projects from a GH URL

DavidKorczynski:
Adds flow for building an OSS-Fuzz set up from scratch.

At the moment it's not integrated into the oss-fuzz-gen workflow. The logic of this is based on running within the OSS-Fuzz containers entirely (with exception of a wrapper that triggers the run), which makes some things a bit easier e.g. developing inside the environment where things run and having easy access to an introspector compiler as well as build utilities.


from-repo/c-cpp/README.md
```diff
@@ -0,0 +1,14 @@
+# Auto-build oss-fuzz projects from a GitHub repo URL
```
oliverchang, line 1: Can we put this all under `experimental`? 

We can put all the other language support there too. e.g. `experimental/java`

from-repo/c-cpp/README.md
```diff
@@ -0,0 +1,14 @@
...
+
+
+To run:
+
+```sh
+# prepare
+export OPENAI_API_KEY=your-api-key
```
oliverchang, line 8: Does this work with vertex ai as well? if so, we should add some setup steps for that also.

from-repo/c-cpp/build-generator.py
```diff
@@ -0,0 +1,1341 @@
+#!/usr/bin/env python3
```
oliverchang, line 1: Can you add some docs (e.g. a markdown doc) explaining the high level design ideas / flow for how this works?

from-repo/c-cpp/README.md
DavidKorczynski: yes, done

from-repo/c-cpp/build-generator.py
DavidKorczynski: Yes, done

from-repo/c-cpp/README.md
DavidKorczynski: Not yet. I need to adjust a few things because then harness generation happens inside a Docker container and there's no API key. I don't think this is a blocker as such but need to find out how to easily handle vertex calls within a container environment (i.e. no gcloud login).

Is this a blocker for landing?

oliverchang: not a blocker for this one, but something we definitely need to address in a subsequent PR.

oliverchang: can you add a TODO? one way is to just use a Google cloud service account, and set GOOGLE_APPLICATION_CREDENTIALS: https://cloud.google.com/docs/authentication/application-default-credentials#GAC

experimental/c-cpp/runner.py
```diff
@@ -0,0 +1,302 @@
...
+  """Launch auto-gen analysis within OSS-Fuzz container."""
+
+  initiator_cmd = 'python3 /src/build-generator.py %s -o %s' % (github_url,
+                                                                outdir)
+  if disable_autofuzz:
+    initiator_cmd += ' --disable-fuzzgen'
+  initiator_cmd += ' --targets-per-heuristic=%d' % (targets_per_heuristic)
+
+  cmd = [
+      "docker",
```
oliverchang, line 132: nit: can we use single quotes for all strings other than docstrings?

experimental/c-cpp/README.md
```diff
@@ -0,0 +1,83 @@
...
+The core part of approach in this module takes the following steps:
+
+1) An input in the form of a GitHub repo is given to `runner.py`.
+2) `runner.py` creates a temporary OSS-Fuzz project that contains
+  `build-generator.py`
+3) `runner.py` launches the OSS-Fuzz image of said project with
+  `build-generator.py` as entrypoint.
+
+`build-generator.py` will then proceed to carry out the following tasks within
+the Docker container:
```
oliverchang, line 15: does this mean the OSS-Fuzz builder Docker container?

experimental/c-cpp/build-generator.py
```diff
@@ -0,0 +1,1486 @@
...
+    // end of fuzzer contents
+
+    return 0;
+}"""
+
+
+############################################################
+#### Logic for auto building a given source code folder ####
+############################################################
+class AutoBuildContainer:
```
oliverchang, line 60: Can we separate the autobuild stuff out into a separate module?

experimental/c-cpp/README.md
```diff
@@ -0,0 +1,83 @@
...
+The algorithm takes the arguments:
+1) Target repository
+2) Number of functions to target
+
+This means the total amount of build/harness configurations to perform is:
+
+N x B x G where:
+
+- N is the number of functions to target
+- B is the number of build heuristics succeeding (max 10 or so at this point)
```
oliverchang, line 64: What's a build heuristic? something that figures out how to build a project? something that figures out how to build an arbitrary target against the project? 

My question nis why this isn't `B + N x G`. Is `B` specific to the target ?

experimental/c-cpp/build-generator.py
```diff
@@ -0,0 +1,1486 @@
...
+    build_command_includes = ""
+    for header_path_to_include in header_paths_to_include:
+      build_command_includes += "-I" + os.path.join(
+          self.test_dir, header_path_to_include) + " "
+
+    return headers_to_include, header_paths_to_include, build_command_includes
+
+  def run_prompt_and_get_fuzzer_source(self, prompt):
+    """Communicate to OpenAI prompt and extract harness source code."""
+    completion = client.chat.completions.create(model="gpt-3.5-turbo",
```
oliverchang, line 593: Can you add a TODO to consolidate this with our prompt calling code?

i.e. https://github.com/google/oss-fuzz-gen/blob/main/llm_toolkit/prompts.py and https://github.com/google/oss-fuzz-gen/blob/620dec55cab5c78badd66c1b23f3edeafd0cda58/llm_toolkit/models.py#L122

experimental/c-cpp/runner.py
DavidKorczynski: Yup, done.

experimental/c-cpp/README.md
DavidKorczynski: This is because the approach is as follows:

```sh
for b in successful_builds:
  for f in functions_to_target:
    for h in fuzz_gen_heuristics:
      evaluate_triplet(b,f,h)
```

The reason we have the outer loop is that each successful build may produce a different output. For example, we have observed multiple successful builds may produce different static archives. Successful build in this case is a build script that runs without erroring and producing some output artifact.

from-repo/c-cpp/README.md
DavidKorczynski: I'll see if I can get this in tomorrow before landing this

experimental/c-cpp/build-generator.py
DavidKorczynski: Yes, done.

experimental/c-cpp/build-generator.py
DavidKorczynski: Yes, done.

experimental/c-cpp/README.md
DavidKorczynski: Yup -- or more precisely it means the Docker image of a temporary project created that inherits from the base builder. This is the Dockerfile template we use for the temporary project: https://github.com/google/oss-fuzz-gen/pull/216/files#diff-86b63b45a6de7bc9bc127980ecfdc91f89646409b7d51ae960a9fe9d1730c97fR60-R70

experimental/c-cpp/README.md
oliverchang: Got it, thanks!

oliverchang (approved):

from-repo/c-cpp/README.md
DavidKorczynski: I've added this now and also adjusted the README

oliverchang (approved):

PR #230, opened 2024-04-26
Check errlog in experiments

cjx10:


experiment/evaluator.py
```diff
@@ -194,6 +194,14 @@ def create_ossfuzz_project(self, name: str, target_file: str) -> str:
     with open(os.path.join(generated_project_path, 'Dockerfile'), 'a') as f:
       f.write(f'\nCOPY {os.path.basename(target_file)} '
               f'{self.benchmark.target_path}\n')
+      # TODO(jimchoi): Remove this.
```
cjx10, line 197: Change made to JCC: google/oss-fuzz#11842

cjx10:
Special cases when get_jcc_errstr() cannot find error message but it should: 

`proftpd`: `cp tests/fuzzing/json_fuzzer.c /src/fuzzer.c`

DonggeLiu:
> `proftpd`: `cp tests/fuzzing/json_fuzzer.c /src/fuzzer.c`
> 
> In all other projects, get_jcc_errstr() can return desired error messages, especially linker errors, or there's build issue from the project source. (In other words, fuzz target is never touched by jcc and error messages from source are not returned mistakenly)

Thanks for documenting this.
I reckon if this is the only case, then it's probably not too important.


However, could you please try building the original fuzz target with JCC and see if that fails?
E.g., build its image in OSS-Fuzz, set JCC in the docker container, and `compile` to build the binaries.
This will confirm if the problem is from JCC, given `proftpd` can build without it:
https://oss-fuzz-build-logs.storage.googleapis.com/index.html#proftpd

cjx10:
> try building the original fuzz target with JCC

To clarify, `proftpd` built in oss-fuzz-gen, but we lost track of the fuzz target in logs because the file name changed.

For testing, in `oss-fuzz` the following 2 lines were added to `projects/{project_name}/Dockerfile`:
```
ENV CC=/usr/local/bin/clang-jcc
ENV CXX=/usr/local/bin/clang++-jcc
```
Then `python3 infra/helper.py shell {project_name}`, manually run `compile` in docker shell.

Below is a list of projects found not building in oss-fuzz-gen, and whether they build with jcc ([7c67544](https://github.com/google/oss-fuzz/commit/7c67544b9c8729d5964d10746753330d51c8df7d)) in oss-fuzz:

- firestore: build
- libidn2: fail
- libpsl: fail
- libtasn1: fail
- mdbtools: fail
- mosh: fail
- myanmar-tools: fail
- oatpp: build
- openvswitch: build
- piex: build
- proj4: fail
- protobuf-c: fail
- qpid-proton: build
- tarantool: fail

llm_toolkit/code_fixer.py
```diff
@@ -217,64 +217,165 @@ def remove_const_from_png_symbols(content: str) -> str:
...
+
+      # Potentially followed by the error block to return
+      # if fuzz target presents.
+      if temp_range[0] is None:
+        compile_target = get_fuzz_target_compile_output(
+            target_names,
+            command_match.group(1).split())
+        if compile_target:
+          target_names.append(compile_target)
+          temp_range[0] = i + 1  # +1 to exclude the command args line.
```
DonggeLiu, line 319: Could you avoid the nested conditions and/or simplify the logic here?
For example, by using `elif` and `if not command_match: continue`, etc.

DonggeLiu (changes requested):

cjx10 (commented):
Thanks so much for the detailed review Dongge :)

llm_toolkit/code_fixer.py
```diff
@@ -217,123 +217,221 @@ def remove_const_from_png_symbols(content: str) -> str:
...
 
-      # Matched a diag line but under an included file line,
-      # indicating the specific error in the included file,
-      if curr_state == state_include and err_src != src_file:
-        curr_block.append(line)
+      # If we were handling an include error block, this is part of the block if
+      # error source has been found, and it's now changed.
+      if handling_include and (not include_error_src_file or
+                               include_error_src_file == err_src):
+        include_error_src_file = err_src
```
DonggeLiu, line 413: I guess if `include_error_src_file == err_src`, we don't have to re-assign it?

llm_toolkit/code_fixer.py
```diff
@@ -217,123 +217,221 @@ def remove_const_from_png_symbols(content: str) -> str:
...
     if diag_match:
       err_src = diag_match.group(1)
       severity = diag_match.group(2)
 
-      # Matched a note diag line under another diag,
-      # giving help info to fix the previous error.
+      # New error block won't start with note error.
       if severity == 'note':
-        curr_block.append(line)
         continue
```
DonggeLiu, line 407: If we do not care about `note`s, could we exclude this case from the pattern above? E.g.,
```python
diag_pattern = re.compile(r'(\S*):\d+:\d+: (?!note)(.+): .+')
```

experiment/builder_runner.py
```diff
@@ -308,20 +308,36 @@ def build_and_run(self, generated_project: str, target_path: str,
...
       shutil.copyfile(
           os.path.join(get_build_artifact_dir(generated_project, "workspace"),
-                       'err.log'),
-          self.work_dirs.error_logs_target(benchmark_target_name, iteration))
+                       'err.log'), jcc_errlog_path)
     except FileNotFoundError as e:
       logging.error('Cannot get err.log for %s: %s', generated_project, e)
+      # Touch err.log in results folder to avoid FileNotFoundError when
+      # extracting errors.
+      open(jcc_errlog_path, 'x')
```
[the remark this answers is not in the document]
DonggeLiu: > Maybe add a magic string in err.log when err.log does not exist?

This can work but not preferred.
Intuitively, if `err.log` is not generated, then it should not exist.

>  creating an empty file is currently consistent with the behaviour of getting the build log and run log

Where do we create empty build log and run logs when they do not exist?

> Might help prevent increasing the complexity of build_and_run() workflow?

Not sure if this relates, but I suppose this will only add two lines in `code_fixer`?
>> Could we check os.path.isfile() at the beginning extract_error_message() instead?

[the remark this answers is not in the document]
cjx10: > Where do we create empty build log and run logs?

It's in cloud builder, we open the local file object before checking if the file blob on cloud exists
https://github.com/google/oss-fuzz-gen/blob/main/experiment/builder_runner.py#L645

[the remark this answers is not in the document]
DonggeLiu: Ah, I see; let's keep it, then.
Thanks!

DonggeLiu:
Thanks for addressing the comments, I have no more suggestions now.
Could you please run a PR experiment and triple-check the report that everything is intact?
Thanks

cjx10:
> Could you please run a PR experiment 

Many thanks Dongge :)
Will do after I double check the parsing is correct, since the code structure has changed quite a bit

DonggeLiu:
nit: 
>-n jim

nit:Could you please add the branch id to the name in the future?
This helps us identify the job/bucket_dir/report.
Thanks : )

cjx10:
> Could you please add the branch id

oops yeah sure

PR #231, opened 2024-04-26
Capturing `null-deref` as a semantic error 

DonggeLiu:
Consider `null-deref` as a semantic error instead of a bug:
1. It is likely due to inadequate parameter initialization or incorrect library usage in the fuzz target 
2. It is hardly exploitable.

### DIFF
--- a/experiment/builder_runner.py
+++ b/experiment/builder_runner.py
@@ -245,7 +245,13 @@ class BuilderRunner:
       symptom = SemanticCheckResult.extract_symptom(fuzzlog)
       crash_stacks = self._parse_stacks_from_libfuzzer_logs(lines)
 
-      # FP case 1: fuzz target crashes at init or first few rounds.
+      # FP case 1: Null-deref, normally indicating inadequate parameter
+      # initialization or wrong function usage.
+      if symptom == 'null-deref':
+        return cov_pcs, total_pcs, True, SemanticCheckResult(
+            SemanticCheckResult.NULL_DEREF, symptom, crash_stacks)
+
+      # FP case 2: fuzz target crashes at init or first few rounds.
       if lastround is None or lastround <= EARLY_FUZZING_ROUND_THRESHOLD:
         # No cov line has been identified or only INITED round has been passed.
         # This is very likely the false positive cases.
@@ -253,7 +259,7 @@ class BuilderRunner:
                SemanticCheckResult(SemanticCheckResult.FP_NEAR_INIT_CRASH,\
                              symptom, crash_stacks)
 
-      # FP case 2: 1st func of the 1st thread stack is in fuzz target.
+      # FP case 3: 1st func of the 1st thread stack is in fuzz target.
       if len(crash_stacks) > 0:
         first_stack = crash_stacks[0]
         # Check the first stack frame of the first stack only.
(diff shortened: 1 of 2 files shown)


DonggeLiu:
JOB: https://console.cloud.google.com/kubernetes/job/us-central1-c/llm-experiment/default/ofg-pr-231-dg
REPORT: https://llm-exp.oss-fuzz.com/Result-reports/ofg-pr/2024-04-26-231-dg-comparison/index.html
BUCKET: https://console.cloud.google.com/storage/browser/oss-fuzz-gcb-experiment-run-logs/Result-reports/ofg-pr/2024-04-26-231-dg-comparison
BUCKET GS: `gs://oss-fuzz-gcb-experiment-run-logs/Result-reports/ofg-pr/2024-04-26-231-dg-comparison`

experiment/fuzz_target_error.py
```diff
@@ -39,17 +41,24 @@ class SemanticCheckResult:
...
+    match = cls.SYMPTOM_SCARINESS.search(fuzzlog)
+    if match:
+      return match.group(1)
+
+    match = cls.SYMPTOM_ASAN.search(fuzzlog)
     if match:
       return f'ASAN-{match.group(0)}'
 
-    match = cls.SYMPTOM_LIBFUZZER.match(fuzzlog)
+    match = cls.SYMPTOM_LIBFUZZER.search(fuzzlog)
```
DonggeLiu, line 61: @happy-qop: We need to use `search()` instead of [`match()`](https://docs.python.org/3/library/re.html#re.search:~:text=or%20more%20characters-,at%20the%20beginning%20of%20string,-match%20the%20regular) here because the pattern is never the beginning of the `fuzzlog`. This is why our previous report missed the crash reason, e.g., [here](https://llm-exp.oss-fuzz.com/Result-reports/ofg-pr/2024-04-26-228-dg-comparison/sample/output-fribidi-fribidi_log2vis/08).

Now we can correctly capture the crash reason and use the corresponding text in the fixing prompt.
For example, this captured a signal and fixed it:
https://llm-exp.oss-fuzz.com/Result-reports/ofg-pr/2024-04-26-231-dg-comparison/sample/output-libusb-libusb_get_bos_descriptor/06

occia: thanks for identifying and fixing this!

experiment/builder_runner.py
```diff
@@ -245,12 +245,19 @@ def _parse_libfuzzer_logs(
...
+      # FP case 1: Common fuzz target errors.
+      # Null-deref, normally indicating inadequate parameter initialization or
+      # wrong function usage.
       if symptom == 'null-deref':
         return cov_pcs, total_pcs, True, SemanticCheckResult(
             SemanticCheckResult.NULL_DEREF, symptom, crash_stacks)
 
+      # Signal, normally indicating assertion failure due to inadequate
+      # parameter initialization or wrong function usage.
+      if symptom == 'signal':
```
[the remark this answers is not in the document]
DonggeLiu: Yes, I was matching the signal for simplicity, I am happy to refine this if it turns out to be too broad in [the latest full exp](https://llm-exp.oss-fuzz.com/Result-reports/scheduled/2024-04-27-weekly-all/), e.g., search for `assert` in the call stack.
We can capture this by downloading the result dir, [searching related cases](https://github.com/google/oss-fuzz-gen/blob/main/helper/result_string_search.py), and coming up with a more fine-grained pattern.

PR #232, opened 2024-04-26
Color results in reports for better readability

DonggeLiu:
In the benchmark page:
1. Use green if the value in the `Builds` column is `True`.
2. Use red if the value in the `Crashes` or `Bug` columns is `False`.

### DIFF
--- a/report/templates/benchmark.html
+++ b/report/templates/benchmark.html
@@ -39,9 +39,9 @@ td, th {
         <td><a href="../../sample/{{ benchmark|urlencode }}/{{ sample.id }}">{{ sample.id }}</a></li></td>
         <td>{{ sample.status }}</td>
         {% if sample.result %}
-        <td style="color: {{ sample.result.compiles ? 'green' : 'black' }}">{{ sample.result.compiles }}</td>
-        <td style="color: {{ sample.result.crashes ? 'red' : 'black' }}">{{ sample.result.crashes }}</td>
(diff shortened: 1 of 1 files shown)


DonggeLiu:
JOB: https://console.cloud.google.com/kubernetes/job/us-central1-c/llm-experiment/default/ofg-pr-232-dg
REPORT: https://llm-exp.oss-fuzz.com/Result-reports/ofg-pr/2024-04-26-232-dg-comparison/index.html
BUCKET: https://console.cloud.google.com/storage/browser/oss-fuzz-gcb-experiment-run-logs/Result-reports/ofg-pr/2024-04-26-232-dg-comparison
BUCKET GS: `gs://oss-fuzz-gcb-experiment-run-logs/Result-reports/ofg-pr/2024-04-26-232-dg-comparison`

DonggeLiu:
TBR.

DonggeLiu:
> nit: I generally prefer setting background colour for things like this. Changing the text colour often means the text no longer has enough contrast to be easily readable. Also as someone who's red/green colour blind it's hard to distinguish text colour whereas a more subtle background colour with a stronger hue can be easily distinguished.

That's a really good point, thanks!

> I really like the old material colours from https://m2.material.io/design/color/the-color-system.html#tools-for-picking-colors
> 
> You could use these background colours for example: #C8E6C9 for green, #FFCDD2 for red

Thanks for the recommendations, I will use them before our next full exp.

PR #236, opened 2024-04-30
Add support for JVM based projects

arthurscchan:
This PR attempts to add support in the base logic for JVM based projects.

PR #242, opened 2024-05-02
Make it possible to create benchmarks during experimentation

DavidKorczynski:
This makes it possible to run experiments without having to generate benchmarks prior, and also makes it possible select one or more benchmarks to generate for X amount of projects.

Example run:

```sh
./run_all_experiments.py \
  --model=gpt-3.5-turbo -g low-cov-with-fuzz-keyword,far-reach-low-coverage -p htslib -m 6
```

Run for multiple projects:

```sh
./run_all_experiments.py \
  --model=gpt-3.5-turbo -g low-cov-with-fuzz-keyword,far-reach-low-coverage -p htslib,tinyxml2 -m 6
```

### DIFF
--- a/data_prep/introspector.py
+++ b/data_prep/introspector.py
@@ -35,7 +35,7 @@ from experiment import oss_fuzz_checkout
 
 T = TypeVar('T', str, list, dict)  # Generic type.
 
-TIMEOUT = 10
+TIMEOUT = 45
 MAX_RETRY = 5
 DEFAULT_INTROSPECTOR_ENDPOINT = 'https://introspector.oss-fuzz.com/api'
 INTROSPECTOR_ENDPOINT = ''
@@ -336,12 +336,21 @@ def _group_function_params(param_types: list[str],
 
 
 def populate_benchmarks_using_introspector(project: str, language: str,
-                                           limit: int, target_oracle: str):
+                                           limit: int, target_oracles: List[str]):
   """Populates benchmark YAML files from the data from FuzzIntrospector."""
-  functions = query_introspector_for_targets(project, target_oracle)
+
+  functions = []
+  for target_oracle in target_oracles:
+    logging.info(f'Extracting functions using oracle {target_oracle}')
+    tmp_functions = query_introspector_for_targets(project, target_oracle)
+
+    # Limit the amount of functions from each oracle.
+    functions += tmp_functions[:limit]
+
   if not functions:
-    logging.error('No unreached functions found')
+    logging.error('No functions found using the oracles: {target_oracles}')
     return []
+
 
   filenames = [
       os.path.basename(function['function_filename']) for function in functions
@@ -385,8 +394,9 @@ def populate_benchmarks_using_introspector(project: str, language: str,
                                target_name,
                                function_dict=function))
 
-    if len(potential_benchmarks) >= limit:
+    if len(potential_benchmarks) >= (limit * len(target_oracles)):
       break
+  print("Length of potential targets: %d"%(len(potential_benchmarks)))
 
   return potential_benchmarks
 
@@ -548,7 +558,7 @@ if __name__ == '__main__':
   benchmarks = populate_benchmarks_using_introspector(args.project,
                                                       cur_project_language,
                                                       args.max_functions,
-                                                      args.target_oracle)
+                                                      [args.target_oracle])
   if benchmarks:
     benchmarklib.Benchmark.to_yaml(benchmarks, args.out)
   else:
--- a/data_prep/introspector.py
+++ b/data_prep/introspector.py
@@ -23,7 +23,7 @@ import re
 import subprocess
 import sys
 import time
-from typing import Dict, List, Optional, TypeVar
+from typing import Any, Dict, List, Optional, TypeVar
 from urllib.parse import urlencode
 
 import requests
@@ -46,6 +46,16 @@ INTROSPECTOR_SOURCE = ''
 INTROSPECTOR_XREF = ''
 INTROSPECTOR_TYPE = ''
 INTROSPECTOR_FUNC_SIG = ''
+
+
+def get_oracle_dict() -> Dict[str, Any]:
+  """Returns the oracles available to identify targets."""
+  # Do this in a function to allow for forward-declaration of functions below.
+  oracle_dict = {
+      'far-reach-low-coverage': get_unreached_functions,
+      'low-cov-with-fuzz-keyword': query_introspector_for_keyword_targets
+  }
+  return oracle_dict
 
 
 def set_introspector_endpoints(endpoint):
@@ -155,11 +165,7 @@ def query_introspector_for_keyword_targets(project: str) -> list[dict]:
 
 def query_introspector_for_targets(project, target_oracle) -> list[Dict]:
   """Queries introspector for target functions."""
-  oracle_dict = {
-      'far-reach-low-coverage': get_unreached_functions,
-      'low-cov-with-fuzz-keyword': query_introspector_for_keyword_targets
-  }
-  query_func = oracle_dict.get(target_oracle, None)
+  query_func = get_oracle_dict().get(target_oracle, None)
   if not query_func:
     logging.error('No such oracle "%s"', target_oracle)
     sys.exit(1)
--- a/data_prep/introspector.py
+++ b/data_prep/introspector.py
@@ -336,12 +336,13 @@ def _group_function_params(param_types: list[str],
 
 
 def populate_benchmarks_using_introspector(project: str, language: str,
-                                           limit: int, target_oracles: List[str]):
+                                           limit: int,
+                                           target_oracles: List[str]):
   """Populates benchmark YAML files from the data from FuzzIntrospector."""
 
   functions = []
   for target_oracle in target_oracles:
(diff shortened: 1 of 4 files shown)


DavidKorczynski:
This will also be useful for https://github.com/google/oss-fuzz-gen/pull/236 where we're currently observing it will likely be useful to have some java-specific APIs

run_all_experiments.py
```diff
@@ -162,6 +198,19 @@ def parse_args() -> argparse.Namespace:
                       '--introspector-endpoint',
                       type=str,
                       default=introspector.DEFAULT_INTROSPECTOR_ENDPOINT)
+  parser.add_argument('-g',
+                      '--generate-benchmarks',
+                      help='Benchmarks to create',
```
oliverchang, line 203: Can you better document this in the help string? 

This takes in a comma separated list of  heuristics / oracles? What are the allowed values?

run_all_experiments.py
```diff
@@ -162,6 +198,19 @@ def parse_args() -> argparse.Namespace:
...
+                      type=str)
+  parser.add_argument('-p',
+                      '--projects',
+                      help='Projects to target, comma separated string.',
+                      type=str)
+  parser.add_argument('-m',
+                      '--max-benchmark-targets',
```
oliverchang, line 210: same here: `--generate-benchmarks-max` would be more consistent.

run_all_experiments.py
```diff
@@ -63,6 +77,28 @@ def get_experiment_configs(
     print(f'A benchmark yaml file ({args.benchmark_yaml}) is provided. '
           f'Will use it and ignore the files in {args.benchmarks_directory}.')
     benchmark_yamls = [args.benchmark_yaml]
+  elif args.generate_benchmarks:
+    logging.info('Generating benchmarks.')
+    outdir = get_next_generated_benchmarks_dir()
+    logging.info('Setting benchmark directory to %s.', outdir)
+    os.makedirs(outdir, exist_ok=True)
+    args.benchmarks_directory = outdir
+    benchmarks_to_generate = [
```
oliverchang, line 86: this variable name seems a bit confusing to me. Would it be better to call this `benchmark_oracles` instead?

data_prep/introspector.py
```diff
@@ -336,11 +336,20 @@ def _group_function_params(param_types: list[str],
...
+  for target_oracle in target_oracles:
+    logging.info('Extracting functions using oracle %s.', target_oracle)
+    tmp_functions = query_introspector_for_targets(project, target_oracle)
+
+    # Limit the amount of functions from each oracle.
+    functions += tmp_functions[:limit]
+
   if not functions:
-    logging.error('No unreached functions found')
+    logging.error('No functions found using the oracles: {target_oracles}')
```
DonggeLiu, line 352: Could you please use lazy format to avoid linter warnings? E.g.,
```python
logging.error('No functions found using the oracles: %s', target_oracles)
```

oliverchang:
For testing this, we now have support for running experiments via gcbrun (https://github.com/google/oss-fuzz-gen/pull/244). You should be able to run these with your current privileges. 

However, there's no way to propagate run_all_experiments.py flags via this yet (#245). One way around this is to temporarily modify the arguments here: https://github.com/google/oss-fuzz-gen/blob/3dd239a104eadef1868f94c709a3dbf4375fc1db/report/docker_run.sh#L113 in either this PR or a separate one branched from this.

run_all_experiments.py
```diff
@@ -42,6 +42,7 @@
 
 BENCHMARK_DIR: str = './benchmark-sets/comparison'
 RESULTS_DIR: str = run_one_experiment.RESULTS_DIR
+GENERATED_BENCHMARK_DIR: str = 'generated-benchmark-'
```
DonggeLiu, line 45: Could you keep this under the current `benchmkar-sets/` dir, so that all benchmarks are under it? E.g.,
```python
os.path.join('benchmark-sets', 'generated-')
```

run_all_experiments.py
```diff
@@ -63,6 +77,28 @@ def get_experiment_configs(
     print(f'A benchmark yaml file ({args.benchmark_yaml}) is provided. '
           f'Will use it and ignore the files in {args.benchmarks_directory}.')
     benchmark_yamls = [args.benchmark_yaml]
+  elif args.generate_benchmarks:
+    logging.info('Generating benchmarks.')
+    outdir = get_next_generated_benchmarks_dir()
+    logging.info('Setting benchmark directory to %s.', outdir)
+    os.makedirs(outdir, exist_ok=True)
```
DonggeLiu, line 84: Nit: I thought `get_next_generated_benchmarks_dir()` intends to ensure the `outdir` will not exist at this point?

run_all_experiments.py
DonggeLiu: +1
If they are known functions from projects not yet integrated into `OSS-Fuzz,` would it be a good idea to provide them as a default value? If there are too many, we can read them from a file.

run_all_experiments.py
```diff
@@ -182,10 +231,12 @@ def parse_args() -> argparse.Namespace:
             benchmark_yaml.endswith('yml')), (
                 "--benchmark-yaml needs to take an YAML file.")
 
-  assert bool(benchmark_yaml) != bool(args.benchmarks_directory), (
-      'One and only one of --benchmark-yaml and --benchmarks-directory is '
-      'required. The former takes one benchmark YAML file, the latter '
-      'takes: a directory of them.')
+  assert bool(benchmark_yaml) != bool(
+      args.benchmarks_directory) or args.generate_benchmarks, (
```
DonggeLiu, line 235: nit: The logic is not accurate, because it allows taking `--benchmark-yaml`, `--benchmarks-directory`, and `--generate-benchmarks` at the same time.
We need exactly one of them.

DonggeLiu (commented):
Thanks for adding this so quickly, @DavidKorczynski !

run_all_experiments.py
```diff
@@ -63,6 +77,28 @@ def get_experiment_configs(
     print(f'A benchmark yaml file ({args.benchmark_yaml}) is provided. '
           f'Will use it and ignore the files in {args.benchmarks_directory}.')
     benchmark_yamls = [args.benchmark_yaml]
+  elif args.generate_benchmarks:
+    logging.info('Generating benchmarks.')
```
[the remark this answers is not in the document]
DavidKorczynski: Yes, done.

run_all_experiments.py
DavidKorczynski: Done

run_all_experiments.py
DavidKorczynski: good call, adjusted to `benchmark_oracles`

data_prep/introspector.py
DavidKorczynski: Yes, done

run_all_experiments.py
DavidKorczynski: Yes, done

run_all_experiments.py
DavidKorczynski: Correct, I think my thought pattern was "doesn't hurt to allow exist_ok` but removed it now

run_all_experiments.py
DavidKorczynski: Fixed this to allow only one

run_all_experiments.py
DavidKorczynski: Done, PTAL

oliverchang (approved):

run_all_experiments.py
```diff
@@ -57,14 +58,35 @@ def __init__(self, benchmark, result):
...
+                    int(benchmark_folder.replace(GENERATED_BENCHMARK, '')))
     except (ValueError, TypeError) as _:
       pass
   max_idx += 1
-  return f'{GENERATED_BENCHMARK_DIR}{max_idx}'
+  return os.path.join(BENCHMARK_ROOT, f'{GENERATED_BENCHMARK}{max_idx}')
+
+
+def generate_benchmarks(args: argparse.Namespace) -> None:
+  """Generate benchmarks, write to filesystem and set args benchmark dir."""
```
DonggeLiu, line 72: nit: `Generates` for consistency.

run_all_experiments.py
```diff
@@ -198,16 +201,21 @@ def parse_args() -> argparse.Namespace:
...
-  parser.add_argument('-m',
-                      '--max-benchmark-targets',
+  parser.add_argument(
+      '-g',
+      '--generate-benchmarks',
+      help=('Generate benchmarks and use those for analysis. This is a string '
+            'of comma-separated heuristics to use when identifying benchmark '
+            'targets. Options available: '
+            f'{", ".join(introspector.get_oracle_dict().keys())}'),
+      type=str)
```
DonggeLiu, line 211: nit: [`choices`](https://docs.python.org/3/library/argparse.html#choices) is a simple way to ensure a set of allowed values for the arg.

DonggeLiu (approved):
Thanks @DavidKorczynski !

run_all_experiments.py
DavidKorczynski: There are some other options missing this, will address in a follow-up

run_all_experiments.py
DavidKorczynski: Sounds good, done

PR #250, opened 2024-05-06
experimental: add extended cmake options build heuristic

DavidKorczynski:
Sometimes projects will have default options on that can cause build issues, and being able to switch these off enable builds to succeed. This cmake option adds the ability to run a CMake run listing all options available, and enables turning off those default values that are ON.

### DIFF
--- a/experimental/c-cpp/build_generator.py
+++ b/experimental/c-cpp/build_generator.py
@@ -15,8 +15,6 @@
 """Utilities for generating builder scripts for a GitHub repository."""
 
 import os
-import sys
-import json
 import shutil
 import subprocess
 from abc import abstractmethod
@@ -300,6 +298,7 @@ class AutogenConfScanner(AutoBuildBase):
   def name(self):
     return 'autogen-ConfMake'
 
+
 class CMakeScannerOptsParser(AutoBuildBase):
   """Calls cmake to extract options from the CMakeLists.txt file of a project
   and creates a build string where all BOOL values are set to OFF except those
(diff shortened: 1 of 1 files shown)


experimental/c-cpp/build_generator.py
```diff
@@ -298,6 +299,120 @@ def name(self):
...
+    build_container = AutoBuildContainer()
+    build_container.list_of_commands = cmds_to_exec_from_root
+    build_container.heuristic_id = self.name + '1'
+    yield build_container
+
+  @property
+  def name(self):
+    return 'autogen-ConfMakeOpt'
+
+  def match_files(self, file_list: List[str]) -> None:
```
oliverchang, line 330: Can we add a docstring here?

experimental/c-cpp/build_generator.py
```diff
@@ -52,10 +53,10 @@ def match_files(self, file_list):
 
   def is_matched(self):
     """Returns True if the build heuristic found matching files."""
-    for matches in self.matches_found:
-      if len(matches) == 0:
-        return False
-    return True
+    for found_matches in self.matches_found.values():
+      if len(found_matches) > 1:
```
oliverchang, line 57: Why > 1 and not > 0? Can you add a comment here?

oliverchang (approved):

experimental/c-cpp/build_generator.py
DavidKorczynski: It should be > 0 -- thanks for catching!

experimental/c-cpp/build_generator.py
DavidKorczynski: done

DavidKorczynski:
No need for trial build in experimental

PR #260, opened 2024-05-09
experimental: extend and improve various places

DavidKorczynski:
This commit improves the logic by:
- Identifying language of target project, to make language specific
  prompts
- Adds context to the prompts based on cross-references, source code and
  extended typing
- Adds two new prompts for C
- Adds an addition harness run for each auto-gen harness with
  `detect_leaks=0`. This is to get idea of how well a harness performs
  in the event it runs into a leak fast.
- Saves the introspector reports used per analysis. This is important
  for development purposes to make it easy to track what the prompts
  used.
- Saves the prompts generated per auto-gen harness.

### DIFF
--- a/experimental/c-cpp/manager.py
+++ b/experimental/c-cpp/manager.py
@@ -344,7 +344,8 @@ class FuzzHeuristicGeneratorBase:
     return uniq_targets
 
 
-def get_fuzzer_source_code(func) -> str:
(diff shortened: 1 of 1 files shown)


DavidKorczynski:
This code is not tested in the CI so will skip trial runs

oliverchang (approved):

PR #259, opened 2024-05-09
Pass code block to fixer prompt, not the raw LLM response

DonggeLiu:
Passing the raw response can confuse LLM when it contains more text than the code block, which occurred on `Gemini 1.5`.
Also, `Gemini 1.5`'s response structure seems to be different from `code-bison-32k`.

This PR does two things:
1. Make the response parser compatible with both models.
2. Pass the code block in response to LLM fixer, not the raw response.

PR #261, opened 2024-05-10
Pass code block to fixer prompt, not the raw LLM response (#259)

DonggeLiu:
Passing the raw response can confuse LLM when it contains more text than the code block, which occurred on `Gemini 1.5`.
Also, `Gemini 1.5`'s response structure seems to be different from `code-bison-32k`.

This PR does three things:
1. Make the response parser compatible with both models.
2. Pass the code block in response to LLM fixer, not the raw response.
3. More instructions in the code-fixing prompt to avoid common mistakes.


DonggeLiu:
Merging this as it is cherry-picked from #259.

PR #272, opened 2024-05-14
Modify context generation slightly (xrefs)

trashvisor:
Truncate xrefs, so that the source code of the entire function(s) xref'ing the function under test is not included.

Additionally, modify context generation so we talk about lines where the code is xref'd, as opposed to functions.

### DIFF
--- a/data_prep/project_context/context_introspector.py
+++ b/data_prep/project_context/context_introspector.py
@@ -162,7 +162,31 @@ class ContextRetriever:
       logging.warning(
           'Could not retrieve xrefs for project: %s '
           'function_signature: %s', project, func_sig)
+
+    # At times, xrefs can be noisy (multiple 100s of loc).
+    # Truncate them by default
+    xrefs = self._truncate_xrefs(xrefs)
     return xrefs
+
+  def _truncate_xrefs(self, xrefs: list[str]) -> list[str]:
+    truncated = []
+    for xref in xrefs:
+      lines = xref.split('\n')
+      line_index = -1
+      start_index = 0
+      end_index = len(lines) - 1
+      for index, line in enumerate(lines):
+        if self._benchmark.function_name in line:
+          line_index = index
+          break
+      # If name was not found, then just return the entire function.
+      # If it was, truncate it.
+      if line_index != -1:
+        start_index = 0 if line_index <= 10 else line_index - 10
+        end_index = end_index if line_index >= end_index - 10 else line_index + 10
+      truncated.append('\n'.join(lines[start_index:end_index]))
+
+    return truncated
 
   def get_context_info(self) -> dict:
     """Retrieves contextual information and stores them in a dictionary."""
--- a/data_prep/project_context/context_introspector.py
+++ b/data_prep/project_context/context_introspector.py
@@ -169,7 +169,7 @@ class ContextRetriever:
     return xrefs
 
   def _truncate_xrefs(self, xrefs: list[str]) -> list[str]:
-    """Truncates xrefs to 10 lines before and after the 
+    """Truncates xrefs to 10 lines before and after the
     function name is referenced."""
     truncated = []
     for xref in xrefs:
--- a/data_prep/project_context/context_introspector.py
+++ b/data_prep/project_context/context_introspector.py
@@ -169,12 +169,14 @@ class ContextRetriever:
     return xrefs
 
   def _truncate_xrefs(self, xrefs: list[str]) -> list[str]:
+    """Truncates xrefs to 10 lines before and after the 
+    function name is referenced."""
     truncated = []
     for xref in xrefs:
       lines = xref.split('\n')
       line_index = -1
-      start_index = 0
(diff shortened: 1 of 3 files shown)


data_prep/project_context/context_introspector.py
```diff
@@ -162,8 +162,34 @@ def _get_xrefs_to_function(self) -> list[str]:
...
+      end = len(lines) - 1
+      for index, line in enumerate(lines):
+        if self._benchmark.function_name in line:
+          line_index = index
+          break
+      # If name was not found, then just return the entire function.
+      # If it was, truncate it.
+      if line_index != -1:
+        start = start if line_index <= 10 else line_index - 10
+        end = end if line_index >= end - 10 else line_index + 10
```
DonggeLiu, line 188: nit:
```python
start = max(0, line_index - 10) 
end = min(len(lines)-1, line_index+10)
```

DonggeLiu:
@trashvisor Could you please check if the report:
1. Has any unexpected behavior (e.g., failed to include certain xrefs)
2. Significant improvements / regressions?
https://llm-exp.oss-fuzz.com/Result-reports/ofg-pr/2024-05-15-272-dg-272-comparison/index.html
Thanks!

if not, I reckon this is ready to merge.

DonggeLiu:
A silly question:
What is the intention of [this block](https://github.com/google/oss-fuzz-gen/blame/39bebbaad6087a1f1486faa8045c04a6ca268b7b/prompts/template_xml/context.txt#L12-L17) in the prompt?

I noticed that it adds [the function declaration](https://llm-exp.oss-fuzz.com/Result-reports/scheduled/2024-05-11-weekly-all/benchmark/output-abseil-cpp-_zn4absl19str_format_internal13formatargimpl8dispatchiaeebns1_4dataens0_24formatconversionspecimplepv/index.html#:~:text=You%20must%20insert%20code%20in%20the%20below%20code%20block%20before%20the%20function%20being%20tested%3A%0A%3Ccode%3E%0Abool%20absl%3A%3Astr_format_internal%3A%3AFormatArgImpl%3A%3ADispatch%3Cabsl%3A%3Astr_format_internal%3A%3AVoidPtr%3E(DW_TAG_union_typeData%2C%20FormatConversionSpecImpl%2C%20void%20*)%3B%0A%3C/code%3E), which is the same as the function signature added earlier.

Related:
https://github.com/google/oss-fuzz-gen/blob/39bebbaad6087a1f1486faa8045c04a6ca268b7b/llm_toolkit/prompt_builder.py#L165

trashvisor:
> A silly question: What is the intention of [this block](https://github.com/google/oss-fuzz-gen/blame/39bebbaad6087a1f1486faa8045c04a6ca268b7b/prompts/template_xml/context.txt#L12-L17) in the prompt? Is it to mitigate the 'undefined function' error?
> 
> I noticed that it adds [the function declaration](https://llm-exp.oss-fuzz.com/Result-reports/scheduled/2024-05-11-weekly-all/benchmark/output-abseil-cpp-_zn4absl19str_format_internal13formatargimpl8dispatchiaeebns1_4dataens0_24formatconversionspecimplepv/index.html#:~:text=You%20must%20insert%20code%20in%20the%20below%20code%20block%20before%20the%20function%20being%20tested%3A%0A%3Ccode%3E%0Abool%20absl%3A%3Astr_format_internal%3A%3AFormatArgImpl%3A%3ADispatch%3Cabsl%3A%3Astr_format_internal%3A%3AVoidPtr%3E(DW_TAG_union_typeData%2C%20FormatConversionSpecImpl%2C%20void%20*)%3B%0A%3C/code%3E), which is the same as the function signature added earlier.
> 
> Related:
> 
> https://github.com/google/oss-fuzz-gen/blob/39bebbaad6087a1f1486faa8045c04a6ca268b7b/llm_toolkit/prompt_builder.py#L165

It was originally intended to solve:
* Not having the appropriate header file for the function declaration (so we would just embed the function decl itself without a header and things might work).
*  Compatibility with c programs, as an `extern "C"` would be added to solve linkage issues. 

Should this block be removed or some templating logic added?

DonggeLiu:
> It was originally intended to solve:
> 
> * Not having the appropriate header file for the function declaration (so we would just embed the function decl itself without a header and things might work).

Ah, thanks for the explanation!

I doubt this will help LLM get the correct header file, though:
1. We already have the signature in the prompt, so this doesn't add new information.
2. LLM is not good at searching for the corresponding header file given the function signature, it can only guess.

But don't worry; AdaLogic will look into finding the correct header file in a more programmatic way.

> * Compatibility with c programs, as an `extern "C"` would be added to solve linkage issues.

In this case, maybe:
1. Make the instruction more explicit, e.g.,
```
Include the following extern C linkage specification in the fuzz target, because the function-under-test is defined in C and the fuzz target is written in C++:
extern ''C" <function-signature>
```
2. Only add this code if the project-under-test is in C.


Would this sound reasonable? Thanks : )

PR #277, opened 2024-05-16
Match another oom pattern in crash log

DonggeLiu:
The previous pattern does not catch this:
https://llm-exp.oss-fuzz.com/Result-reports/ofg-pr/2024-05-16-276-dg-276-1-comparison/sample/output-libusb-libusb_get_string_descriptor_ascii/03


DonggeLiu:
TBR.

PR #281, opened 2024-05-18
More full exp fixes

DonggeLiu:
1. Update the report one more time right before exiting, just to ensure it is up-to-date at the end.
2. Add package `chartdet` to dynamically update the encoding for `textbox`, updated `requirement.*` accordingly.
3. Downgrade a distracting warning when generating training data.


experiment/builder_runner.py
```diff
@@ -110,7 +110,11 @@ def _libfuzzer_args(self) -> list[str]:
         f'-max_total_time={self.run_timeout}',
         # Without this flag, libFuzzer only consider short inputs in short
         # experiments, which lowers the coverage for quick performance tests.
-        '-len_control=0'
+        '-len_control=0',
+        # Timeout per testcase.
+        '-timeout=30',
+        # Truncate testcase.
+        '-max_len=4096',
```
oliverchang, line 117: This is a bit dangerous and can cause real bugs to be missed. Was there any reason in particular for this one?

requirements.txt
```diff
@@ -2,7 +2,7 @@
 # This file is autogenerated by pip-compile with Python 3.11
```
oliverchang, line 2: can you add the description for this change into the PR as well?

experiment/textcov.py
```diff
@@ -146,6 +149,30 @@ class Textcov:
...
+    """Reads file_handle assuming its encoding is utf-8, detects the encoding
+    if otherwise."""
+    file_content = file_handle.read()
+
+    try:
+      # Try decoding the file content with UTF-8 encoding
+      return file_content.decode('utf-8')
+    except UnicodeDecodeError:
+      # If UTF-8 decoding fails, detect the file's encoding
+      raw_data = file_content[:sample_size]
```
oliverchang, line 165: Can you add the description for this change into the PR?

oliverchang (commented):
nice! just some questions.

experiment/builder_runner.py
DonggeLiu: Ah, I was going to consult you about whether this is a good idea but forgot to comment on it.
I was worried about the same thing (missing bugs) but was wondering if this can help reduce OOM (e.g., many memory allocations depend on `data` or use `size`).

Let me remove it, then.

experiment/textcov.py
DonggeLiu: Added this and the one above to 2.

oliverchang (approved):

PR #293, opened 2024-05-23
Loosen all safety restrictions of `VertexAI` so that response won't be blocked

DonggeLiu:
1. Loosen all safety restrictions of `VertexAI` so that response won't be blocked
2. When an LLM API error happens, use the default value (`''` for Gemini). This helps prevent non-terminating experiments.
3. Rename class name `Gemini1_5` to `Gemini1D5` to avoid linter complaints.


DonggeLiu:
Only 7 benchmark trials failed to terminate in https://llm-exp.oss-fuzz.com/Result-reports/scheduled/2024-05-22-manual-all/, all likely due to this.

oliverchang (commented):
would https://cloud.google.com/vertex-ai/generative-ai/docs/multimodal/configure-safety-attributes also help to reduce this? 

Not sure if we are allowed to use `BLOCK_NONE`, but at least `BLOCK_ONLY_HIGH` should eliminate a lot of these errors?

DonggeLiu:
The `blocked` benchmarks work now.

DonggeLiu:
https://llm-exp.oss-fuzz.com/Result-reports/ofg-pr/2024-05-27-293-dg-blocked/index.html

oliverchang (approved):

PR #294, opened 2024-05-23
experimental: extend builds and minor refactoring

DavidKorczynski:
- Adjusts the autogen.sh build to be more general (no need to have
  Makefile as this can be generated by the autogen.sh in some cases).
- Uses a unified place for templating
- Minor adjustment in directory name of shared folder
  "autogen-results-X" is now always just "autogen-results"

### DIFF
--- a/experimental/c-cpp/runner.py
+++ b/experimental/c-cpp/runner.py
@@ -26,8 +26,8 @@ import templates
 
 silent_global = False
 
+SHARED_MEMORY_RESULTS_DIR = 'autogen-results'
 
-SHARED_MEMORY_RESULTS_DIR='autogen-results'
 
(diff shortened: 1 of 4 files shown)


experimental/c-cpp/runner.py
```diff
@@ -277,7 +225,7 @@ def run_on_targets(target,
 
   openai_api_key = os.getenv('OPENAI_API_KEY', None)
 
-  outdir = '/out/autogen-results-%d' % (idx)
+  outdir = '/out/' + SHARED_MEMORY_RESULTS_DIR
```
DonggeLiu, line 228: nit: `os.path.join()`

DonggeLiu (approved):

PR #295, opened 2024-05-23
Analyze linker error messages and add in source context for undefined function reference

cjx10:
Before merging:

- [ ] Remove selected benchmark set


llm_toolkit/code_fixer.py
```diff
@@ -381,71 +397,7 @@ def extract_error_message(log_path: str,
       _strip_color_code(line).rstrip()
       for line in log_lines[error_lines_range[0]:error_lines_range[1]]
   ]
-  return group_error_messages(errors)
-
-
-def group_error_messages(error_lines: list[str]) -> list[str]:
```
DonggeLiu, line 387: Were there any code changes when moving this to the new file?

DonggeLiu (approved):
Thanks!

llm_toolkit/code_fixer.py
cjx10: No except for [this part](https://github.com/google/oss-fuzz-gen/pull/295/files#diff-651bac2b0566f48748ab30d11c42c8e826aa1ac777aafb6f96d6bab5626240dbR89-R93): instead of skipping lines that has `DWARF error`, trim the error string instead, new line in gnu ld error message is not very consistent

DonggeLiu: Wow, thanks sooo much for continuing to reply to this, @cjx10!

DonggeLiu:
Outdated

PR #298, opened 2024-05-27
oss-fuzz-checkout: enable using own venv

DavidKorczynski:
This is useful when having a shared virtual environment with other projects. For example, on my local dev I use a virtualenv environment with both fuzz introspector and oss-fuzz-gen requirements, but this is currently being blocked by the logic here trying to launch a new venv.

### DIFF
--- a/experiment/oss_fuzz_checkout.py
+++ b/experiment/oss_fuzz_checkout.py
@@ -104,6 +104,10 @@ def postprocess_oss_fuzz() -> None:
   if os.path.exists(os.path.join(OSS_FUZZ_DIR, VENV_DIR)):
     return
 
+  # If already in a virtualenv environment assume all is set up
+  if os.environ['VIRTUAL_ENV']:
+    return
+
   result = sp.run(['python3', '-m', 'venv', VENV_DIR],
                   check=True,
                   capture_output=True,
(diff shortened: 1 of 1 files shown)


DonggeLiu (approved):

DonggeLiu:
The experiment failed due to a key error when not in a venv.
BTW, I often use `-n dg` to name experiments because that's my name (DongGe), please feel free to use other names for yours : )

experiment/oss_fuzz_checkout.py
```diff
@@ -104,6 +104,10 @@ def postprocess_oss_fuzz() -> None:
   if os.path.exists(os.path.join(OSS_FUZZ_DIR, VENV_DIR)):
     return
 
+  # If already in a virtualenv environment assume all is set up
+  if os.environ['VIRTUAL_ENV']:
```
DonggeLiu, line 108: Maybe `os.getenv('VIRTUAL_ENV')`?

DonggeLiu (changes requested):

experiment/oss_fuzz_checkout.py
oliverchang: +1. this should exception otherwise right?

DavidKorczynski: Thanks! Fixed

DonggeLiu (approved):
Thanks!

PR #299, opened 2024-05-28
run_all_experiments: set introspector endpoint as early as possible

DavidKorczynski:
Otherwise we will use the default endpoint when we also generate benchmarks during the run (happens inside of `get_experiment_configs`). This commit ensures we use the right endpoint in both local and cloud-based runs.


DavidKorczynski:
For some reason the result HTML page is blank, and I'm not sure why.

DonggeLiu (approved):
The status in the report appears to have a minor regression, but I don't think that's related to this change.
It's likely a stochastic issues.

PR #308, opened 2024-06-07
Fix file.yaml

DonggeLiu:
Project `file` does not use the fuzz target in the project repo, but the one in OSS-Fuzz directory.
Proof:
1. The fuzz target used is called [`magic_fuzzer.cc`](https://llm-exp.oss-fuzz.com/Result-reports/scheduled/2024-05-22-manual-all/sample/output-file-apprentice_buf/01), the same as the one in [OSS-Fuzz dir](https://github.com/google/oss-fuzz/blob/74e3c9413ac02ea72bc3be161096a390f05de846/projects/file/magic_fuzzer.cc). The one from its repo is called [`magic_fuzzer.c`](https://github.com/file/file/blob/99d21e9f799e415ea8f616b9de093b4e9a1bae56/fuzz/magic_fuzzer.c).
2. [This fuzz target](https://llm-exp.oss-fuzz.com/Result-reports/scheduled/2024-05-22-manual-all/sample/output-file-apprentice_buf/01#:~:text=Crash%20reason%3A%20NO_SEMANTIC_ERR-,Final%20code,-%23include%20%3Cfuzzer/FuzzedDataProvider) does not have 52 lines, the [one in OSS-Fuzz dir](https://github.com/google/oss-fuzz/blob/74e3c9413ac02ea72bc3be161096a390f05de846/projects/file/magic_fuzzer.cc) does.


oliverchang (commented):
lgtm

PR #310, opened 2024-06-07
run_one_experiment: ensure no parallel executions in local runs

DavidKorczynski:
This causes issues with docker container reuse.

It's mentioned in the code: https://github.com/google/oss-fuzz-gen/blob/18e1af1569be850e588ee5b0319303bca583b622/run_one_experiment.py#L34-L39 

But I think a better workflow would be to adjust this automatically if there is no cloud run anyways.

### DIFF
--- a/run_one_experiment.py
+++ b/run_one_experiment.py
@@ -223,6 +223,7 @@ def run(benchmark: Benchmark,
         run_timeout: int = RUN_TIMEOUT,
         dry_run: bool = False) -> Optional[AggregatedResult]:
   """Generates code via LLM, and evaluates them."""
+  global NUM_EVA
   model.cloud_setup()
   logging.basicConfig(level=logging.INFO)
 
--- a/run_one_experiment.py
+++ b/run_one_experiment.py
@@ -226,6 +226,11 @@ def run(benchmark: Benchmark,
   model.cloud_setup()
   logging.basicConfig(level=logging.INFO)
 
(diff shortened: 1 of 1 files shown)


DonggeLiu:
Thanks @DavidKorczynski 

> This causes issues with docker container reuse.

Do you mean two trials may accidentally use the same docker containers (e.g., due to duplicate docker image/container names)?

> # WARN: Avoid NUM_EVA for local experiments. 

There is a typo in my comment here: I meant to say, `Avoid high value for NUM_EVA for local experiments.` This is because users may unintentionally use a lot more computation power.

DavidKorczynski:
> Thanks @DavidKorczynski
> 
> > This causes issues with docker container reuse.
> 
> Do you mean two trials may accidentally use the same docker containers (e.g., due to duplicate docker image/container names)?
> 
> > `# WARN: Avoid NUM_EVA for local experiments. `
> 
> There is a typo in my comment here: I meant to say, `Avoid high value for NUM_EVA for local experiments.` This is because users may unintentionally use a lot more computation power.

The restarting of containers was in the end due to the issue fixed here: https://github.com/google/oss-fuzz-gen/pull/311 

Left this PR open because of the typo, as in either case I think we should do the following to make it easy to control this var: https://github.com/google/oss-fuzz-gen/pull/316

Let's close this and focus on the other two -- can submit a patch for the typo in case.

PR #311, opened 2024-06-07
project_src: stop containers that were started to collected src

DavidKorczynski:
Without doing this the containers are left hanging, which causes issues for local runs since the specific container is used multiple times in a single experiment.

### DIFF
--- a/data_prep/project_src.py
+++ b/data_prep/project_src.py
@@ -271,7 +271,15 @@ def _copy_project_src_from_local(project: str, out: str):
     logging.error('Failed to copy /src from OSS-Fuzz image of %s:', project)
     logging.error('STDOUT: %s', result.stdout)
     logging.error('STDERR: %s', result.stderr)
+
+    # Shut down the container that was just started before throwing exception.
+    sp.run(['docker', 'container', 'stop', f'{project}-container'])
+
     raise Exception(f'Failed to run docker command: {" ".join(copy_src)}')
+
+  # Shut down the container that was just started.
+  sp.run(['docker', 'container', 'stop', f'{project}-container'])
+
   logging.info('Done copying %s /src to %s.', project, out)
 
 
--- a/data_prep/project_src.py
+++ b/data_prep/project_src.py
@@ -279,7 +279,12 @@ def _copy_project_src_from_local(project: str, out: str):
     logging.info('Done copying %s /src to %s.', project, out)
   finally:
     # Shut down the container that was just started.
-    sp.run(['docker', 'container', 'stop', f'{project}-container'])
+    result = sp.run(['docker', 'container', 'stop', f'{project}-container'],
+                    capture_output=True, stdin=sp.DEVNULL, check=False)
+    if result.returncode:
+      logging.error('Failed to stop container image: %s-container', project)
+      logging.error('STDOUT: %s', result.stdout)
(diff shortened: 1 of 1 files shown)


data_prep/project_src.py
```diff
@@ -271,7 +271,15 @@ def _copy_project_src_from_local(project: str, out: str):
     logging.error('Failed to copy /src from OSS-Fuzz image of %s:', project)
     logging.error('STDOUT: %s', result.stdout)
     logging.error('STDERR: %s', result.stderr)
+
+    # Shut down the container that was just started before throwing exception.
+    sp.run(['docker', 'container', 'stop', f'{project}-container'])
```
oliverchang, line 276: Can we do a `try ... finally` here and put the stop in in the `finally` ?

DonggeLiu: +1

DavidKorczynski: As there's no `try-except` in the first place I just wanted to double check -- do you suggest wrapping the initial `result = sp.run(copy_src, capture_output=True, stdin=sp.DEVNULL, check=False)` in a try?

DonggeLiu: Maybe something like this so that we only need one `sp.run(['docker', 'container', 'stop', f'{project}-container'], check=True)`?
```python
try:
    result = sp.run(copy_src, capture_output=True, stdin=sp.DEVNULL, check=False)
    if result.returncode:
        logging.error('Failed to copy /src from OSS-Fuzz image of %s:', project)
        logging.error('STDOUT: %s', result.stdout.decode())
        logging.error('STDERR: %s', result.stderr.decode())
        raise Exception('Copy operation failed')
finally:
    sp.run(['docker', 'container', 'stop', f'{project}-container'], check=True)
```

DavidKorczynski: Done, thanks!

DonggeLiu: Thanks for fixing this, the report also looks great.

I added [377aab2](https://github.com/google/oss-fuzz-gen/pull/311/commits/377aab2d86e71133de8ba5bd18ae61f5c5d33b6c) to:
1. Avoid weird failure when `stdin=sp.DEVNULL` is not set; we noticed some of this in the past.
2. Log errors if the stop command fails.
3. Set `check=False` and `capture_output=True` to make logging consistent with other `sp` calls above.

DavidKorczynski: > Thanks for fixing this, the report also looks great.
> 
> I added [377aab2](https://github.com/google/oss-fuzz-gen/pull/311/commits/377aab2d86e71133de8ba5bd18ae61f5c5d33b6c) to:
> 
> 1. Avoid weird failure when `stdin=sp.DEVNULL` is not set; we noticed some of this in the past.
> 2. Log errors if the stop command fails.
> 3. Set `check=False` and `capture_output=True` to make logging consistent with other `sp` calls above.
> 
> Hope you don't mind : )

For bug fixes I would faster landing and follow up PRs. This makes it easier to use the fix for continued dev.

DonggeLiu (approved):

DonggeLiu:
The report looks good.
https://llm-exp.oss-fuzz.com/Result-reports/ofg-pr/2024-06-15-311-dk1000-4-comparison/index.html

PR #312, opened 2024-06-07
[JVM] Add handling for generic argument types

arthurscchan:
For Java projects, some of the methods may require arguments with generic types. This PR proposes a fix to include generic type suggestions in the prompts to allow the LLM generate a better result related to methods with generic type arguments.


DavidKorczynski (approved):
Experiment in https://github.com/google/oss-fuzz-gen/pull/320

PR #316, opened 2024-06-08
experiment: make LLM_FIX_LIMIT adjustable

DavidKorczynski:
During development it's convenient to being able to adjust this, for example when focus during development is on aspects of OSS-Fuzz-gen that is unrelated to code fixing it's convenient to decrease this to speed up evaluation.


DonggeLiu (approved):
Thanks!

DonggeLiu:
> During development it's convenient to being able to adjust this, 

This is a good point.
In addition to development, we can dynamically adjust the number of fixes by checking if the fix is progressing.
If two fix results are identical, we can stop early, too.
If the last result appears promising, we can extend it for a few more.

PR #323, opened 2024-06-10
Expand the function candidates for projects

arthurscchan:
For some of the projects, the query oracle from fuzz-introspector API may return a very limited set of function candidates for generating fuzzers. After certain type of filtering, it could result in an empty list of candidates. This PR proposes an improvement by extending the search from the full function list if there are no more candidates to work on after the filtering is applied on oss-fuzz-gen.


DavidKorczynski (commented):
I think we should aim to expose this by way of an API rather, or, perhaps add arguments to `all-public-functions` that enable you to do filtering/sorting as a way of returning a list with only a subset of the total elements.

arthurscchan:
No longer needed.

PR #324, opened 2024-06-11
[JVM] Fix bug which JVM project does not generate err.log

arthurscchan:
JVM project does no generate err.log in the docker workspace from OSS-Fuzz, this PR proposes a fix to add a language check when retrieving err.log and skip it if the target project is a JVM project to avoid FileNotFoundError.


DavidKorczynski:
I'll test this once we get introspector.oss-fuzz.com's API up and running again

DavidKorczynski (approved):

PR #325, opened 2024-06-11
Fix bug in jvm prebuild check

arthurscchan:
The prebuild check for JVM projects has missed the consideration of a method call across multiple lines, this cause some prebuild check failed to locate the target method call when it is split into multiple lines which result in false negative denial of possible valid fuzzing harnesses. This PR proposes a fix in the prebuild check to ensure multiple lines target method call could be recognised and fix the problem of false negative denial of possible valid fuzzing harnesses.

### DIFF
--- a/experiment/builder_runner.py
+++ b/experiment/builder_runner.py
@@ -134,13 +134,13 @@ class BuilderRunner:
     # overloaded methods. Thus it need to use the regex to identify
     # if there are method calls with unknown variable names that match
     # the target method.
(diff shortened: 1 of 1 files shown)


DavidKorczynski:
I'll test this once we get introspector.oss-fuzz.com's API up and running again

DavidKorczynski:
experiment in https://github.com/google/oss-fuzz-gen/pull/340

DavidKorczynski (approved):

PR #317, opened 2024-06-08
data_prep: only include c/cpp files as potential harnesses

DavidKorczynski:
A more precise approach is needed when searching for potential harnesses due to performance issues.

For example, when analyzing `croaring` the following files all match the existing set up, but none of them have the potential of being the actual source code of a harness.

```sh
/tmp/tmpjg7oqx4m/out/src/croaring/build-dir/CMakeFiles/CMakeScratch/TryCompile-9ReWwb/cmake_install.cmake
/tmp/tmpjg7oqx4m/out/src/croaring/build-dir/CMakeFiles/CMakeScratch/TryCompile-9ReWwb/CMakeFiles/cmake.check_cache
/tmp/tmpjg7oqx4m/out/src/croaring/build-dir/CMakeFiles/CMakeScratch/TryCompile-9ReWwb/CMakeFiles/CMakeDirectoryInformation.cmake
/tmp/tmpjg7oqx4m/out/src/croaring/build-dir/CMakeFiles/CMakeScratch/TryCompile-9ReWwb/CMakeFiles/cmTC_bc5f1.dir/DependInfo.cmake
/tmp/tmpjg7oqx4m/out/src/croaring/build-dir/CMakeFiles/CMakeScratch/TryCompile-9ReWwb/CMakeFiles/cmTC_bc5f1.dir/CMakeCXXCompilerABI.cpp.o.d
/tmp/tmpjg7oqx4m/out/src/croaring/build-dir/CMakeFiles/CMakeScratch/TryCompile-9ReWwb/CMakeFiles/cmTC_bc5f1.dir/cmake_clean.cmake
/tmp/tmpjg7oqx4m/out/src/croaring/build-dir/CMakeFiles/CMakeScratch/TryCompile-9ReWwb/CMakeFiles/cmTC_bc5f1.dir/CMakeCXXCompilerABI.cpp.o
```

For `croaring`, a total of 1200+ potential harnesses are identified with the current approach, and it takes a fairly long time (tens of minutes) to run through scanning for potential harnesess as a fair amount of the wrong files causes the 100sec timeout in `clang-format`.

This PR makes the logic more precise to only include relevant source code files. For `croaring` it reduces to 65 potential harnesses (including the relevant harnesses) and it takes a couple of seconds to go through scanning for potential harnesses.

PR #333, opened 2024-06-13
project_src: fix C harness identification

DavidKorczynski:
There is an issue when identifying harnesses for C projects. Currently, because of a missing language check all files in a C project will be considered valid C harnesses. Thus, when we identify the harnesss to subtitute our auto-gen code with we pick one from the "potential harness" list

https://github.com/google/oss-fuzz-gen/blob/0bed2b5a60d4af36bc0bc7565c57558bb08897da/data_prep/introspector.py#L422 

which effectively means that the majority of times (chance of `pick_one` choosing a valid harness is total harnesses divided by total C files (and many others prior to https://github.com/google/oss-fuzz-gen/pull/317)) a non-harness will be selected for substitution. Consequently, the auto-gen code will not be used.

This was causing a lot of weirdness with a C-specific prompt I'm building since we would substitute fuzzing code into randomly picked files, essentially causing an entire build to fail as we would replace arbitrary files with harness code.

### DIFF
--- a/data_prep/project_src.py
+++ b/data_prep/project_src.py
@@ -125,8 +125,7 @@ def _get_harness(src_file: str, out: str, language: str) -> tuple[str, str]:
 
   content = _format_source(src_file)
 
-  if (language == 'c++' or
-      language == 'c') and 'int LLVMFuzzerTestOneInput' not in content:
+  if language in {'c++', 'c'} and 'int LLVMFuzzerTestOneInput' not in content:
     return '', ''
   if language == 'jvm' and 'static void fuzzerTestOneInput' not in content:
     return '', ''
--- a/data_prep/project_src.py
+++ b/data_prep/project_src.py
@@ -125,9 +125,11 @@ def _get_harness(src_file: str, out: str, language: str) -> tuple[str, str]:
 
   content = _format_source(src_file)
 
-  if language in {'c++', 'c'} and 'int LLVMFuzzerTestOneInput' not in content:
+  if language.lower() in {'c++', 'c'
+                         } and 'int LLVMFuzzerTestOneInput' not in content:
     return '', ''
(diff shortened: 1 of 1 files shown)


DavidKorczynski:
I think this means that (1) previous builds for C projects would likely fail due to build errors non-related to harness generation and (2) it would likely not be the auto-gen C code that was running, but rather the existing harnesses as the substitution would just have substituted some file in the repo  (which could be a non-code file before https://github.com/google/oss-fuzz-gen/pull/317, which is likely when the build would succeed).

DavidKorczynski:
(with a WIP prompt):

Before: 
Screenshot 2024-06-14 010022

After this change:
Screenshot 2024-06-14 010409

DavidKorczynski:
FWIW this was introduced in the JVM integration so is relatively recent

DavidKorczynski:
Ref: https://github.com/google/oss-fuzz-gen/issues/337

DonggeLiu (approved):

data_prep/project_src.py
```diff
@@ -125,7 +125,7 @@ def _get_harness(src_file: str, out: str, language: str) -> tuple[str, str]:
 
   content = _format_source(src_file)
 
-  if language == 'c++' and 'int LLVMFuzzerTestOneInput' not in content:
+  if language in {'c++', 'c'} and 'int LLVMFuzzerTestOneInput' not in content:
```
DonggeLiu, line 128: `language. lower()` would be more robust.
Just in case the language string [is capitalized](https://github.com/google/oss-fuzz-gen/blob/e5a4fbff52496dd157dc967e1af11c7f08837499/experiment/oss_fuzz_checkout.py#L151).

DavidKorczynski: done, thanks

DonggeLiu (approved):

DavidKorczynski:
Report is looking good.

PR #345, opened 2024-06-18
A script to run Gemini models

DonggeLiu:
Adding a script to run Gemini models with a text prompt.
Steps:
```shell
python -m venv .venv
pip install -r requirements.txt
gcloud auth login && gcloud auth application-default login && gcloud auth application-default set-quota-project oss-fuzz
python -m experimental.manual.prompter -p prompt.txt -l vertex_ai_gemini-1-5
ls `./responses/`
```

See the [`name` attribute](https://github.com/google/oss-fuzz-gen/blob/main/llm_toolkit/models.py#L422) for more supported model names.


DonggeLiu:
TBR.

inferno-chromium:
Thanks so much. How do you also provide a project source code, along with the prompt, or do you just include it inside the prompt.

PR #348, opened 2024-06-18
introspector: avoid static functions for targets by default 

DavidKorczynski:
Introspector added a heuristic for avoiding static functions and we should use this by default when picking fuzz targets. This avoids hitting harnesses that will fail.

I made this default, but with an option as I assume it will more or less never be desired to have static functions as potential targets. However, there are cases where harnesses for static functions are interesting, although it would require either (1) removal of `static` by the harness construction logic or (2) adding the harness to the relevant file. For these reasons I keep it as an option, in case someone wants to give those types of harnesses a go.

Tested with:

```sh
./run_all_experiments.py \
  --model=${MODEL}
  -g far-reach-low-coverage,low-cov-with-fuzz-keyword,easy-params-far-reach \
  -gp croaring \
  -gm 15 \
  --prompt-builder CSpecific \
  -e http://127.0.0.1:8080/api
```

Can confirm the success build ratio of auto generated fuzz harnesses increases.

### DIFF
--- a/data_prep/introspector.py
+++ b/data_prep/introspector.py
@@ -38,6 +38,10 @@ T = TypeVar('T', str, list, dict)  # Generic type.
 TIMEOUT = 45
 MAX_RETRY = 5
 
+# By default exclude static functions when identifying fuzz target candidates
+# to generate benchmarks.
+ORACLE_AVOID_STATIC_FUNCTIONS=bool(int(os.getenv('OSS_FUZZ_AVOID_STATIC_FUNCTIONS', '1')))
+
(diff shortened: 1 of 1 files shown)


DonggeLiu (approved):
Do you plan to change benchmarks selection in the following 2 weeks?
Please let me know when it is settled, I will update the benchmark-sets and re-run a full experiment.

DavidKorczynski:
> Do you plan to change benchmarks selection in the following 2 weeks? Please let me know when it is settled, I will update the benchmark-sets and re-run a full experiment.

Sounds good! We can aim for running a full experiment the week starting in 10 days

PR #352, opened 2024-06-18
report: use argparse to parse arguments

DavidKorczynski:
Use argparse for the report generation and adjust USAGE.md accordingly. A benefit of this is that instructions will only be deprecated if we change argument names in the future.

### DIFF
--- a/report/web.py
+++ b/report/web.py
@@ -514,12 +514,25 @@ class GenerateReport:
 def _parse_arguments() -> argparse.Namespace:
   """Parses command line args."""
   parser = argparse.ArgumentParser(
-      description = 'Parse arguments to launch the web report generation.')
+      description='Parse arguments to launch the web report generation.')
 
-  parser.add_argument('--results-dir', '-r', help='Directory with results from OSS-Fuzz-gen.', required=True)
-  parser.add_argument('--output-dir', '-o', help='Directory to store statically generated web report.', default='results-report')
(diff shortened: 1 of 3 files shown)


report/web.py
```diff
@@ -510,18 +511,41 @@ def _write_benchmark_sample(self, benchmark_id: str, sample_id: str):
       print(f'Failed to write sample/{benchmark_id}/{sample_id}:\n{e}')
 
 
+def _parse_arguments() -> argparse.Namespace:
+  """Parses command line args."""
+  parser = argparse.ArgumentParser(
+      description='Parse arguments to launch the web report generation.')
```
erfanio, line 517: This should be a description of what the program does. Maybe something like "Report generation tool reads raw experiment output files and generates a report in the form of HTML files in a directory hierarchy."

erfanio (approved):
Looks good overall, just some nits.

report/web.py
DavidKorczynski: Sounds good, done

DonggeLiu (approved):
Unrelated to this PR, but `avahi` failed to include human-written fuzz target again:
https://llm-exp.oss-fuzz.com/Result-reports/ofg-pr/2024-06-19-352-dk1000-15-comparison/benchmark/output-avahi-avahi_dns_packet_new_reply/index.html

PR #357, opened 2024-06-19
introspector: add option for only collecting oracles with references

DavidKorczynski:
FI added an option for oracles to only return functions with cross references. This is convenient to avoid e.g. rewritten `main` functions, and is also convenient for cases where the prompt integrates features that are based on cross-references. This PR makes it possible to leverage this feature.

Ref: https://github.com/ossf/fuzz-introspector/pull/1614

### DIFF
--- a/data_prep/introspector.py
+++ b/data_prep/introspector.py
@@ -178,11 +178,12 @@ def _get_data(resp: Optional[requests.Response], key: str,
 
 def query_introspector_oracle(project: str, oracle_api: str) -> list[dict]:
   """Queries a fuzz target oracle API from Fuzz Introspector."""
-  resp = _query_introspector(oracle_api, {
-      'project': project,
-      'exclude-static-functions': ORACLE_AVOID_STATIC_FUNCTIONS,
-      'only-referenced-functions': ORACLE_ONLY_REFERENCED_FUNCTIONS,
-  })
+  resp = _query_introspector(
+      oracle_api, {
+          'project': project,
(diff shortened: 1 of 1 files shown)


DavidKorczynski:
gcbrun exp -n dk1000-15

DonggeLiu:
The report LGTM.
BTW, the build rate of [`avahi_string_list_add_pair `](https://llm-exp.oss-fuzz.com/Result-reports/ofg-pr/2024-06-19-357-dk1000-16-comparison/index.html) is 50%, which is a lot higher than the past (0-10%).
But I don't think it is due to this PR?

data_prep/introspector.py
```diff
@@ -42,6 +42,8 @@
 # to generate benchmarks.
 ORACLE_AVOID_STATIC_FUNCTIONS = bool(
     int(os.getenv('OSS_FUZZ_AVOID_STATIC_FUNCTIONS', '1')))
+ORACLE_ONLY_REFERENCED_FUNCTIONS = bool(
+    int(os.getenv('OSS_FUZZ_ONLY_REFERENCED_FUNCTIONS', '0')))
```
DonggeLiu, line 46: Shall we enable this by default?

DonggeLiu (approved):

data_prep/introspector.py
DavidKorczynski: I'd prefer this is opt-in so think we should keep it off at the moment

PR #360, opened 2024-06-19
[JVM] Fix bug in source file checking

arthurscchan:
For JVM projects, the Java source files are missing from the out directory in some cases. In that case, checking for interesting files will fail for many cases and skip most of the possible function targets. This PR fixes that by skipping the source file check for JVM projects.


arthurscchan:
No longer needed.

PR #361, opened 2024-06-19
Expand the benchmark list

arthurscchan:
In many projects, the fuzzing harnesses generation process is not good and result in many failing fuzzing harnesses. The current logic of oss-fuzz-gen first generates a limited list of benchmark from introspector oracles. Then it will try to generate  fuzzing harnesses for all targets in the benchmark list. This logic will significantly reduce the number of success fuzzing harnesses if the original benchmark list contains some bad targets.
This PR fixes that by adding a multiplier for the limit and generate a much larger benchmark list in the beginning to provides a bigger list of candidates. During the generation process, if the fuzzing harnesses build fails or crash during execution test, it will skip that target and try on the next target on list. This PR also adds in logic to determine if enough successful fuzzing harnesses has been generated. If yes, then it will exit and print the result to ensure the number of successful fuzzing harnesses does not passed the limit while ensuring higher ratio of success result.

**This PR is still in testing stage and are not yet ready for review.**

### DIFF
--- a/data_prep/introspector.py
+++ b/data_prep/introspector.py
@@ -37,6 +37,7 @@ T = TypeVar('T', str, list, dict)  # Generic type.
 
 TIMEOUT = 45
 MAX_RETRY = 5
+REDUNDANT_LIMIT_MULTIPLIER = 10
 
 # By default exclude static functions when identifying fuzz target candidates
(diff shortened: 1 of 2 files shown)


arthurscchan:
Not needed anymore.

PR #362, opened 2024-06-20
Move common logic for report generation into its own module.

erfanio:
No code change, just straight copy paste.

Moving the common classes into their own module was raised in #343.


report/common.py
```diff
@@ -0,0 +1,455 @@
...
+import yaml
+from google.cloud import storage
+
+import run_one_experiment
+from experiment import evaluator
+from experiment.workdir import WorkDirs
+
+MAX_RUN_LOGS_LEN = 16 * 1024
+
+TARGET_EXTS = ('.c', '.cc', '.cpp', '.cxx', '.c++', '.java', '.py')
```
DonggeLiu, line 33: Sorry I missed this last time: Would it be a good idea to import [this](https://github.com/google/oss-fuzz-gen/blob/main/data_prep/project_src.py#L30) and append `['.java', '.py']`?
So that we only define it once (to ensure it is consistent across the project and simplify future changes).

report/common.py
```diff
@@ -0,0 +1,455 @@
...
+  _gcs_client = None
+
+  @classmethod
+  def _get_gcs_client(cls):
+    """
+    Returns a cached storage client (a new one is created on first call.
+
+    A new client does authentication on first call, so caching the client will
+    same multiple authentication round trips to GCP.
+    """
```
DonggeLiu, line 104: Thanks for always making the comments so informative!

report/common.py
```diff
@@ -0,0 +1,455 @@
...
+    self._path = path
+    self._gcs_bucket: Optional[storage.Bucket] = None
+
+    if path.startswith('gs://'):
+      path = path.removeprefix('gs://')
+      self._gcs_bucket = FileSystem._get_gcs_client().bucket(path.split('/')[0])
+      self._path = '/'.join(path.split('/')[1:])
+
+  def listdir(self) -> List[str]:
+    """listdir returns a list of files and directories in path."""
```
DonggeLiu, line 121: BTW, I noticed that all doc-string starts the function name, is this auto-generated or based on some convention?
Thanks!

DonggeLiu (approved):

DonggeLiu:
The report looks good:
https://llm-exp.oss-fuzz.com/Result-reports/ofg-pr/2024-06-20-362-abc-comparison/index.html

report/common.py
erfanio: It's just a convention I like. Golang enforces function/module comments to start with the name which IMO generally makes golang docs more clear because it forces the sentences to focus on what the function/module does.

report/common.py
erfanio: No worries :)

report/common.py
erfanio: Done!

DonggeLiu:
The report looks good
https://llm-exp.oss-fuzz.com/Result-reports/ofg-pr/2024-06-24-362-abc-comparison/index.html

Shall we merge this?

PR #364, opened 2024-06-21
Prompter script fixes

inferno-chromium:
* Fix max token limit based on latest VertexAI docs
* Fix temperature max in help message
* Fix hardcoded responses dir
* Update to use latest Gemini 1.5 preview in models.py
* Clean requirements. Add GitPython dependency for upcoming script.


experimental/manual/prompter.py
```diff
@@ -32,14 +32,17 @@ def parse_args() -> argparse.Namespace:
       '--temperature',
       type=float,
       default=TEMPERATURE,
-      help=('A value between 0 and 2 representing the variety of the targets '
+      help=('A value between 0 and 1 representing the variety of the targets '
```
DonggeLiu, line 35: Thanks, I probably should have made this help message more general.
This value depends on the model we use. For example, the max value is 2 for `pro-1.5`, but 1 for `ultra`/`ultra-vision`:
https://pantheon.corp.google.com/vertex-ai/generative/multimodal/create/chat?hl=en&e=-13802955&mods=logs_tg_prod&project=oss-fuzz

Same for the `MAX_TOKENS`.
I will fix them later.

DonggeLiu (approved):
Thanks!

experimental/manual/prompter.py
inferno-chromium: Oops, I shouldn't have done auto-merge. Can you please fix it in another CL, unsure if we need MAX_TOKENS in prompter.py since models.py already has those constants ?

PR #377, opened 2024-06-24
Fix for Incorrect File Extension in Generated Report Links

MarkLee131:
Currently, when generating reports using the command outlined [here](https://github.com/google/oss-fuzz-gen/blob/main/USAGE.md#results-report), the sample pages are named solely by `sample_id`, with no file extension. This results in the browser prompting users to download the file, named like '01', instead of displaying it as an HTML page when they click the link. This behavior occurs because the links lack the `.html` suffix, which is crucial for web browsers to recognize and render the content appropriately.

image

To address this issue, I have modified the code in `.report/web.py` and updated the templates in `report/template/benchmark.html` to automatically append the .html suffix to the filenames. This change ensures that the generated links function correctly, allowing the HTML pages to be displayed directly in the browser.

image

Hope it helps.

### DIFF
--- a/report/templates/benchmark.html
+++ b/report/templates/benchmark.html
@@ -36,7 +36,7 @@ td, th {
     </tr>
     {% for sample in samples %}
     <tr>
-        <td><a href="../../sample/{{ benchmark|urlencode }}/{{ sample.id }}">{{ sample.id }}</a></li></td>
+        <td><a href="../../sample/{{ benchmark|urlencode }}/{{ sample.id }}.html">{{ sample.id }}</a></li></td>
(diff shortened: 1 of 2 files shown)


DonggeLiu:
Thanks for reporting and fixing this, @MarkLee131!
Inviting @erfanio  to review this as he is the expert : )

We did not notice this in the past likely because it only affects external users somehow.
It's great that you pointed this out and submitted this PR.
I've created a new PR #378 and rebased your branch on it so that we can test it before merging :)

erfanio (commented):
LGTM.

PR #382, opened 2024-06-25
Revive C-specific prompt and examples

DonggeLiu:
Observing many compatibility issues caused by building a C++ fuzz target with C projects, I want to test whether separating them can lead to a higher compilation rate.

This PR experiment uses C-specific prompts and examples for C projects.
It also uses the fix in #12116


DonggeLiu:
Seeing improvements on both `hiredis` benchmarks and consistent good performance on `inih`.
`flex` failed due to the non-existent `main2` function.

DonggeLiu:
https://llm-exp.oss-fuzz.com/Result-reports/scheduled/2024-06-26-weekly-all/

PR #388, opened 2024-06-26
JVM: Fix bug on inner class name handling

arthurscchan:
This PR fixes a bug in JVM full class name handling process for inner class.


llm_toolkit/prompt_builder.py
```diff
@@ -602,6 +602,7 @@ def _format_target_method(self, signature: str) -> str:
   def _format_import_mapping(self, full_class_name: str) -> str:
     """Formats the import mapping row on the prompt template."""
     class_name = full_class_name.rsplit('.')[-1]
+    full_class_name = full_class_name.split('$')[0]
```
DavidKorczynski, line 605: Could we add a comment? Perhaps with an example

DavidKorczynski (changes requested):

DavidKorczynski (approved):

PR #390, opened 2024-06-26
JVM: Fix non-exist method call

arthurscchan:
For unknown reasons, when a random java.lang.Object instance is needed for method arguments, the LLM model always generates the needed instance with `FuzzedDataProvider::consumeObject()` or `FuzzedDataProvider::getObject()`. These two methods are not valid and never exist. Similar case happens for integer instance. the LLM model use a non-exist `FuzzedDataProvider::consumeInt(int)` for integer data generation.

This PR adds a temporary fix to replace the call to these methods with a valid one.

The `FuzzedDataProvider::consumeObject()` or `FuzzedDataProvider::getObject()` are replaced by `FuzzedDataProvider::consumeString(int)` as String is one of the subclasses for the Object instance.

The `FuzzedDataProvider::consumeInt(int)` is replaced by `FuzzedDataProvider::consumeInt(int, int)` for getting the correct internet within range.

### DIFF
--- a/llm_toolkit/prompt_builder.py
+++ b/llm_toolkit/prompt_builder.py
@@ -754,24 +754,22 @@ class DefaultJvmTemplateBuilder(PromptBuilder):
 
   def post_process_generated_code(self, generated_code: str) -> str:
     """Allows prompt builder to adjust the generated code."""
-    # For unknown reasons, the LLM model keeps using wrong
-    # FuzzedDataProvider::consumeObject() or FuzzedDataProvider::getObject()
-    # methods to generate random java.lang.Object instance. These methods are
-    # not valid in FuzzedDataProvider. The fixes here change the calling of
-    # them to FuzzedDataProvider::consumeString(int) instead.
+    # From observation, the LLM model keeps using wrong method calls including
+    # FuzzedDataProvider::consumeObject() or FuzzedDataProvider::getObject() or
+    # FuzzedDataProvider::consumeInt(int) to generate random Object / Integer
+    # instance. These methods are not valid in FuzzedDataProvider.
+
+    # The fixes here change the calling of data.consumeObject() and data.getObject()
+    # to data.consumeString(int)
     generated_code = generated_code.replace(
         'data.consumeObject()', 'data.consumeString(data.remainingBytes()/2)')
     generated_code = generated_code.replace(
         'data.getObject()', 'data.consumeString(data.remainingBytes()/2)')
 
-    # For unknown reasons, the LLM model keeps using wrong
-    # FuzzedDataProvider::consumeInt(int) method to generate random int or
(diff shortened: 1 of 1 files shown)


DavidKorczynski:
Nice, It's nice to see more use of the post processing logic -- I think we could expand on it quite a bit

llm_toolkit/prompt_builder.py
```diff
@@ -751,6 +752,29 @@ def build_triager_prompt(self, benchmark: Benchmark, driver_code: str,
...
+        'data.consumeObject()', 'data.consumeString(data.remainingBytes()/2)')
+    generated_code = generated_code.replace(
+        'data.getObject()', 'data.consumeString(data.remainingBytes()/2)')
+
+    # For unknown reasons, the LLM model keeps using wrong
+    # FuzzedDataProvider::consumeInt(int) method to generate random int or
+    # java.lang.Integer instance. This method is not valid. The fixes here
+    # change the calling of that to FuzzedDataProvider::consumeInt(int,int)
+    # instead.
+    for item in re.findall(r'(data\.consumeInt\(([0-9]*)\))', generated_code):
```
DavidKorczynski, line 772: can we have a better word than `item` here?

llm_toolkit/prompt_builder.py
```diff
@@ -751,6 +752,29 @@ def build_triager_prompt(self, benchmark: Benchmark, driver_code: str,
...
+    """Allows prompt builder to adjust the generated code."""
+    # For unknown reasons, the LLM model keeps using wrong
+    # FuzzedDataProvider::consumeObject() or FuzzedDataProvider::getObject()
+    # methods to generate random java.lang.Object instance. These methods are
+    # not valid in FuzzedDataProvider. The fixes here change the calling of
+    # them to FuzzedDataProvider::consumeString(int) instead.
+    generated_code = generated_code.replace(
+        'data.consumeObject()', 'data.consumeString(data.remainingBytes()/2)')
+    generated_code = generated_code.replace(
+        'data.getObject()', 'data.consumeString(data.remainingBytes()/2)')
```
DavidKorczynski, line 765: Could we have some samples of this substitution in the PR description? Why are `data.consumeString` the right substition?

llm_toolkit/prompt_builder.py
```diff
@@ -751,6 +752,29 @@ def build_triager_prompt(self, benchmark: Benchmark, driver_code: str,
...
+
+    # For unknown reasons, the LLM model keeps using wrong
```
DavidKorczynski, line 767: Could we avoid using the same comment twice? I'm also not sure what `For unknown reasons` adds -- perhaps just based on experiments, or many observations shows or something like that.

DavidKorczynski (changes requested):

DavidKorczynski (approved):

PR #393, opened 2024-06-27
Add `extern` only when a c project has c++ fuzz targets

DonggeLiu:
Almost all regressions in https://github.com/google/oss-fuzz-gen/pull/382#issuecomment-2190303945 are due to C projects using C++ fuzz targets.
This PR resolves that.


DonggeLiu:
Seeing more buildable benchmarks with the fix:
image

jonathanmetzman (commented):
LGTM

PR #395, opened 2024-06-28
JVM: Fix name of fuzzing harness

arthurscchan:
This PR includes the needed name in the prompts to allow the LLM model to generate a fuzzing harness with the correct name. This removes the need to replace the name afterwards. This change could avoid errors when the class name is used by other logic in the generated fuzzing harness class which has not been updated.

### DIFF
--- a/experiment/evaluator.py
+++ b/experiment/evaluator.py
@@ -197,17 +197,6 @@ class Evaluator:
 
     shutil.copytree(existing_project_path, generated_project_path)
 
-    # Fix public java class name in target_file
-    if self.benchmark.language == 'jvm':
-      with open(target_file, 'r') as file:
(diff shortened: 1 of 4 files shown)


DavidKorczynski:
Let's first land https://github.com/google/oss-fuzz-gen/pull/399 so we can run some more thorough experiments on this

DavidKorczynski:
This seems to be causing a lot of issues: From a run before this PR:
Screenshot 2024-07-01 100141


From a run after this PR (after 90% is Done):
Screenshot 2024-07-01 100255

DavidKorczynski (changes requested):
highlighted above

DavidKorczynski (approved):

PR #396, opened 2024-06-28
JVM: Revamp JVM prompts

arthurscchan:
This PR is the first of a list of PRs to revamp the JVM prompts to make the LLM model generate better fuzzing harnesses. This PR targets the base problem description and base requirements of the JVM prompts.


DavidKorczynski (changes requested):
Let's first land https://github.com/google/oss-fuzz-gen/pull/399 so we can run some more thorough experiments on this

DavidKorczynski (approved):

PR #398, opened 2024-06-29
Add the support for GPT-4o model

MarkLee131:
1. add the model support for `gpt-4o` (https://platform.openai.com/docs/models/gpt-4o) and 
2. update the token estimation logic to support this model.


llm_toolkit/models.py
```diff
@@ -198,7 +198,10 @@ def estimate_token_num(self, text) -> int:
     """Estimates the number of tokens in |text|."""
     # https://cookbook.openai.com/examples/how_to_count_tokens_with_tiktoken
     try:
-      encoder = tiktoken.encoding_for_model(self.name)
+      if 'gpt-4' in self.name:  # gpt-4 and gpt-4o
+        encoder = tiktoken.encoding_for_model('gpt-4')
```
DavidKorczynski, line 202: What is the reason for not specifying `gpt-4o` directly? https://github.com/openai/tiktoken/blob/c0ba74c238d18b4824c25f3c27fc8698055b9a76/tiktoken/model.py#L22-L23

DavidKorczynski (changes requested):

llm_toolkit/models.py
MarkLee131: Hi, @DavidKorczynski. I see the link you provided claimed the support for `gpt-4o`. But it seems now it gets the error below if running the code `encoder = tiktoken.encoding_for_model(self.name)` for `gpt-4o` directly:
```
Python 3.11.7 (main, Dec  8 2023, 18:56:57) [GCC 9.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import tiktoken
>>> encoder = tiktoken.encoding_for_model("gpt-4o")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/kaixuan/FDG_LLM/oss-fuzz-gen/.venv/lib/python3.11/site-packages/tiktoken/model.py", line 97, in encoding_for_model
    return get_encoding(encoding_name_for_model(model_name))
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/kaixuan/FDG_LLM/oss-fuzz-gen/.venv/lib/python3.11/site-packages/tiktoken/model.py", line 84, in encoding_name_for_model
    raise KeyError(
KeyError: 'Could not automatically map gpt-4o to a tokeniser. Please use `tiktoken.get_encoding` to explicitly get the tokeniser you expect.'
```
you can confirm it again on your env.

DavidKorczynski: Likely because of https://github.com/google/oss-fuzz-gen/blob/3d209143f94acf6a1863acdaed69c03734e1602c/requirements.in#L14 -- let's bump?

MarkLee131: Ok~ 
I have tried `tiktoken 0.7.0`, it works now if directly running `encoder = tiktoken.encoding_name_for_model("gpt-4o")` after upgrading the tiktoken.

MarkLee131: @DavidKorczynski Do I need to recreate a PR to directly use `tiktoken` to get the encoding for gpt-4o? Or we can continue use this PR (if you can revise the code)?

MarkLee131: Got it. I have finished it~

requirements.txt
```diff
@@ -1,5 +1,5 @@
 #
-# This file is autogenerated by pip-compile with Python 3.12
+# This file is autogenerated by pip-compile with Python 3.11
```
DavidKorczynski, line 2: Are you sure changing the version doesn't cause regressions?

DavidKorczynski (changes requested):

requirements.txt
DavidKorczynski: In addition, do we really need to update all the libs for a tiktoken update?

MarkLee131: Hi, I've actually been running oss-fuzz-gen with python 3.11 and it should be fine.

Meanwhile, since some packages may be dependent on tiktoken, for insurance purposes, I just tried to update all the dependencies directly. :)

MarkLee131: Hi @DavidKorczynski, I have updated the dependencies by using python 3.12 to avoid regressions. And I found the `pandas` and `astroid` should be updated if we want to update tiktoken to 0.7.0. The update was tested on my local env and works well. :) 
you can check it again. Thanks.

mihaimaruseac (approved):
This looks reasonable to me

mihaimaruseac:
Tahnks @DavidKorczynski for the initial review. I was OOO and missed this.

@MarkLee131 , can you please rebase this PR on main branch?

mihaimaruseac (approved):
Thank you!

DavidKorczynski (changes requested):
let's just test this through as a previous state of this PR caused the CI experiments to fail https://github.com/google/oss-fuzz-gen/pull/433

DavidKorczynski (approved):

PR #399, opened 2024-06-30
benchmarks: add extended JVM set

DavidKorczynski:
The JVM benchmark set has only 11 projects. This extends it to around 30 projects. The goal is to facilite more thorough experiments since we're improving the JVM logic a fair amount currently.


DavidKorczynski:
/gcbrun exp -n dk-test-jvm123 -m vertex_ai_gemini-1-5 -b jvm

DonggeLiu (approved):

PR #401, opened 2024-06-30
experimental: make it possible to pick harness generation heuristics

DavidKorczynski:
Makes it easier to focus during development and make experiments targeted

### DIFF
--- a/experimental/c-cpp/manager.py
+++ b/experimental/c-cpp/manager.py
@@ -20,7 +20,7 @@ import os
 import shutil
 import subprocess
 from abc import abstractmethod
-from typing import Any, Dict, List, Optional, Tuple
+from typing import Any, Dict, List, Optional, Tuple, Type
 
 import build_generator
 import cxxfilt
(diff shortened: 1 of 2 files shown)


experimental/c-cpp/manager.py
```diff
@@ -137,12 +137,16 @@ def add_to_source_cache(heuristic_name, target_func, fuzzer_source):
 
 class FuzzHeuristicGeneratorBase:
   """Base class for fuzzer heuristics generator."""
+  language = ''
+  name = ''
```
DonggeLiu, line 141: nit: Would it be better to initialize them in `__ini__()`?

experimental/c-cpp/manager.py
```diff
@@ -1230,6 +1235,24 @@ def load_introspector_report():
...
+      FuzzerGenHeuristic6, FuzzerGenHeuristic5, FuzzerGenHeuristic4,
+      FuzzerGenHeuristic3, FuzzerGenHeuristic2, FuzzerGenHeuristic1
+  ]
+  if heuristics_to_use == 'all':
+    heuristics_to_apply = all_possible_heuristics
+  else:
+    possible_heuristics = set(heuristics_to_use.split(','))
+    for heuristic in all_possible_heuristics:
+      if heuristic.name in possible_heuristics:
+        heuristics_to_apply.append(heuristic)
```
DonggeLiu, line 1252: Just out of curiosity: Does the order in all_possible_heuristics reflect priority? E.g., heuristic 6 has a higher/lower priority than 5.

DonggeLiu (approved):

experimental/c-cpp/manager.py
DavidKorczynski: Not in this case because we query the language before initializing an object of the class.

experimental/c-cpp/manager.py
DavidKorczynski: No, the numbering is purely for ID purposes.

PR #403, opened 2024-07-01
Ask LLM to remove `#include 'non-existent filepath` and suggest alternatives

DonggeLiu:
LLM insists on including non-existent files, likely because LLM _thinks_ that's necessary (for the function-under-test) or does not know what's the correct alternatives
```
isc_lex_gettoken.c:5:10: fatal error: 'isc/boolean.h' file not found
    5 | #include <isc/boolean.h>
      |          ^~~~~~~~~~~~~~~
```

This PR specifically asks LLM to remove that incorrect statement (and suggest any existent alternatives).

### DIFF
--- a/data_prep/introspector.py
+++ b/data_prep/introspector.py
@@ -228,6 +228,15 @@ def query_introspector_cfg(project: str) -> dict:
   return _get_data(resp, 'project', {})
 
 
+def query_introspector_source_file_path(project: str, func_sig: str) -> str:
+  """Queries FuzzIntrospector API for file path of |func_sig|."""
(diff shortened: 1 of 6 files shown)


DonggeLiu:
There are more TBD, but I consider this pretty good improvement as the first iteration:
https://llm-exp.oss-fuzz.com/Result-reports/ofg-pr/2024-07-01-403-dg-2-comparison/index.html

Baseline:
https://llm-exp.oss-fuzz.com/Result-reports/scheduled/2024-07-01-daily-comparison/index.html

PR #405, opened 2024-07-01
Fix regex to parse function name when checking if it is called in the fuzz target

DonggeLiu:
Function `operator` is a special case that can be followed by special chars like `++`.
This PR includes that in the regex.

For example, the following line uses `++`, but was not parsed by the current regex:
```C++
// Iterator<std::__1::vector<int, std::__1::allocator<int> >&, std::__1::vector<bool, std::__1::allocator<bool> >&> & iter::impl::Compressed<std::__1::vector<int, std::__1::allocator<int> >&, std::__1::vector<bool, std::__1::allocator<bool> >&>::Iterator<std::__1::vector<int, std::__1::allocator<int> >&, std::__1::vector<bool, std::__1::allocator<bool> >&>::operator++(Iterator<std::__1::vector<int, std::__1::allocator<int> > &, std::__1::vector<bool, std::__1::allocator<bool> > &> *);
it++;
```

### DIFF
--- a/experiment/builder_runner.py
+++ b/experiment/builder_runner.py
@@ -138,9 +138,7 @@ class BuilderRunner:
       return func_sig
 
     function_name = match.group(1).strip()
-    if function_name.startswith('operator'):
-      return function_name.removeprefix('operator')
-    return function_name
+    return function_name.removeprefix('operator')
 
   def _contains_target_jvm_method(self, target_path: str) -> bool:
     """Validates if the LLM-generated code contains the target jvm methods."""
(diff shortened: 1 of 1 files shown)


DonggeLiu:
This fix `operator++` in [`cppitertools`](https://llm-exp.oss-fuzz.com/Result-reports/ofg-pr/2024-07-01-405-dg-comparison/benchmark/output-cppitertools-_zn4iter4impl10compressedirnst3__16vectoriins2_9allocatoriieeeerns3_ibns4_ibeeeee8iteratoris7_sa_epp/index.html) without affecting others.
The benchmark still failed to build, likely because of the project's [recent build failures](https://oss-fuzz-build-logs.storage.googleapis.com/index.html#cppitertools) in OSS-Fuzz.
Strange that the some of the [other benchmarks](https://llm-exp.oss-fuzz.com/Result-reports/ofg-pr/2024-07-01-405-dg-comparison/benchmark/output-cppitertools-_zn4iter4impl10combinatorirnst3__112basic_stringicns2_11char_traitsiceens2_9allocatoriceeeee8iterato/index.html) can build.

PR #409, opened 2024-07-01
Ask LLM to not use FDP in C fuzz targets

DonggeLiu:
This sometimes happens when a C++ project has a C fuzz target.
The error message may confuse LLM, so I added direct instructions.


jonathanmetzman (commented):
LGTM

Do we trust the LLM to do this?

DonggeLiu:
> Do we trust the LLM to do this?

I am unsure if there is a better solution: Is there a programmatical way to 1) identify all variable initializations using FDP and 2) replace them with the 'ordinary' alternative?

PR #410, opened 2024-07-01
JVM: Add exceptions in jvm prompts

arthurscchan:
This PR adds additional exceptions for each target into the prompts to allow the LLM generate good exception handling in the fuzzing harnesses.


DavidKorczynski:
Please rebase this to catch the changes of https://github.com/google/oss-fuzz-gen/pull/412 which should make it easier to pass in data from benchmarks

DavidKorczynski (approved):

Issue #356, opened 2024-06-19
Maximize overall coverage by showing overlapping benchmarks


DonggeLiu:
Extended from #354.
Context:
Currently, OSS-Fuzz-Gen selects functions with low coverage. However, some selected functions may execute the same code blocks of the project, leading to minimal project-level coverage improvement.

Goal:
Displaying the "per-project coverage increment" alongside the "per-benchmark" coverage increment will help us identify instances where overlapping code blocks are being tested. This insight will highlight cases where we should re-select benchmarks to enhance overall efficiency.

DonggeLiu:
ACK, I will close this then.

DavidKorczynski:
I still think this is a good thing to pursue though! Introspector's "Optimal targets" is just one heuristic trying to solve a similar problem, and the feature suggested in this issue is also relevant now that we can extract target functions from a diverse set of oracles.

PR #415, opened 2024-07-02
introspector: add optimal target API

DavidKorczynski:
Create benchmarks using FI's optimal target analysis.

Ref:
https://github.com/google/oss-fuzz-gen/issues/356#issuecomment-2178716065


DonggeLiu (approved):

DonggeLiu (approved):
Approved, but could you please resolve the conflict?

PR #416, opened 2024-07-02
JVM: Add static information of target method in prompts

arthurscchan:
This PR adds objection creation requirement according to the target method's static property.


DavidKorczynski (changes requested):
Do I get this right that this is ultimately dependent on https://github.com/ossf/fuzz-introspector/pull/1648 -- i.e. we have to bump introspector first on oss-fuzz before noticing the change having effect here (since the experiments here do not rely on the locally build base-builder images)?

DavidKorczynski (changes requested):
I think we should watch out with the name `is_static` here -- it's only used for `jvm` at this stage as and without knowing this (which is not apparent) then it can quickly become confusing. We have a couple of options: (1) make it general, i.e. also for c/c++; (2) change then ame to `is_jvm_static` or so; (3) separate benchmarks into different classes, e.g. `C_BENCHMARK`, `JVM_BENCHMARK`, `CPP_BENCHMARK`.

I think all three holds merit, but, for now I suggest we go with option 2 to move forward with this

DavidKorczynski (approved):

PR #418, opened 2024-07-02
introspector: add api to get header files where a function is likely declared

DavidKorczynski:
Also uses it in the C-targeted prompt. In essence an alternative to having this in the prompt is avoiding having these in the prompt and just including the header files mechanically, but I'll explore that later for now.

This has been tested locally on a myriad of C projects, so suggest leaving the CI experimentation for now (I'll add the ability to apply a given model shortly)

### DIFF
--- a/data_prep/introspector.py
+++ b/data_prep/introspector.py
@@ -60,6 +60,7 @@ INTROSPECTOR_FUNC_SIG = ''
 INTROSPECTOR_ADDR_TYPE = ''
 INTROSPECTOR_ALL_HEADER_FILES = ''
 INTROSPECTOR_ALL_FUNC_TYPES = ''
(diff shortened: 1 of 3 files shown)


DonggeLiu:
This is a pretty good direction, could you please explain the underlying heuristic of the API?
I might be able to replace [my function](https://github.com/google/oss-fuzz-gen/blob/338bbab425f3a307e865840f45b62e4423d6c26b/data_prep/project_context/context_introspector.py#L244-L251) with it.

DonggeLiu:
> suggest leaving the CI experimentation for now (I'll add the ability to apply a given model shortly)

Yep, please do; I am curious to see its improvements.

DonggeLiu (approved):

PR #420, opened 2024-07-03
Add instruction for a C++ fuzz target includes a C file from a C++ project

DonggeLiu:
Example:
```
In file included from /src/libvpx/examples/vpx_dec_fuzzer.cc:28:
/src/libvpx/vp9/decoder/vp9_decodeframe.c:69:11: error: cannot initialize a variable of type 'TX_MODE' with an rvalue of type 'int'
   69 |   TX_MODE tx_mode = vpx_read_literal(r, 2);
```

### DIFF
--- a/llm_toolkit/code_fixer.py
+++ b/llm_toolkit/code_fixer.py
@@ -31,6 +31,9 @@ ERROR_LINES = 20
 NO_MEMBER_ERROR_REGEX = r"error: no member named '.*' in '([^':]*):?.*'"
 FILE_NOT_FOUND_ERROR_REGEX = r"fatal error: '([^']*)' file not found"
 UNKNOWN_TYPE_ERROR = 'error: unknown type name'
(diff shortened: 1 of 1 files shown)


DonggeLiu:
@jonathanmetzman Do you happen to know if there is a solution to this error?
This error is caused by including a C file in the C++ fuzz target.
C allows type casting, but C++ does not, hence the error happened on the `#include` statement.
The original fuzz target is in C++ and so is the project-under-test (`openssl`).
But that specific file `vp9_decodeframe.c` is in C:
```
In file included from /src/libvpx/examples/vpx_dec_fuzzer.cc:28:
/src/libvpx/vp9/decoder/vp9_decodeframe.c:69:11: error: cannot initialize a variable of type 'TX_MODE' with an rvalue of type 'int'
   69 |   TX_MODE tx_mode = vpx_read_literal(r, 2);
```

I tried adding `extern "C"` around the `#include` statement, but it does not work:
```cpp
extern "C" {
    #include "/src/libvpx/vp9/decoder/vp9_decodeframe.c"
}
```

PR #421, opened 2024-07-03
report: add option for continuous report serving

DavidKorczynski:
When running experiments locally the workflow is to continuously rebuild the static pages while having a http server launched. This is a bit inconvenient. This commit adds an option to the web module such that both a http server is launched and the static generator continues to generate files as long as the http server is launched.

### DIFF
--- a/report/web.py
+++ b/report/web.py
@@ -174,7 +174,8 @@ class GenerateReport:
                     sample.id, e)
 
 
-def generate_report(args):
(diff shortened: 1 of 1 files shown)


DavidKorczynski:
`python3 -m report.web -r ./results -o result-report-1 -s` will spawn a server and run it until the process is stopped

DonggeLiu (approved):

PR #422, opened 2024-07-03
introspector: show url when timeouts happen

DavidKorczynski:
I've run into several regressions during API development that's causing timeout issues, and having the logs ready makes debugging a lot easier.


DonggeLiu (approved):
Just a minor note: Google cloud log explorer will likely break multi-line logs like this into multiple logs so we normally prefer single line (until we have the time to change to google.cloud.log).

DavidKorczynski:
> Just a minor note: Google cloud log explorer will likely break multi-line logs like this into multiple logs so we normally prefer single line (until we have the time to change to google.cloud.log).

Gotcha -- I'll keep it as is now because it follows the other loggings of similar nature, but we can revisit all the logs in case we want to create one-liners

PR #426, opened 2024-07-03
data_prep: Add query for /api/function-with-matching-return-type

arthurscchan:
This PR adds a new introspector query to the function-with-matching-return-type api which retrieves a list of functions and constructors that returns the needed type of object.

### DIFF
--- a/data_prep/introspector.py
+++ b/data_prep/introspector.py
@@ -319,10 +319,8 @@ def query_introspector_jvm_source_path(project: str) -> List[str]:
 
 
 def query_introspector_matching_function_type(project: str,
-                                              return_type: str) -> List[str]:
-  """
-    Queries for all functions that returns a given type in a given project.
-  """
+                                              return_type: str) -> List[Dict]:
+  """Queries for all functions that returns a given type in a given project."""
   simple_types_should_not_process = [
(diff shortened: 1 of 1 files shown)


data_prep/introspector.py
```diff
@@ -308,12 +312,56 @@ def query_introspector_sample_xrefs(project: str, func_sig: str) -> List[str]:
...
   resp = _query_introspector(INTROSPECTOR_ALL_JVM_SOURCE_PATH,
                              {'project': project})
   return _get_data(resp, 'src_path', [])
 
 
+def query_introspector_matching_function_type(project: str,
+                                              return_type: str) -> List[str]:
+  """
+    Queries for all functions that returns a given type in a given project.
+  """
```
DavidKorczynski, line 325: Could we fix the function docs to be similar style to the rest of the module?

DavidKorczynski (changes requested):

data_prep/introspector.py
DavidKorczynski: i.e. starting on same line

DavidKorczynski (changes requested):
On second thought, I think there's too much code duplication here. Let's add an argument to control the return value.

PR #431, opened 2024-07-04
Update benchmarks

DonggeLiu:
1. Allow generating benchmarks with multiple oracles, prioritizing by oracle order.
2. Update benchmarks using the new oracles: `optimal-targets`, `all-public-candidates`.

@DavidKorczynski It appears that `optimal-targets` does not work well with many benchmarks.
E.g., empty list for `libxml2`, 'Project not in the database' for `avahi`. Is this expected?
```
https://introspector.oss-fuzz.com/api/optimal-targets?project=libxml2&exclude-static-functions=True&only-referenced-functions=False&only-with-header-file-declaration=True
{'functions': [], 'result': 'success'}
https://introspector.oss-fuzz.com/api/optimal-targets?project=avahi&exclude-static-functions=True&only-referenced-functions=False&only-with-header-file-declaration=True
{'msg': 'Project not in the database', 'result': 'error'}
```


data_prep/introspector.py
```diff
@@ -747,7 +755,8 @@ def _parse_arguments() -> argparse.Namespace:
   parser.add_argument('-t',
                       '--target-oracle',
                       type=str,
-                      default='far-reach-low-coverage',
+                      nargs='+',
+                      default=['optimal-targets', 'all-public-candidates'],
```
DavidKorczynski, line 759: I'd keep `far-reach-low-coverage` as the primary benchmark -- there will still be improvements with respect to build rate from the use of `exclude-static-functions=True&only-referenced-functions=False&only-with-header-file-declaration=True`

`optimal-targets` is a bit more intricate and has far fewer functions per project. For some there may be no due to additional filtering from `exclude-static-functions=True&only-referenced-functions=False&only-with-header-file-declaration=True`

DavidKorczynski (changes requested):

data_prep/introspector.py
DonggeLiu: I see, I will use `['optimal-targets', 'far-reach-low-coverage']` then.

> all-public-candidates is primarily a JVM oracle. 

Could you please reflect that in the name of `INTROSPECTOR_ORACLE_ALL_CANDIDATES`, like `INTROSPECTOR_ALL_JVM_SOURCE_PATH`?

DavidKorczynski (changes requested):
>   2. Update benchmarks using the new oracles: optimal-targets, all-public-candidates.

I understood it as you'd like to also add new benchmarks? At the moment it's only removing `all`

DonggeLiu:
> > 2. Update benchmarks using the new oracles: optimal-targets, all-public-candidates.
> 
> I understood it as you'd like to also add new benchmarks? At the moment it's only removing `all`

Yep, I need to confirm which oracles to use before generate new benchmarks : )

DavidKorczynski:
> Yep, I need to confirm which oracles to use before generate new benchmarks : )

Do: `far-reach-but-low-coverage`, `optimal-targets`, `easy-params-far-reach`

DonggeLiu:
Looking OK:
https://llm-exp.oss-fuzz.com/Result-reports/ofg-pr/2024-07-05-431-dg-comparison/index.html

DavidKorczynski (approved):

DonggeLiu:
Looking good:
https://llm-exp.oss-fuzz.com/Result-reports/scheduled/2024-07-06-weekly-all/

Will select a new comparison set for daily improvement/regression monitoring.

DonggeLiu:
The new `comparison` set is selected from the [07-06 experiment](https://llm-exp.oss-fuzz.com/Result-reports/scheduled/2024-07-06-weekly-all/), containing benchmarks with:
1. Top performance
2. Mid/Low performing one
3. 0 build rate (majority).

PR #432, opened 2024-07-04
Fix bug in file search iteration logic.

inferno-chromium:
Search for file in git commit tree at the particular commit, instead of at tip of tree trunk.


DavidKorczynski (approved):

DonggeLiu (commented):
LGTM

PR #441, opened 2024-07-06
experimental: ignore generated status files

DavidKorczynski:
The from-scratch logic creates *.txt files during analysis and also often involves .txt files for input, these should not be committed.


experimental/c-cpp/.gitignore
```diff
@@ -0,0 +1 @@
+*.txt
```
DonggeLiu, line 1: This will also match and ignore new prompt files (https://github.com/google/oss-fuzz-gen/tree/main/prompts/template_xml).
Could the pattern be more specific, or could the from-scratch logic create these files using a different extension?

Alternatively, we can change all prompts to `*.prompt`, which will require changing many places (e.g., read prompts, create reports, etc.).

DonggeLiu (changes requested):

experimental/c-cpp/.gitignore
DavidKorczynski: The `.gitignore` works relative to the directory, so it will only catch files in `experimental/c-cpp/`

DonggeLiu: Oh sorry I thought this is the one under root.

DonggeLiu (approved):

PR #444, opened 2024-07-06
report: fix presubmit warnings

DavidKorczynski:
There's a handful of warnings when changing things in report/common.py and this fixes them all.


DonggeLiu (approved):
Thanks for doing this in a separate PR!

Issue #450, opened 2024-07-08
Merge experimental/c-cpp with core


DavidKorczynski:
Merge the logic that is used to generate oss-fuzz integrations from scratch (https://github.com/google/oss-fuzz-gen/tree/main/experimental/c-cpp) with the core oss-fuzz-gen logic. The primary goal is to enable using the core prompt generation logic in combination with auto-building procedures and to make sure improvements in one benefits both.

DavidKorczynski:
Ref for merging the code bases: https://github.com/google/oss-fuzz-gen/pull/440#discussion_r1667672538

PR #454, opened 2024-07-08
experimental: refactor to simplify and compartmentalise

DavidKorczynski:
Refactors the from-scratch logic by:
- Adding a container for build workers
- Removing unused code
- Facelifting string formatting since it was added to the CI

Note, in the `manager` and `build_generator` we can't remove the use of `List` and `Dict` since it relies on the Python from OSS-Fuzz, which hasn't migrated to `list` and `dict` yet.

This is working towards https://github.com/google/oss-fuzz-gen/issues/450

### DIFF
--- a/experimental/c-cpp/build_generator.py
+++ b/experimental/c-cpp/build_generator.py
@@ -26,6 +26,17 @@ import manager
 logger = logging.getLogger(name=__name__)
 
 
+class BuildWorker:
+  """Keeper of data on auto generated builds."""
+
+  def __init__(self):
+    self.build_script: str = ''
+    self.build_suggestion: str = ''
+    self.build_directory: str = ''
+    self.executable_files_build: Dict[str, List[str]] = {}
+    self.base_fuzz_build: bool = False
+
+
 ############################################################
 #### Logic for auto building a given source code folder ####
 ############################################################
@@ -708,7 +719,7 @@ def extract_build_suggestions(
 
 def raw_build_evaluation(
     all_build_scripts: List[Tuple[str, str, AutoBuildContainer]],
-    initial_executable_files: Dict[str, List[str]]) -> Dict[str, Any]:
+    initial_executable_files: Dict[str, List[str]]) -> Dict[str, BuildWorker]:
   """Run each of the build scripts and extract any artifacts build by them."""
   build_results = {}
   for build_script, test_dir, build_suggestion in all_build_scripts:
@@ -727,22 +738,18 @@ def raw_build_evaluation(
     # to running the build.
     binary_files_build = get_all_binary_files_from_folder(test_dir)
 
-    new_binary_files = {
-        'static-libs': [],
-        'dynamic-libs': [],
-        'object-files': []
-    }
-    for key, bfiles in binary_files_build.items():
-      for bfile in bfiles:
-        if bfile not in initial_executable_files[key]:
-          new_binary_files[key].append(bfile)
-
-    logger.info('Static libs found %s', str(new_binary_files))
-
     # binary_files_build['static-libs'])
-    build_results[test_dir] = {
-        'build-script': build_script,
-        'executables-build': binary_files_build,
-        'auto-build-setup': (build_script, test_dir, build_suggestion)
-    }
+    #build_results[test_dir] = {
+    #    'build-script': build_script,
+    #    'executables-build': binary_files_build,
+    #    'auto-build-setup': (build_script, test_dir, build_suggestion)
+    #}
+    build_worker = BuildWorker()
+    build_worker.build_script = build_script
+    build_worker.build_suggestion = build_suggestion
+    build_worker.executable_files_build = binary_files_build
+    build_worker.build_directory = test_dir
+
+    build_results[test_dir] = build_worker
+
   return build_results
--- a/experimental/c-cpp/build_generator.py
+++ b/experimental/c-cpp/build_generator.py
@@ -39,11 +39,13 @@ class AutoBuildContainer:
 class BuildWorker:
   """Keeper of data on auto generated builds."""
 
-  def __init__(self, build_suggestion: AutoBuildContainer):
+  def __init__(self, build_suggestion: AutoBuildContainer, build_script: str,
+               build_directory: str, executable_files_build: Dict[str,
+                                                                  List[str]]):
     self.build_suggestion: AutoBuildContainer = build_suggestion
-    self.build_script: str = ''
-    self.build_directory: str = ''
-    self.executable_files_build: Dict[str, List[str]] = {}
+    self.build_script: str = build_script
+    self.build_directory: str = build_directory
+    self.executable_files_build: Dict[str, List[str]] = executable_files_build
     self.base_fuzz_build: bool = False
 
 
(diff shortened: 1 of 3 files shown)


experimental/c-cpp/manager.py
```diff
@@ -332,31 +320,30 @@ def get_fuzzer_intrinsics(self, func: Dict[str, Any]) -> Dict[str, Any]:
...
 The most important part of the harness is that it will build and compile correctly against the target code. Please focus on making the code as simple as possible in order to secure it can be build.
 
-%s
-''' % (self.github_url, func['function_signature'], func_source_code,
-       str(headers_to_include), type_constraints, cross_reference_text)
+{cross_reference_text}
+'''
 
-    logger.info('-' * 45, '\n', prompt, '\n', '-' * 45)
+    logger.info('%s%s%s', '-' * 45, '\n' + prompt + '\n', '-' * 45)
```
DonggeLiu, line 346: nit: `f'\n{prompt}\n'`

experimental/c-cpp/manager.py
```diff
@@ -401,35 +388,34 @@ def get_fuzzer_intrinsics(self, func) -> Dict[str, Any]:
...
+The source code for the function is: ```{func_source_code}```
 
 The harness should be in libFuzzer style, with the code wrapped in `int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size)`. Specifically, do not include `extern "C"` in the fuzzer code.
 
 Please wrap all code in <code> tags and you should include nothing else but the code in your reply. Do not include any other text.
 
 Make sure the ensure strings passed to the target are null-terminated.
 
-There is one rule that your harness must satisfy: all of the header files in this library is %s. Make sure to not include any header files not in this list.
+There is one rule that your harness must satisfy: all of the header files in this library is {str(headers_to_include)}. Make sure to not include any header files not in this list.
```
DonggeLiu, line 413: Do we need to cast `headers_to_include` explicitly with `str()`?

experimental/c-cpp/manager.py
```diff
@@ -474,31 +460,30 @@ def get_fuzzer_intrinsics(self, func: Dict[str, Any]) -> Dict[str, Any]:
...
+I would like for you to write the harness targeting the function `{func['function_signature']}.`
 
 The harness should be in libFuzzer style, with the code wrapped in `int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size)`.
 
 Please wrap all code in <code> tags and you should include nothing else but the code in your reply. Do not include any other text.
 
 Make sure the ensure strings passed to the target are null-terminated.
 
-There is one rule that your harness must satisfy: all of the header files in this library is %s. Make sure to not include any header files not in this list.
+There is one rule that your harness must satisfy: all of the header files in this library is {str(headers_to_include)}. Make sure to not include any header files not in this list.
```
DonggeLiu, line 481: Same above
Do we need to cast headers_to_include explicitly with str()?

DonggeLiu: BTW, could this be in a template file in the final version?
Also, could we define repeated lines once and reuse them to avoid repetitions?

experimental/c-cpp/manager.py
```diff
@@ -788,11 +770,11 @@ def build_empty_fuzzers(results, language):
...
   fuzz_compiler, _, empty_fuzzer_file, fuzz_template = get_language_defaults(
       language)
-  for test_dir in results:
+  for test_dir, build_worker in build_workers.items():
     logger.info('Test dir: %s :: %s', test_dir,
-                str(results[test_dir]['refined-static-libs']))
+                str(build_worker.executable_files_build['refined-static-libs']))
 
-    if len(results[test_dir]['refined-static-libs']) == 0:
+    if len(build_worker.executable_files_build['refined-static-libs']) == 0:
```
DonggeLiu, line 777: if build_worker.executable_files_build['refined-static-libs']:

experimental/c-cpp/manager.py
```diff
@@ -900,8 +886,10 @@ def run_introspector_on_dir(build_results, test_dir,
 
 
 def log_fuzzer_source(full_fuzzer_source: str):
-  logger.info('-' * 20, ' HARNESS SOURCE ', '-' * 20, '\n', full_fuzzer_source,
-              '\n', '-' * 56)
+  harness_source_out = '-' * 20 + ' HARNESS SOURCE ' + '-' * 20 + '\n'
+  harness_source_out += full_fuzzer_source + '\n'
+  harness_source_out += '-' * 56
```
DonggeLiu, line 891: Could we do something similar to `f'{"-"*20} PROMPT {"-"*20}\n{prompt}\n{'-' * 48}'`?

experimental/c-cpp/manager.py
```diff
@@ -943,11 +931,12 @@ def generate_harness_intrinsics(
...
-    final_asan_build_script = results[test_dir]['build-script']
+    final_asan_build_script = results[test_dir].build_script + '\n'
     fuzzer_out = '/src/generated-fuzzer'
-    final_asan_build_script += '\n%s %s -o %s' % (
-        ' '.join(fuzzer_build_cmd), fuzzer_intrinsics['build-command-includes'],
-        fuzzer_out)
+    final_asan_build_script += ' '.join(fuzzer_build_cmd) + ' '
+    final_asan_build_script += fuzzer_intrinsics['build-command-includes']
+    final_asan_build_script += ' -o '
+    final_asan_build_script += fuzzer_out
```
DonggeLiu, line 939: how about f-string?

DonggeLiu (approved):

experimental/c-cpp/build_generator.py
```diff
@@ -36,6 +36,17 @@ def __init__(self):
...
 
 
+class BuildWorker:
+  """Keeper of data on auto generated builds."""
+
+  def __init__(self, build_suggestion: AutoBuildContainer):
+    self.build_suggestion: AutoBuildContainer = build_suggestion
+    self.build_script: str = ''
+    self.build_directory: str = ''
+    self.executable_files_build: Dict[str, List[str]] = {}
```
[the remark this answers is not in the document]
DavidKorczynski: We can't -- this code runs in the oss-fuzz base images which relies on Python 3.8 and the typing is a Python3.9+ thing (see initial PR comment)

experimental/c-cpp/manager.py
DavidKorczynski: > Same above Do we need to cast headers_to_include explicitly with str()?

We don't! thanks, fixed!

DavidKorczynski: > BTW, could this be in a template file in the final version? Also, could we define repeated lines once and reuse them to avoid repetitions?

Yep, I'll look at doing this when merging the two https://github.com/google/oss-fuzz-gen/issues/450 -- for the moment it's nice to have this in a single module as it makes for easy copying in/out of the containers as some development of this module happens within base-builder container environments. I will address this later though.

experimental/c-cpp/build_generator.py
```diff
@@ -707,8 +718,8 @@ def extract_build_suggestions(
 
 
 def raw_build_evaluation(
-    all_build_scripts: List[Tuple[str, str, AutoBuildContainer]],
-    initial_executable_files: Dict[str, List[str]]) -> Dict[str, Any]:
+    all_build_scripts: List[Tuple[str, str, AutoBuildContainer]]
+) -> Dict[str, BuildWorker]:
```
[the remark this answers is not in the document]
DonggeLiu: oh I see. thanks!

PR #456, opened 2024-07-08
core: enable use of dedicated oss-fuzz folder

DavidKorczynski:
This is useful during development purposes where it's convenient being
able to adjust the OSS-Fuzz projects' folders, e.g. to test hypotheses.

PR #459, opened 2024-07-09
experiment: fix readability of conditional

DavidKorczynski:
Ref: https://github.com/google/oss-fuzz-gen/pull/456/files#r1669405244


DonggeLiu (approved):
Thanks!

PR #466, opened 2024-07-11
Filter empty string in comment

DonggeLiu:
The comment parser sometimes include empty string as a command line argument, which breaks python's argparser in [`request_pr_exp.py`](https://github.com/google/oss-fuzz-gen/blob/main/ci/request_pr_exp.py).

This PR removes empty strings from the command parsed from the comment.


DonggeLiu:
Report looks good:
https://llm-exp.oss-fuzz.com/Result-reports/ofg-pr/2024-07-11-466-dg-comparison/index.html

DonggeLiu:
TBR.

oliverchang (commented):
LGTM

PR #469, opened 2024-07-12
Separate osv-scanner workflows so that the action for scheduled and push will not appear on PRs

DonggeLiu:
Avoid showing irrelevant action on PRs: 
image


DonggeLiu:
Once approved, I will add both workflows to https://github.com/google/oss-fuzz-gen/security/code-scanning/tools/osv-scanner/status

oliverchang (approved):

PR #479, opened 2024-07-14
Add set up for corpus generator generation by way of LLMs

DavidKorczynski:
Adds logic for adding a corpus to a given harness. This is based off of using LLMs to ask for it to generate a python script that generates a set of seeds for a given target. This program is then appended to the build script of the project also wrapped such that zip files are created for OSS-Fuzz-style seed corpus.

PR #480, opened 2024-07-14
Rename gen code to query llm

DavidKorczynski:
The main reason for this is that the `generate_code` does not really generate any code but rather it queries a given LLM using a specified prompt. Since we now have prompts of various sort, I feel it might be a bit misplaced the name. This could also make it a bit more clear which API to use if you're working on using LLMs for tasks other than explicit code generation.

This came up while doing https://github.com/google/oss-fuzz-gen/pull/479 where one consideration was to have the LLM generate a corpus explicit without going a seed-corpus-by-way-of-python generation.

Ref: https://github.com/google/oss-fuzz-gen/issues/482

### DIFF
--- a/experimental/manual/prompter.py
+++ b/experimental/manual/prompter.py
@@ -72,4 +72,4 @@ if __name__ == "__main__":
   model = setup_model()
   prompt = construct_prompt()
   os.makedirs(args.response_dir, exist_ok=True)
-  model.generate_code(prompt, response_dir=args.response_dir)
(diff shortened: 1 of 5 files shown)


DavidKorczynski:
/gcbrun exp -n dk-test-infra-99 -m vertex_ai_gemini-1-5 -b minor-for-ci -i

DavidKorczynski:
Experiment is looking good.

mihaimaruseac (approved):

DonggeLiu (commented):
LGTM

PR #487, opened 2024-07-15
unify logging

DavidKorczynski:
We use logging in multiple formats:

- logging.info/debug/..
- print(
- write to files

It's fairly inconsistent. So am unifying it in a more Pythonic way. The format also makes it easier to track performance of the tool as we can find bottlenecks by looking at the logs.

### DIFF
--- a/data_prep/introspector.py
+++ b/data_prep/introspector.py
@@ -874,7 +874,7 @@ def _parse_arguments() -> argparse.Namespace:
 set_introspector_endpoints(DEFAULT_INTROSPECTOR_ENDPOINT)
 
 if __name__ == '__main__':
-  logging.basicConfig(level=logger.INFO)
+  logging.basicConfig(level=logging.INFO)
 
   args = _parse_arguments()
   if args.out:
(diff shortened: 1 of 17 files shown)


DavidKorczynski:
/gcbrun exp -n dk-test-infra-99x3 -m vertex_ai_gemini-1-5 -b comparison

DonggeLiu:
Thanks for unifying the logs, @DavidKorczynski. In addition to the benefit of the consistent format (as you mentioned), this will keep the logs in the correct order, which will help us debug.

However, since you happen to work on this, could you please also help us switch to [`google-cloud-logging`](https://pypi.org/project/google-cloud-logging/)?
The current code prints all logs in stderr, which shows all logs are errors:
image

Using `google-cloud-logging` will fix this, here is [how it is used in FuzzBench](https://github.com/google/fuzzbench/blob/8bb3c88cc6a9e9e2bc43f5e452e94a38a972ff2a/common/logs.py#L23).

DavidKorczynski:
> However, since you happen to work on this, could you please also help us switch to [google-cloud-logging](https://pypi.org/project/google-cloud-logging/)?
> The current code prints all logs in stderr, which shows all logs are errors:

Am happy to do this, can we do this in follow-up PRs?

DonggeLiu (approved):
Yep sure, the sooner the better as we won't be able to filter logs by severity once we merge this.

PR #490, opened 2024-07-16
JVM: Fix multiple bugs in JVM coverage calculation

arthurscchan:
This PR fixes multiple bugs that cause problems in calculating line coverage diff for the JVM projects.
The bugs are listed below.
1. The path to finding the existing coverage report (jacoco.xml) is wrong, making the logic always compare with an empty report when calculating for line coverage diff. This PR adds a new functions to retrieve and parse the existing coverage report correctly.
2. The interpretation logic for the JVM coverage report on array-type arguments or empty arguments is wrong, resulting in a non-matching method name when merging the coverage report. This PR fixes the java name and argument interpretation to get a correct method signature as identifiers when combining and subtracting coverage reports.
3. The subtraction of that line diff values between the existing coverage report and the new coverage report is reversed, making the calculation wrong. This PR fixes that.
4. The pre-check process of builder_runner.py will check if the target method exists in the generated harness before building it. The checking is done by the _contains_target_jvm_method() function. In Java source code, a single method call could contain an unlimited amount of space and could also split into multiple lines. The current checking logic fails to discover if spaces or line feeds exist within a method invocation. This PR fixes that by pre-processing the code to remove line feed and space before processing to correctly discover if the target method is being called or not.
5. Since constructors of classes are some of our targets and there are too many ways to invoke them, the current logic in _contains_target_jvm_method() fails to discover existing constructor invocation and causes false negative cases. This PR skip the pre-check for the constructor target of the JVM project to avoid those false negative cases.
6. JVM coverage information is mapped towards random line name but not the correct line number information. This PR fixes that by modifying the logic of the from_jvm_file method to correctly interpret the jacoco.xml report.
7. Subtraction of coverage report only done by comparing the line coverage count from the coverage reports without checking if different lines has been covered and result in same covered line count. This PR fixes the logic of those functions to calculate the coverage gain correctly.

### DIFF
--- a/experiment/evaluator.py
+++ b/experiment/evaluator.py
@@ -93,7 +93,7 @@ def load_existing_textcov(project: str) -> textcov.Textcov:
     if not blob.name.endswith('.covreport'):
       continue
 
-   logger.info(f'Loading existing textcov from {blob.name}')
+    logger.info(f'Loading existing textcov from {blob.name}')
     with blob.open('rb') as f:
       existing_textcov.merge(textcov.Textcov.from_file(f))
 
--- a/experiment/evaluator.py
+++ b/experiment/evaluator.py
@@ -65,8 +65,7 @@ class Result:
     return dataclasses.asdict(self)
 
 
-def load_existing_textcov(project: str,
-                          language: str = 'c++') -> textcov.Textcov:
+def load_existing_textcov(project: str) -> textcov.Textcov:
   """Loads existing textcovs."""
   storage_client = storage.Client.create_anonymous_client()
   bucket = storage_client.bucket(OSS_FUZZ_COVERAGE_BUCKET)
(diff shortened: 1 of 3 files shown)


experiment/textcov.py
```diff
@@ -257,12 +271,24 @@ def from_jvm_file(cls, file_handle) -> Textcov:
...
             if method_item.attrib['name'] not in JVM_SKIPPED_METHOD:
-              class_method_items.append((class_name, method_item))
+              class_method_items.append((class_name, method_item, coverage))
 
-    for class_name, method_item in class_method_items:
+    for class_name, method_item, coverage in class_method_items:
       method_dict = method_item.attrib
       method_name = method_dict['name']
 
+      # Deteremine start index in coverage list
```
DavidKorczynski, line 280: typo -- `Determine`

experiment/evaluator.py
```diff
@@ -91,22 +90,40 @@ def load_existing_textcov(project: str,
...
-      if not blob.name.endswith('.covreport'):
-        continue
+    if not blob.name.endswith('.covreport'):
+      continue
 
-      logger.info(f'Loading existing textcov from {blob.name}')
-      with blob.open('rb') as f:
-        existing_textcov.merge(textcov.Textcov.from_file(f))
+   logger.info(f'Loading existing textcov from {blob.name}')
+    with blob.open('rb') as f:
```
DavidKorczynski, line 97: there looks to be some indentation issues here?

experiment/textcov.py
```diff
@@ -242,13 +238,31 @@ def from_jvm_file(cls, file_handle) -> Textcov:
...
+
+    # Process methods
     class_method_items = []
     for item in jacoco_report.iter():
       if item.tag == 'class':
         # Skip fuzzer classes
         if textcov.is_fuzzer_class(item):
           continue
 
+        # Get line covgerage information for this class
```
DavidKorczynski, line 263: typo

experiment/textcov.py
```diff
@@ -257,12 +271,24 @@ def from_jvm_file(cls, file_handle) -> Textcov:
...
 
+      # Deteremine start index in coverage list
+      start_line = int(method_dict['line'])
+      start_index = -1
+      for count, item in enumerate(coverage):
+        if item[0] == start_line:
+          start_index = count
+          break
+
+      # Coverage information failed to retrieve, skipping this method
```
DavidKorczynski, line 288: ` # Failed to retrieve coverage information, ...`

experiment/textcov.py
```diff
@@ -257,12 +271,24 @@ def from_jvm_file(cls, file_handle) -> Textcov:
...
-              class_method_items.append((class_name, method_item))
+              class_method_items.append((class_name, method_item, coverage))
 
-    for class_name, method_item in class_method_items:
+    for class_name, method_item, coverage in class_method_items:
       method_dict = method_item.attrib
       method_name = method_dict['name']
 
+      # Deteremine start index in coverage list
+      start_line = int(method_dict['line'])
```
DavidKorczynski, line 281: Instead of type casting here can we do it when we store the line number?

DavidKorczynski (changes requested):

experiment/textcov.py
arthurscchan: The number is not stored before, the method_dict is lively retrieved on line 277 from the XML ET.

DavidKorczynski: Ah sorry, for some reason I thought we assembled this dict ourselves.

DavidKorczynski (approved):

Issue #494, opened 2024-07-17
Logic for test-to-harness conversion


DavidKorczynski:
Test-to-harness conversion by way of LLM sounds like an interesting avenue and is very commonly an approach taken by security engineers when first approaching a given project.

I see multiple steps:

1) Enable in experimental without use of FI
2) Enable in core without use of FI
3) Assess quality overall
4) See if improvements can be made using more program analysis data by way of FI

DavidKorczynski:
An example OSS-Fuzz project that has low coverage (5% at the time of writing) but a wealth of tests that can be converted: https://github.com/neomutt/neomutt/tree/main/test

DavidKorczynski:
Option (1) above has been implemented in https://github.com/google/oss-fuzz-gen/pull/495

PR #495, opened 2024-07-17
experimental: add test-to-harness conversion logic

DavidKorczynski:
Adds a fuzz harness heuristic that relies on converting existing tests. At this stage, it's done without relying on FI, we simply (1) find tests files in the target project; (2) read them; (3) for each test file we use a simple prompt to convert it into a harness.

At this stage, it already out-performs on some existing projects, e.g: https://github.com/jkuhlmann/cgltf/blob/master/test/main.c

In this case, we have a harness generated that looks quite nice:

```c
// Heuristic: TestConverterPrompt :: Target: 
#include <stdlib.h>
#include <stdint.h>
#include <stdio.h>
#include <string.h>

#define CGLTF_IMPLEMENTATION
#include "cgltf.h"

extern "C" int LLVMFuzzerTestOneInput(const uint8_t* data, size_t size) {
    if (size < 1) {
        return 0;
    }

    cgltf_options options;
	memset(&options, 0, sizeof(cgltf_options));
	cgltf_data* parsed_data = NULL;
	cgltf_result result;

    // Parse input data
    result = cgltf_parse(&options, data, size, &parsed_data);

    if (result == cgltf_result_success) {
        result = cgltf_validate(parsed_data);
    }

    if (result == cgltf_result_success) {
        // Use the parsed data in some way
        // For example, print file type and mesh count
		printf("Type: %u\n", parsed_data->file_type);
		printf("Meshes: %u\n", (unsigned)parsed_data->meshes_count);
    }

    cgltf_free(parsed_data);

    return 0;
}
```

Ref: https://github.com/google/oss-fuzz-gen/issues/494


experimental/c-cpp/manager.py
```diff
@@ -398,6 +413,79 @@ def get_fuzzer_intrinsics(self, func: Dict[str, Any]) -> Dict[str, Any]:
...
+               all_header_files: List[str], test_dir: str):
+    super().__init__(introspector_report, all_header_files, test_dir)
+    self.introspector_report = introspector_report
+    self.all_header_files = all_header_files
+    self.github_url = ''
+
+  def get_fuzzing_targets(self) -> List[Any]:
+    return []
+
+  def get_fuzzer_test_intrinsics(self, test_case: Test) -> Dict[str, Any]:
```
DonggeLiu, line 431: nit: function doc-string?

experimental/c-cpp/manager.py
```diff
@@ -398,6 +413,79 @@ def get_fuzzer_intrinsics(self, func: Dict[str, Any]) -> Dict[str, Any]:
...
+    (headers_to_include, _,
+     build_command_includes) = self.get_header_intrinsics()
+
+    # Include any weird macros defined that does not have any values. This
+    # was found empirically to be valuable.
+    macros_defined_in_test = []
+    for line in test_case.test_content.split('\n'):
+      if '#define' in line and len(line.split(' ')) == 2:
```
DonggeLiu, line 439: Nit: Maybe exclude commented lines later?
```c
// #define a b
/*
#define a b
*/
```

experimental/c-cpp/manager.py
```diff
@@ -923,6 +1011,37 @@ def log_fuzzer_source(full_fuzzer_source: str):
...
+    fuzz_converter = FuzzerGenHeuristicTestConverter({}, all_header_files,
+                                                     test_dir)
+    fuzz_converter.github_url = github_url
+    fuzzer_intrinsics = fuzz_converter.get_fuzzer_test_intrinsics(test)
+
+    # Generate a build script for compiling the fuzzer with ASAN.
+    final_asan_build_script = build_results[test_dir].build_script + '\n'
+    fuzzer_out = '/src/generated-fuzzer'
+    fuzz_cmd = ' '.join(fuzzer_build_cmd) + ' '
+    fuzz_includes = fuzzer_intrinsics["build-command-includes"]
```
DonggeLiu, line 1033: nit: `'build-command-includes'` for consistency.

experimental/c-cpp/manager.py
```diff
@@ -923,6 +1011,37 @@ def log_fuzzer_source(full_fuzzer_source: str):
...
+
+    fuzz_converter = FuzzerGenHeuristicTestConverter({}, all_header_files,
+                                                     test_dir)
+    fuzz_converter.github_url = github_url
+    fuzzer_intrinsics = fuzz_converter.get_fuzzer_test_intrinsics(test)
+
+    # Generate a build script for compiling the fuzzer with ASAN.
+    final_asan_build_script = build_results[test_dir].build_script + '\n'
+    fuzzer_out = '/src/generated-fuzzer'
+    fuzz_cmd = ' '.join(fuzzer_build_cmd) + ' '
```
DonggeLiu, line 1032: May I ask why the trailing space is necessary, given `final_asan_build_script += f'{fuzz_cmd} {fuzz_includes} ...'`?

experimental/c-cpp/manager.py
```diff
@@ -1256,6 +1375,25 @@ def get_heuristics_to_use() -> List[Type[FuzzHeuristicGeneratorBase]]:
...
+  """Returns a list of the test files in the target source path."""
+  all_files = get_all_files_in_path(target_source_path)
+  all_tests = []
+  for file in all_files:
+    split_path = file.split('/')
+    is_test = any(['test' in path for path in split_path])
+    if not is_test:
+      continue
+    if not file.endswith('.c') and not file.endswith('.cpp'):
+      continue
```
DonggeLiu, line 1388: Would this miss files like `.cc` or `.cxx`?
Maybe using [`is_c_file(file)`](https://github.com/google/oss-fuzz-gen/blob/4e3798ef7a6af994d7e09094cd399876e701d1a2/experiment/benchmark.py#L225) and [`is_cpp_file(file)`](https://github.com/google/oss-fuzz-gen/blob/4e3798ef7a6af994d7e09094cd399876e701d1a2/experiment/benchmark.py#L230)?

DonggeLiu (approved):
Thanks! Some nits:

experimental/c-cpp/manager.py
```diff
@@ -398,6 +413,79 @@ def get_fuzzer_intrinsics(self, func: Dict[str, Any]) -> Dict[str, Any]:
     return fuzzer_intrinsics
 
 
+class FuzzerGenHeuristicTestConverter(FuzzHeuristicGeneratorBase):
+  """Heuristic that provides context around target function."""
+  language = 'c'
+  name = 'TestConverterPrompt'
+
+  def __init__(self, introspector_report: Dict[str, Any],
+               all_header_files: List[str], test_dir: str):
```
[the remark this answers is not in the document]
DavidKorczynski: Can't do, see here: https://github.com/google/oss-fuzz-gen/pull/454#discussion_r1670028212

experimental/c-cpp/manager.py
DavidKorczynski: good catch, there's no need.

experimental/c-cpp/manager.py
DavidKorczynski: Used the array! We're not copying the core OFG in so we can't used the module directly (would also cause issues with the Python version regarding the typing)

experimental/c-cpp/manager.py
DavidKorczynski: I'll leave this for now -- it works well but ultimately I would want some stronger logic for macros (e.g. using IR/AST stuff). Will leave as is and monitor if it shows up as a limitation.

experimental/c-cpp/manager.py
DonggeLiu: Ah, the inconsistency in the Python version is annoying. But I guess we have to bear with it as there are more important changes.
The array should suffice.

PR #496, opened 2024-07-17
Simplify the nested condition in textcov

arthurscchan:
The conditional check of array_count larger than 0 is not necessary since it is only used in repeating '[]' in the next_arg and the repeat of string won't happen if the array_count is less than or equals 0. This PR removes the unnecessary conditional check to simplify the logic and decrease the amount of nested conditional checking.


oliverchang (approved):

experiment/textcov.py
```diff
@@ -396,9 +396,8 @@ class name specification use single upper case letter for
         if c == 'L':
           start = True
           if next_arg:
-            if array_count > 0:
-              next_arg = f'{next_arg}{"[]" * array_count}'
-              array_count = 0
+            next_arg = f'{next_arg}{"[]" * array_count}'
```
DavidKorczynski, line 399: Could we change this to `next_arg += f'{"[]" * array_count}'` -- isn't it a bit easier to read?

DavidKorczynski (changes requested):

DavidKorczynski (approved):

PR #497, opened 2024-07-17
experimental: delete docker images after runs

DavidKorczynski:
otherwise the storage is getting eaten too fast

### DIFF
--- a/experimental/c-cpp/runner.py
+++ b/experimental/c-cpp/runner.py
@@ -251,6 +251,13 @@ def run_on_targets(target,
               build_heuristics=build_heuristics,
               generator_heuristics=generator_heuristics)
 
+  # Cleanup the OSS-Fuzz docker image
+  clean_up_cmd = ['docker', 'image', 'rm', f'gcr.io/oss-fuzz/{worker_project_name}']
+  try:
+    subprocess.check_call(' '.join(clean_up_cmd), shell=True)
+  except subprocess.CalledProcessError:
+      pass
+
   if semaphore is not None:
     semaphore.release()
 
--- a/experimental/c-cpp/runner.py
+++ b/experimental/c-cpp/runner.py
@@ -252,11 +252,13 @@ def run_on_targets(target,
               generator_heuristics=generator_heuristics)
 
   # Cleanup the OSS-Fuzz docker image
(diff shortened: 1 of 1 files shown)


experimental/c-cpp/runner.py
```diff
@@ -251,6 +251,15 @@ def run_on_targets(target,
               build_heuristics=build_heuristics,
               generator_heuristics=generator_heuristics)
 
+  # Cleanup the OSS-Fuzz docker image
+  clean_up_cmd = [
+      'docker', 'image', 'rm', f'gcr.io/oss-fuzz/{worker_project_name}'
+  ]
+  try:
+    subprocess.check_call(' '.join(clean_up_cmd), shell=True)
```
DonggeLiu, line 259: No need to join, IIRC?

DonggeLiu (approved):

experimental/c-cpp/runner.py
DavidKorczynski: For `check_call` here it's needed -- a list doesn't do the job. That said, I'm not 100% sure about when subprocess wants a list/string in general.

experimental/c-cpp/runner.py
```diff
@@ -251,6 +251,15 @@ def run_on_targets(target,
               build_heuristics=build_heuristics,
               generator_heuristics=generator_heuristics)
 
+  # Cleanup the OSS-Fuzz docker image
+  clean_up_cmd = [
+      'docker', 'image', 'rm', f'gcr.io/oss-fuzz/{worker_project_name}'
```
oliverchang, line 256: hmm, I thought we were always running `docker run` with `--rm` which should do this cleanup automatically on exit. Is that not the case?

DavidKorczynski: We build an oss-fuzz project per project that is being analysed. It's the image build from the following that we need to clean up: https://github.com/google/oss-fuzz-gen/blob/4e3798ef7a6af994d7e09094cd399876e701d1a2/experimental/c-cpp/runner.py#L67-L72

i.e. after the execution there will be in the shared folders:
```
build/out/temp-project-0
build/out/temp-project-1
...
```
each corresponding to auto-generation of a given repository

DonggeLiu: I thought `docker image rm` removes the image while `docker run --rm` removes the container?

PR #502, opened 2024-07-24
Add support for Claude models.

inferno-chromium:


llm_toolkit/models.py
```diff
@@ -263,6 +264,67 @@ class GPT4o(GPT):
   name = 'gpt-4o'
 
 
+class Claude(LLM):
+  """Anthropic's Claude model encapsulator."""
+
+  _max_output_tokens = 4096
+  context_window = 200000
+
+  name = "claude-3-haiku@20240307"
```
oliverchang, line 273: nit: should we hide the version number from the actual users of oss-fuzz-gen? 

This is how we handle Gemini for example: https://github.com/google/oss-fuzz-gen/blob/bdcc09b3ed2c8b85929ce262009847b436a0ba5f/llm_toolkit/models.py#L461

We have a public "vertex_ai_gemini-1-5" name that can be passed to the CLI, but under the hood "gemini-1.5-pro-001" is what gets passed to the Vertex AI API.

oliverchang (commented):
thanks! just one minor comment

llm_toolkit/models.py
```diff
@@ -263,6 +264,67 @@ class GPT4o(GPT):
...
+    return prompts.ClaudePrompt
+
+  # ============================== Generation ============================== #
+  def query_llm(self,
+                prompt: prompts.Prompt,
+                response_dir: str,
+                log_output: bool = False) -> None:
+    """Queries Claude's API and stores response in |response_dir|."""
+    if self.ai_binary:
+      logger.info(f'Claude does not use local AI binary: {self.ai_binary}')
```
oliverchang, line 292: nit: can we just exception out here instead?

llm_toolkit/models.py
inferno-chromium: Done

llm_toolkit/models.py
inferno-chromium: Done

inferno-chromium:
/gcbrun exp -n claude-support-fixes-2 -m vertex_ai_claude-3-5-sonnet

oliverchang (approved):

PR #506, opened 2024-07-24
JVM: Fix jvm-small benchmark

arthurscchan:
This PR changes the jvm-small benchmarks to include better method targets with better coverage gain in JVM projects.

PR #507, opened 2024-07-25
JVM: fix benchmark for jvm-medium

arthurscchan:
This PR relates to #506 and continue to modify, add and remove the benchmark yaml files in the jvm-medium benchmark set for JVM projects. The end goal is to have better coverage gain result shown by the benchmark.


arthurscchan:
It is found that the LLM is sometime strange for JVM projects. I have execute the benchmark for multiple times, and sometimes there are benchmark that does not generate a fuzzing harness without any coverage gain. But sometimes LLM does returns quite some good fuzzing harnesses that does increase coverage. Thus the next step is to consider how the prompt could be fixed to make the result more "stable.

arthurscchan:
There will be 1 more PR in the future to modify the benchmarks in the jvm-full directory later.

arthurscchan:
/gcbrun exp -n ascc-jvm-pr-507-1 -m vertex_ai_gemini-1-5 -b jvm-medium

arthurscchan:
/gcbrun exp -n ascc-jvm-pr-507-2 -m vertex_ai_gemini-1-5 -b jvm-medium

DavidKorczynski (approved):

PR #512, opened 2024-07-26
JVM: Fix introspector querying logic

arthurscchan:
It is found that the query of the introspector oracles are customised for C/C++ projects. This PR adds an additional function to process the query of the introspector oracles for JVM projects which works in a different manner.

### DIFF
--- a/data_prep/introspector.py
+++ b/data_prep/introspector.py
@@ -613,8 +613,34 @@ def _combine_functions(a: list[str], b: list[str], c: list[str],
   return list(combined)
 
 
+def _select_functions_from_jvm_oracles(project: str, limit: int,
+                                       target_oracles: list[str]) -> list[dict]:
+  """Selects functions from oracles designated for jvm projects, with
+  jvm-public-candidates as the prioritised oracle"""
+  all_functions = OrderedDict()
+
+  if 'jvm-public-candidates' in target_oracles:
+    # JPC is the primary oracle for JVM projects. If it does exist, all other
+    # oracles are ignored because the results from all other oracles are subsets
+    # of the results from JPC oracle for JVM projects.
+    jpc_targets = _select_top_functions_from_oracle(project, limit,
+                                                    'jvm-public-candidates',
+                                                    target_oracles)
+    all_functions.update(jpc_targets)
+  else:
+    # JPC does not exist in target_oracles, follow given oracle order and
+    # dedulicate instead.
+    for target_oracle in target_oracles:
+      tmp_functions = _select_top_functions_from_oracle(project, limit,
+                                                        target_oracle,
+                                                        target_oracles)
+      all_functions.update(tmp_functions)
+
+  return list(all_functions.values())[:limit]
+
+
 def _select_functions_from_oracles(project: str, limit: int,
-                                   target_oracles: List[str]) -> list[dict]:
+                                   target_oracles: list[str]) -> list[dict]:
   """Selects function-under-test from oracles."""
   all_functions = OrderedDict()
   frlc_targets = _select_top_functions_from_oracle(project, limit,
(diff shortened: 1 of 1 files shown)


arthurscchan:
/gcbrun exp -n ascc-jvm-pr-512-1 -m vertex_ai_gemini-1-5 -b jvm-small

arthurscchan:
/gcbrun exp -n ascc-jvm-pr-512-1 -m vertex_ai_gemini-1-5 -b jvm-small

DonggeLiu:
Thanks for looking into this @arthurscchan.
Previously, `_combine_functions()` has both `b` and `c` because I wanted to add at least one benchmark from each set so that the final set has at least one from each oracle.

I implemented the previous logic specifically for C/C++ benchmarks because they only have 3 oracles: `far-reach-but-low-coverage`, `easy-params-far-reach`, and `optimal-targets`. Please let me know if I was wrong.

Sorry that I did not realize we needed to update the Java benchmark frequently. Given Java needs a different orcale set, would it be easier to add a new layer to 1) check which language it is, and 2) use a different logic for Java?

arthurscchan:
I am OK with both approach, I could separate the logic for jvm on oracle querying without touching the original c/c++ logic, it maybe easier to maintain.

DavidKorczynski (approved):

data_prep/introspector.py
```diff
@@ -613,8 +613,34 @@ def _combine_functions(a: list[str], b: list[str], c: list[str],
...
+  else:
+    # JPC does not exist in target_oracles, follow given oracle order and
+    # dedulicate instead.
+    for target_oracle in target_oracles:
+      tmp_functions = _select_top_functions_from_oracle(project, limit,
+                                                        target_oracle,
+                                                        target_oracles)
+      all_functions.update(tmp_functions)
+
+  return list(all_functions.values())[:limit]
```
DonggeLiu, line 639: nit, avoid repeated code:
```python
if 'jvm-public-candidates' in target_oracles:
  target_oracles = ['jvm-public-candidates']
for target_oracle in target_oracles:
      tmp_functions = _select_top_functions_from_oracle(project, limit,
                                                        target_oracle,
                                                        target_oracles)
      all_functions.update(tmp_functions)
  return list(all_functions.values())[:limit]
```
A cleaner way is to enforce this when [parsing args](https://github.com/google/oss-fuzz-gen/blob/9c4780d8c743beb86b19b578a3703a32bddc8623/data_prep/introspector.py#L909). E.g.:
```python
args = parser.parse_args()
if 'jvm-public-candidates' in args.target_oracles:
  args.target_oracles = ['jvm-public-candidates']
```
or even
```python
args = parser.parse_args()
assert ('jvm-public-candidates' not in args.target_oracles or args.target_oracles == ['jvm-public-candidates'], 'JPC is the primary oracle for JVM projects...'
```

DonggeLiu (commented):
Thanks for fixing the logic, a nit comment below.

data_prep/introspector.py
arthurscchan: I though the first way is better, cause it allows better flexibility when we don't want jvm-public-candidates in some situation and when we want to test some other oracles solely.

DonggeLiu: Please correct me if I am wrong, but I thought all four snippets (yours + my three in the comment) were functionally equivalent: They all remove other oracles if and only if `'jvm-public-candidates'` is in `target_oracles`.

The main difference is _where_ to implement this logic. I suggested the last one (`assert`) so that our users will be reminded that other oracles will not be used if `jvm-public-candidates` exists. Similarly, the second last way intends to prepare the **correct** oracle list as early as possible so that the **wrong** list won't confuse us/users.

If `jvm-public-candidates` is not in the oracle list, then all four snippets should behave the same.

PR #515, opened 2024-07-29
data_prep: Fix bug for jvm argument handling

arthurscchan:
It is found that there is a temporary function in data_prep/introspector.py for fixing a wrong function name with struct for c projects. But this function accidentally affected Java projects because the function assumed that all function names wouldn't have a dot(.) character. This assumption is wrong for Java projects since most of the Java classes will have a dot(.) character in the fully qualified class name to identify its package. This PR fixes the problem by skipping the process if it is a Java project. This could be removed if the temporary function is removed in later PRs.


DavidKorczynski:
@DonggeLiu meant to ask for review on https://github.com/google/oss-fuzz-gen/pull/512

DavidKorczynski (approved):

DonggeLiu (commented):
LGTM

DonggeLiu:
> @DonggeLiu meant to ask for review on #512

Done!

PR #530, opened 2024-08-08
Always delete old bucket dir and GKE job of the same name in PR experiments

DonggeLiu:
To avoid repeatedly manually adding `-f` in `/gcb exp` commands.

### DIFF
--- a/ci/request_pr_exp.py
+++ b/ci/request_pr_exp.py
@@ -111,11 +111,6 @@ def _parse_args(cmd) -> argparse.Namespace:
       help=('Delay each benchmark experiment by N seconds, default: '
             f'{EXP_DELAY}.'))
   parser.add_argument(
-      '-f',
-      '--force',
-      action='store_true',
-      help='Remove existing GKE job and bucket before creating new ones.')
-  parser.add_argument(
       '-to',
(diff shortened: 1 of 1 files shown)


ci/request_pr_exp.py
```diff
@@ -110,11 +110,6 @@ def _parse_args(cmd) -> argparse.Namespace:
       default=EXP_DELAY,
       help=('Delay each benchmark experiment by N seconds, default: '
             f'{EXP_DELAY}.'))
-  parser.add_argument(
```
oliverchang, line 113: will this break invocations for people passing `-f` out of habit?

oliverchang (approved):

ci/request_pr_exp.py
DonggeLiu: Yep, it would. Let me notify everyone.
@DavidKorczynski @arthurscchan: We no longer need to add `-f` to `gcbrun` commands, old bucket/job will be deleted automatically when the new one is requested.

DonggeLiu:
Let me double-check that this can relaunch a new exp under the same name.

DonggeLiu:
Yep the latter one launched successfully.
I will merge.

PR #528, opened 2024-08-06
JVM: Add resources closing suggestion to JVM prompts

arthurscchan:
This PR fixes the prompt generation of JVM projects. It includes new requirements to close resource classes to reduce the chance of OOM in generated fuzzers before too many opened resource objects.

PR #531, opened 2024-08-08
JVM: Revamp JVM specific properties retrieval for prompt generation

arthurscchan:
There are some properties in the benchmarks that only exist for JVM projects. In the current logic, these properties are being added to the benchmark class and only JVM projects will use them. This PR reduces the size of benchmarks by removing these properties from the benchmarks. Instead, these properties are lively queries from the introspector new API without the need to store them in the benchmark objects. 
In addition to the two properties (exceptions and is_jvm_static) for JVM projects, a third property is also added to indicate if the target method is a resource-type object (implements AutoClosable) that needs to be closed after use. This helps to reduce possible OOM in generated fuzzing harnesses because of memory leaks from unclosed resource objects. This is the replacement of the work originally shown in #528.

### DIFF
--- a/data_prep/introspector.py
+++ b/data_prep/introspector.py
@@ -297,15 +297,17 @@ def query_introspector_function_line(project: str, func_sig: str) -> list:
   return [_get_data(resp, 'src_begin', 0), _get_data(resp, 'src_end', 0)]
 
 
-def query_introspector_function_props(project: str, func_sig: str) -> tuple:
+def query_introspector_function_props(project: str, func_sig: str) -> dict:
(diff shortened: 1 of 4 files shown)


arthurscchan:
/gcbrun exp -f -n ascc-jvm-pr-531-2 -m vertex_ai_gemini-1-5 -b comparison

arthurscchan:
/gcbrun exp -n ascc-jvm-pr-531-5 -m vertex_ai_gemini-1-5 -i -b jvm-medium

DonggeLiu (approved):

DavidKorczynski (approved):

PR #532, opened 2024-08-08
Benchmark: removed unnecessary keys in benchmark

arthurscchan:
Following #531, some JVM-specific properties have been removed from the Benchmark object. Thus these properties are no longer needed in all those project benchmark YAML files. This PR removes those properties from all the existing benchmark YAML files.


arthurscchan:
/gcbrun exp -n ascc-jvm-pr-532-1 -m vertex_ai_gemini-1-5 -i -b jvm-all

PR #534, opened 2024-08-12
[DO NOT MERGE] Agent-based fuzz targert generation

DonggeLiu:
A preliminary script to give LLM complete bash access to all project files and build environment to assist code generation and fixing. All structural/design suggestions are welcome. Please ignore the bad coding style for now, I will fix them later. LLM performance is not stable yet, I will fix the prompts too.

```bash
# Generation
python -m experimental.agent.gen_agent ada-url "struct url ada::url_aggregator ada::parser::parse_url<ada::parse_url<ada::url>(string_view, const struct url *)"

# Fix
python -m experimental.agent.fix_agent aspell aspell.cc aspell.err
```

`gen_agent` Overall steps:
1. Create and keep a docker container to build the fuzz target,
2. Create and keep an LLM chat session.
3. LLM iteratively tells us the command to run to investigate project files
4. We run the command in the container and send LLM the output
5. Repeat 3 and 4 until LLM is ready and send us the fuzz target.

`fix_agent` follows the similar design to fix build errors.

To scale this via cloud experiments, I plan to do this inside a Cloud Build for each benchmark trial.
For LLM-based crash triaging, we will connect LLM to lldb.

### DIFF
--- a/experimental/agent/gen_agent.py
+++ b/experimental/agent/gen_agent.py
@@ -1,0 +1,164 @@
+"Use Agent to chat with LLM and fix build errors."
+import os
+import re
+import subprocess
+import sys
+
+import vertexai
+import yaml
+from vertexai import generative_models
+from vertexai.generative_models import GenerativeModel
+
+PROJECT = sys.argv[1]
+FUNCTION_UNDER_TEST = sys.argv[2]
+
+IMAGE_NAME = f'gcr.io/oss-fuzz/{PROJECT}'
+with open(f'benchmark-sets/all/{PROJECT}.yaml', 'r') as benchmark_file:
+  data = yaml.safe_load(benchmark_file)
+  FUZZ_TARGET_PATH = data['target_path'].strip()
+  FUZZ_TARGET_BINARY = data['target_name'].strip()
+
+# with open(sys.argv[2]) as fuzz_target_file:
+#   CODE_TO_BE_FIXED = fuzz_target_file.read()
+
+# with open(sys.argv[3]) as error_message_file:
+#   ERROR_MESSAGES = error_message_file.read()
+
+# if len(sys.argv) > 4:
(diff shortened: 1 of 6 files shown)


DonggeLiu:
We will start thinking about refactoring OFG to adapt to this at some point.
But we can start small, e.g., only add this as an optional generation/fixing method.

experimental/agent/gen_agent.py
```diff
@@ -0,0 +1,164 @@
...
+  try:
+    while cur_round < 100:
+      cur_round += 1
+      print(f"ROUND {cur_round} Prompt:\n{prompt}")
+      response = chat.send_message(
+          prompt or ' ', stream=False,
+          safety_settings=SAFETY_CONFIG).text  # type: ignore
+
+      print(f"ROUND {cur_round} LLM response:\n{response}")
+      if _parse_tag(response, 'conclusion'):
```
oliverchang, line 124: Could we start refactoring all of this into a more generalized framework that we can reuse and extend? e.g. something along the lines of:

```python
class AgentPromptSession:
  initial_prompt: str

  def add_response_handler(tag_name: str, action: Callable):
    ...
   

class FixerPromptSession(AgentPromptSession):
  def __init__(self):
    self.initial_prompt = ...
    self.add_response_handler('conclusion', self._handle_conclusion)
    self.add_response_handler('bash', self._handle_bash)
```

(just first pass at a suggestion above -- it's might not be the best way to structure this).

oliverchang (commented):
very nice!

experimental/agent/gen_agent.py
DonggeLiu: Yep, thanks!
An `lldb-agent` is also planned and will be added soon.
I will follow this when designing the framework for them all.

DonggeLiu:
Implementing this PoC via a sequence of PRs in #558

PR #536, opened 2024-08-13
Ask `pylint` to check lazy logging

DonggeLiu:
Using lazy % formatting in logging functions provides better thread safety as string evaluation is done in the logging module.


DonggeLiu:
TBR

PR #545, opened 2024-08-18
Add support for Azure's GPT model

MarkLee131:
Added classes to support Azure's GPT model usage, which inherits from the class `GPT` according to the tutorial of Azure: 
[Create a new Python application](https://learn.microsoft.com/en-us/azure/ai-services/openai/quickstart?tabs=command-line%2Cpython-new&pivots=programming-language-python#create-a-new-python-application)

Our changes: 
- overwrite the function `query_llm`, since the client is a bit different from the original GPT calling.
- overwrite the function `estimate_token_num` because we added a suffix (i.e., `-azure`) for the model name to distinguish the GPT model from the original GPT model (`gpt-3.5-turbo`, `gpt-4`, and `gpt-4o`). 

We have tested it in my local environment, it works well for all of the above Azure's GPT models.

Hope it helps.

### DIFF
--- a/llm_toolkit/models.py
+++ b/llm_toolkit/models.py
@@ -204,7 +204,7 @@ class GPT(LLM):
     """Estimates the number of tokens in |text|."""
     # https://cookbook.openai.com/examples/how_to_count_tokens_with_tiktoken
     try:
-      encoder = tiktoken.encoding_for_model(self.name)
+      encoder = tiktoken.encoding_for_model(self.name.replace('-azure', ''))
     except KeyError:
       logger.info('Could not get a tiktoken encoding for %s.', self.name)
       encoder = tiktoken.get_encoding('cl100k_base')
@@ -235,7 +235,14 @@ class GPT(LLM):
       logger.info('OpenAI does not allow temperature list: %s',
                   self.temperature_list)
 
-    client = openai.OpenAI(api_key=os.getenv('OPENAI_API_KEY'))
+    if self.name.endswith('azure'):
+      client = openai.AzureOpenAI(azure_endpoint=os.getenv(
+        "AZURE_OPENAI_ENDPOINT", "https://api.openai.com"),
+                                api_key=os.getenv("AZURE_OPENAI_API_KEY",
+                                                  "YOUR_API_KEY"),
+                                api_version="2024-02-01")
+    else:    
+      client = openai.OpenAI(api_key=os.getenv('OPENAI_API_KEY'))
 
     completion = self.with_retry_on_error(
         lambda: client.chat.completions.create(messages=prompt.get(),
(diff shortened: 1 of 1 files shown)


llm_toolkit/models.py
```diff
@@ -263,6 +263,78 @@ class GPT4o(GPT):
...
+
+    num_tokens = 0
+    for message in text:
+      num_tokens += 3
+      for key, value in message.items():
+        num_tokens += len(encoder.encode(value))
+        if key == 'name':
+          num_tokens += 1
+    num_tokens += 3
+    return num_tokens
```
mihaimaruseac, line 291: This seems similar to the GPT classes (it's a wrapper of OpenAI's GPT). Can you re-use the existing classes instead?

The only change seems to be getting the client.

mihaimaruseac (changes requested):

llm_toolkit/models.py
MarkLee131: Yes! It would be better. I have updated it.:)

llm_toolkit/models.py
```diff
@@ -204,7 +204,7 @@ def estimate_token_num(self, text) -> int:
     """Estimates the number of tokens in |text|."""
     # https://cookbook.openai.com/examples/how_to_count_tokens_with_tiktoken
     try:
-      encoder = tiktoken.encoding_for_model(self.name)
+      encoder = tiktoken.encoding_for_model(self.name.replace('-azure', ''))
```
mihaimaruseac, line 207: I missed this one before, I thought it was only the client that changed.

This looks good but I have one more suggestion to improve readability and allow for potential future extensions: We should add two methods: `_get_tiktoken_encoding` and `_get_client` to this class. Here they would return the GPT values and then override them in `AzureGPT` below to return the Azure versions.

mihaimaruseac (commented):
This is great, thank you!

I have one more suggestion to make the code more resilient

llm_toolkit/models.py
MarkLee131: That's ok, I have finished the update. :)  @mihaimaruseac

MarkLee131: Also, I can help to update the `README` and `Usage` doc for the Azure support & usage if the above code looks good. :)

llm_toolkit/models.py
```diff
@@ -263,6 +272,41 @@ class GPT4o(GPT):
...
+
+  name = 'gpt-3.5-turbo-azure'
+
+  def _get_tiktoken_encoding(self):
+    """Returns the tiktoken encoding for the model."""
+    try:
+      return tiktoken.encoding_for_model(self.name.replace('-azure', ''))
+    except KeyError:
+      logger.info('Could not get a tiktoken encoding for %s.', self.name)
+      return tiktoken.get_encoding('cl100k_base')
```
mihaimaruseac, line 286: Very minor nit: We can also make the method receive as argument the name of the model (in the parent class), then this one would just be

```
super()._get_tiktoken_encoding(self.name.replace('-azure', ''))
```

and the rest of the lines would be handled by the parent class.

But let's leave this for future work

llm_toolkit/models.py
```diff
@@ -263,6 +272,41 @@ class GPT4o(GPT):
...
+
+  def _get_client(self):
+    """Returns the Azure client."""
+    return openai.AzureOpenAI(azure_endpoint=os.getenv(
+        "AZURE_OPENAI_ENDPOINT", "https://api.openai.com"),
+                              api_key=os.getenv("AZURE_OPENAI_API_KEY",
+                                                "YOUR_AZURE_OPENAI_API_KEY"),
```
mihaimaruseac, line 293:
```suggestion
                              api_key=os.getenv("AZURE_OPENAI_API_KEY"),
```

The other one is not a valid key, and we'd prefer to error if the key is not defined rather than trying with an invalid one.

mihaimaruseac (approved):

mihaimaruseac (approved):

llm_toolkit/models.py
MarkLee131: > Also, I can help to update the `README` and `Usage` doc for the Azure support & usage if the above code looks good. :)

Hi @mihaimaruseac . Do I need to update these two files in this PR? If so, I can also do the update below at the same time:
> Very minor nit: We can also make the method receive as argument the name of the model (in the parent class), then this one would just be
`super()._get_tiktoken_encoding(self.name.replace('-azure', ''))`
and the rest of the lines would be handled by the parent class.

mihaimaruseac: Sorry, I forgot to trigger some of the CI and was waiting on it to complete before merging. This is ready to merge, I'll merge it by the end of today.

Please feel free to do the README/Usage and `super` changes in a new PR, or in this one. It's up to you

PR #552, opened 2024-08-20
enable end-to-end generation

DavidKorczynski:
This adds a script that performs end-to-end generation. Specifically, the script will generate OSS-Fuzz projects given GitHub URLs to an arbitrary project, find the top performing projects and then pass these to OSS-Fuzz-gen core for further harness generation.

There is still more work to do, such as combining all results into one larger OSS-Fuzz project, as well adjusting the UI to be a bit more meaningful for non-oss-fuzz-integrated projects. This will happen in follow-ups.

Ref: https://github.com/google/oss-fuzz-gen/issues/450


DavidKorczynski:
The below picture is of running `MODEL=gpt-3.5-turbo TARGETS=targets.txt ./run_all.sh`

where 
```
$ cat targets.txt
https://github.com/gregjesl/simpleson
https://github.com/rafagafe/tiny-json
https://github.com/vincenthz/libjson
```

Screenshot 2024-08-20 202518

arthurscchan (approved):

PR #563, opened 2024-08-27
[DO NOT MERGE] Compare the performance of Gemini-1.5 flash and pro

DonggeLiu:
Using a medium size benchmark set (1 benchmark per ~300 projects), selected from [`08-24` full exp](https://llm-exp.oss-fuzz.com/Result-reports/scheduled/2024-08-24-weekly-all/).
All projects in this set have at least one compilable fuzz target, so we know that if a benchmark fails, it is not due to the project.
One benchmark is uniformly randomly selected from each project to ensure they are not biased toward Gemini-1.5-Pro.


DonggeLiu:
JOB: https://pantheon.corp.google.com/kubernetes/job/us-central1/llm-experiment-large/default/manual-08-27/details?e=-13802955&mods=logs_tg_prod&project=oss-fuzz

REPORT: https://llm-exp.oss-fuzz.com/Result-reports/scheduled/2024-08-27-weekly-all/

DonggeLiu:
Comparison done.

PR #564, opened 2024-08-27
test-to-harness: use new API endpoint to extract test code

DavidKorczynski:
This makes it possible to extract a subset of test, for example when a test file is too long.


DavidKorczynski:
/gcbrun exp -n dk-ft-100 -m vertex_ai_gemini-1-5 -b minor-for-ci -i

arthurscchan (approved):

PR #566, opened 2024-08-28
run_all_experiments: update project summary data on-the-fly

DavidKorczynski:
Prior to this commit project summary stats were generated only when all experiments had compled, i.e. all benchmarks running should terminate. This causes issues when, for some reason, some benchmarks may not terminate properly or terminate at all. This was observed on some large-scale experiments where 10-20 benchmarks did not finish, and, consequently the full run_experiment_all didn't terminate proper. However, in this case we stil want project summary stats for the 1000+ other benchmarks that have been analyzed.

This fixes it by updated benchmarks stats after each completed benchmark run.

### DIFF
--- a/run_all_experiments.py
+++ b/run_all_experiments.py
@@ -313,7 +313,6 @@ def _print_and_dump_experiment_result(result: Result):
   add_to_json_report(WORK_DIR, 'project_summary', coverage_gain_dict)
 
 
-
 def _print_experiment_results(results: list[Result],
                               cov_gain: dict[str, dict[str, Any]]):
   """Prints the |results| of multiple experiments."""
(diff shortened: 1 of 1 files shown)


DavidKorczynski:
/gcbrun exp -n dk-ft-102 -m vertex_ai_gemini-1-5 -b minor-for-ci -i

DavidKorczynski:
First experiment is looking good -- as well as all locals

DavidKorczynski:
/gcbrun exp -n dk-ft-105 -m vertex_ai_gemini-1-5 -b from-test-small -i

DavidKorczynski:
second experiment using test migration benchmarks is also looking good with results updating during run

arthurscchan (approved):

PR #569, opened 2024-09-02
Agent-enhancement integration (Step 2)

DonggeLiu:
Implemented:
1. Scaffold for pipeline, stages, agents, and tools.
2. Prototyper for local experiments.
3. LLM Chat model.
4. Placeholders for other stages/agetns/tools/models.
5. Updated dependencies to address known bugs/vuls.

TODO:
1. Prototyper for cloud experiments.
2. Other agents: Evaluator, Analyzer.
3. Better logging and file saving.


DonggeLiu:
The result of a local experiment:
```
2024-09-02 15:02:19,578 [PID: 37604] INFO [run_all_experiments._print_experiment_results]: ================================================================================
*ada-url, bool ada_can_parse_with_base(const char *, size_t, const char *, size_t)*
build success rate: 0.5, crash rate: 0.0, found bug: 0, max coverage: 0.0, max line coverage diff: 0.0
max coverage sample:
max coverage diff sample:
max coverage diff report: None
```

There are some "false negative build failures" because LLM modified `build.sh` and built the fuzz target binary into a different name, e.g.:
```log
+ clang++ -O1 -fno-omit-frame-pointer -gline-tables-only -Wno-error=enum-constexpr-conversion -Wno-error=incompatible-function-pointer-types -Wno-error=int-conversion -Wno-error=deprecated-declarations -Wno-erro
r=implicit-function-declaration -Wno-error=implicit-int -DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION -fsanitize=address -fsanitize-address-use-after-scope -fsanitize=fuzzer-no-link -O1 -fno-omit-frame-pointer -gli
ne-tables-only -Wno-error=enum-constexpr-conversion -Wno-error=incompatible-function-pointer-types -Wno-error=int-conversion -Wno-error=deprecated-declarations -Wno-error=implicit-function-declaration -Wno-error
=implicit-int -DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION -fsanitize=address -fsanitize-address-use-after-scope -fsanitize=fuzzer-no-link -stdlib=libc++ -fsanitize=fuzzer ./ada.o ada_c.o -o /out/ada_c
+ cp /src/ada-url/fuzz/url.dict /src/ada-url/fuzz/ada_c.options /src/ada-url/fuzz/parse.options /out/

2024-09-02 15:02:01,983 [PID: 37604] DEBUG [prototyper._validate_fuzz_target_and_build_script]: ROUND 98 Fuzz target compile Succeessfully: True
2024-09-02 15:02:01,983 [PID: 37604] DEBUG [container_tool.execute]: Executing command (ls /out/parse) in b4a0f1ad77527a5fc63e88c3136f9fc230a02811a54b5cb747a83e1e7da4e5a5:
2024-09-02 15:02:02,102 [PID: 37604] DEBUG [container_tool._execute_command]: Executing command (['docker', 'exec', 'b4a0f1ad77527a5fc63e88c3136f9fc230a02811a54b5cb747a83e1e7da4e5a5', '/bin/bash', '-c', 'ls /out
/parse']) in container b4a0f1ad77527a5fc63e88c3136f9fc230a02811a54b5cb747a83e1e7da4e5a5: Return code 2. STDOUT: , STDERR: ls: cannot access '/out/parse': No such file or directory

2024-09-02 15:02:02,103 [PID: 37604] DEBUG [prototyper._validate_fuzz_target_and_build_script]: ROUND 98 Final fuzz target binary exists: False
2024-09-02 15:02:02,391 [PID: 37604] DEBUG [container_tool._execute_command]: Executing command (['docker', 'stop', 'b4a0f1ad77527a5fc63e88c3136f9fc230a02811a54b5cb747a83e1e7da4e5a5']): Return code 0. STDOUT: b4
a0f1ad77527a5fc63e88c3136f9fc230a02811a54b5cb747a83e1e7da4e5a5
```

pipeline.py
```diff
@@ -0,0 +1,31 @@
...
+import argparse
+
+from result_classes import Result
+from stage.analysis_stage import AnalysisStage
+from stage.evaluation_stage import EvalationStage
+from stage.writing_stage import WritingStage
+
+
+class Pipeline():
+  """The fuzzing main pipeline, with 3 iterative stages."""
```
oliverchang, line 11: can you add some more details (e.g. one sentence per stage) in the docstring?

pipeline.py
```diff
@@ -0,0 +1,31 @@
...
+
+  def _terminate(self, results: list[Result]) -> bool:
+    """Validates if the termination conditions have been satisfied."""
+    return bool(results and results[-1].fuzz_target_source)
+
+  def _execute_one_cycle(self, results: list[Result]) -> None:
+    """Executes the stages once."""
+    results.append(self.writing_stage.execute(prev_stage_results=results))
+
+  def execute(self, results: list[Result]) -> list[Result]:
```
oliverchang, line 27: Can you explain a bit the execution strategy/loop here? Why do we repeatedly execute every stage ?

result_classes.py
```diff
@@ -0,0 +1,81 @@
+"""The data structure of all result kinds."""
```
oliverchang, line 1: nit: let's not call this "result_classes". How about just "results" ?

stage/evaluation_stage.py
```diff
@@ -0,0 +1,10 @@
+"""Evaluating the fuzz target and build script."""
+from result_classes import Result
+from stage.base_stage import BaseStage
+
+
+class EvalationStage(BaseStage):
```
oliverchang, line 6: docstring?

tool/base_tool.py
```diff
@@ -0,0 +1,17 @@
+"""The abstract base class for tools used by LLM agents."""
+from abc import ABC, abstractmethod
+from typing import Any
+
+
+class BaseTool(ABC):
```
oliverchang, line 6: please add a docstring

stage/writing_stage.py
```diff
@@ -0,0 +1,34 @@
...
+logging.basicConfig(level=logging.DEBUG,
+                    format=('%(asctime)s [PID: %(process)d] %(levelname)s '
+                            '[%(module)s.%(funcName)s]: %(message)s'))
+
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.DEBUG)
+
+
+class WritingStage(BaseStage):
+  """Writing to the fuzz target and build script."""
```
oliverchang, line 16: can you add some more information on this stage in the docstring? i.e. what are the expected inputs and outputs?

stage/base_stage.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+class BaseStage(ABC):
+  """The abstract base class for stages in fuzzing pipeline."""
+
+  def __init__(self, args: argparse.Namespace) -> None:
+    self.args = args
+    self.agents: list[BaseAgent] = []
+
+  def add_agent(self, agent: BaseAgent) -> 'BaseStage':
```
oliverchang, line 16: do we need the ' around the `-> 'BaseStage'` here ?

stage/writing_stage.py
```diff
@@ -0,0 +1,34 @@
...
+
+  def _write_new_fuzz_target(self, prev_results: list[Result]) -> Result:
+    """Writes a new fuzz target."""
+    return self.get_agent('Prototyper').execute(prev_results)
```
oliverchang, line 20: If particular stages expect agents of certain types to exist, should we ensure they are passed in the constructor? that way, we can never get into a state where they don't exist.

prompts/agent/container_tool.txt
```diff
@@ -0,0 +1,40 @@
+To help write a simple but valid fuzz target, your can investigate any files in the fuzz target's build environment by iteratively send me a BASH command, and I will provide you with the output of the command.
```
oliverchang, line 1: should we rename this directory "agent_tools" to be clearer?

oliverchang (commented):
nice!

pipeline.py
DonggeLiu: Yep, good point. Done!

pipeline.py
DonggeLiu: Done!

result_classes.py
DonggeLiu: Yep, fixed.
Should it be in a separated dir as well (e.g., `experiment` or `common`)?

stage/evaluation_stage.py
DonggeLiu: Done

tool/base_tool.py
DonggeLiu: Done

stage/writing_stage.py
DonggeLiu: Yep, added.

stage/base_stage.py
DonggeLiu: Yes, because this is in `BaseStage`. Otherwise linter will complain:
image

stage/writing_stage.py
DonggeLiu: Done, now we pass all agents expected by each stage in the pipeline constructor.

prompts/agent/container_tool.txt
DonggeLiu: Fixed.
I was thinking of putting everything into `prompts/agent/` for now and replacing existing prompts with them once this is ready, but your suggestion is better.
I moved it to `prompts/tool/` now.

result_classes.py
oliverchang: we can just keep it where it is for now and move it later if needed.

DonggeLiu: Shall we merge this?
The next step is enabling Prototyper for cloud experiments.
Once done, we will be able to run a full experiment.

agent/prototyper.py
```diff
@@ -0,0 +1,228 @@
...
+    benchmark = results[-1].benchmark
+
+    default_prompt_builder = DefaultTemplateBuilder(model=self.llm,
+                                                    benchmark=benchmark)
+    prompt = default_prompt_builder.build([])
+    work_dirs = results[-1].work_dirs
+    prompt.save(work_dirs.prompt)
+    return prompt
+
+  def _parse_tag(self, response: str, tag: str) -> str:
```
oliverchang, line 64: should some of these go into `BaseAgent`? I imagine a lot of these things related to formatting, parsing/executing commands would be required for all agents?

agent/prototyper.py
```diff
@@ -0,0 +1,228 @@
...
+    """Validates LLM conclusion or executes its command."""
+    logger.debug('ROUND %d LLM response: %s', cur_round, response)
+    if self._parse_tag(response, 'conclusion'):
+      return self._container_handle_conclusion(cur_round, response,
+                                               build_result)
+    return self._container_handle_command(cur_round, response)
+
+  def execute(self, prev_results: list[Result]) -> BuildResult:
+    """Executes the agent based on previous result."""
+    logger.info('Trial: %d Executing Prototyper', prev_results[-1].trial)
```
oliverchang: this `prev_results[-1]` pattern seems a bit confusing. 

Is it always guaranteed that this list is not empty? What happens on the very first run? And also, why does it have to be a list? 

Also, "Previous" implies that these are **only** results from a previous execution, but here it seems like it contains the current trial as well.

agent/prototyper.py
DonggeLiu: Yes, good point, thanks!
I will go through them and move suitable ones to the base class.

agent/prototyper.py
DonggeLiu: > this prev_results[-1] pattern seems a bit confusing.

AH I think we no longer have this in the latest code. I made `trial` a class attribute.

> Is it always guaranteed that this list is not empty?

Yes, the list is never empty. The list contains 1 `Result` item in the first iteration, storing [the benchmark-related and experiment-related information](https://github.com/google/oss-fuzz-gen/pull/569/files/ef9bcd4689a639cdd18ecdccd82bb7998deceb3c#diff-0d648b28bb6b85213eb0ff1c3954edd4b710ff8b219eb7dc89b294108dfedb3eR316). I did this mainly for simplicity reasons: All stages/agents require those info (function signature, project, work directory, etc.) and they all need the previous result too, so I pack those info in Result and pass `Result` around. It also makes sense for `Result` to store those because we can easily know the result's corresponding benchmark/trail/project/dir. I can use parameters to avoid passing a result item at the first run, but I suspect it will complicate data sharing a little bit.

> And also, why does it have to be a list?

Later, we may find the result history helpful for comparisons. For example, the Analysis Stage can look up `Result` item 1 with fuzz target source `F1` covering project code blocks `C1`, and compare it against `Result` item 2 with `F2` covering `C2`. Similar comparison can benefit Enchander in fixing fuzz targets.


> Also, "Previous" implies that these are only results from a previous execution, but here it seems like it contains the current trial as well.

Yep good point. I think that has been fixed in the latest code. I renamed them to `result_history`.

oliverchang (approved):

agent/prototyper.py
DonggeLiu: Done below

DonggeLiu:
Running an experiment locally before merging.

PR #220, opened 2024-04-15
Add sample Java auto generation project

arthurscchan:

PR #572, opened 2024-09-03
JVM: New OSS-Integration for Java project by OSS-Fuzz-Gen

arthurscchan:
This PR is based on #220 and proposes adding a new approach in OSS-Fuzz-Gen through `run_all_experiments.py`. The new approach targets Java projects with a GitHub repository that are not currently integrated into OSS-Fuzz and is intended to analyse and generate a new OSS-Fuzz integration for those projects using the OSS-Fuzz-Gen LLM approach.


run_all_experiments.py
```diff
@@ -91,19 +93,73 @@ def generate_benchmarks(args: argparse.Namespace) -> None:
...
+      benchmarks = introspector.populate_benchmarks_using_introspector(
+          project, project_lang, args.generate_benchmarks_max,
+          benchmark_oracles)
+      if benchmarks:
+        benchmarklib.Benchmark.to_yaml(benchmarks, benchmark_dir)
+  else:
+    # Checkout OSS-Fuzz for static analysis
+    oss_fuzz_checkout.clone_oss_fuzz()
+
+    # Generate benchmarks for new projects from scratch
```
DavidKorczynski, line 118: Could you separate this out of the function? It's getting too long.

DavidKorczynski (changes requested):
Minor nit before in-depth review.

Instead of `from_scratch` could you put it in a folder `auto-build` and then a subfolder under there `jvm` for Java-specific logic?

arthurscchan:
/gcbrun exp -n ascc-jvm-pr-572-1 -m vertex_ai_gemini-1-5 -i -b jvm-medium

experiment/benchmark.py
```diff
@@ -136,6 +157,86 @@ def from_yaml(cls, benchmark_path: str) -> List:
...
                 use_context=use_context,
+                is_new_integration=data['is_new_integration'],
+                jvm_special_properties=function.get('jvm_special_properties'),
+                build_project_name=data.get('build_project_name'),
+                function_dict=function))
+
+    return benchmarks
+
+  @classmethod
+  def from_java_data_yaml(cls, benchmark_path: str, project: str,
```
DavidKorczynski, line 168: I don't think we should have this here.

Rather, can you make it such the auto-building will do the following steps:

1) Find one or more builds that work (can we have multiple builds that work and each have some different characteristics, e.g. may include more/less code in the introspector report?)
2) For the build that works, find a harness that works, it could be empty for that matter, but where the fuzz introspector report includes all the information about the project under analysis
3) use the build that works combined with the harness that works and pass this into the regular processing OFG pipeline, and then let the regular OFG pipeline generate the targets via the existing procedures?

DavidKorczynski (changes requested):

experiment/benchmark.py
DavidKorczynski: So, instead of having a "generate benchmarks from scratch" I'd rather have a : "generate a working build with a primitive benchmark, and use the existing OFG pipeline to take care of the rest"

arthurscchan: That may have problems. Since we are targeting github repository that we assume does not have previous OSS-Fuzz integration. And thus it cannot go for the OFG pipeline because in the original pipeline, it got to query quite some amount to the FI webapp. Which in this new logic is all done in the benchmark which contains all the necessary information, making the pipeline doesn't need to query to FI during prompt generation. But I do agree that this should be done in a separate location. I will consider how should I relocate and revamp this part to make it easier to follow.

arthurscchan: Ok. I will revamp the code to consider that.

experimental/jvm/README.md
```diff
@@ -0,0 +1,65 @@
+# Auto-build OSS-Fuzz Projects from a GitHub Repository URL for Java Projects
+
+This directory contains the logic for running auto-harness generation on projects
```
DavidKorczynski, line 3: The two paragraphs here are only relevant for the C/CPP logic, could you provide documentation targeted the Java integration?

DavidKorczynski (changes requested):

DavidKorczynski (approved):

experimental/jvm/generate_projects.py
```diff
@@ -0,0 +1,97 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Manager for running auto-gen from scratch."""
+
+import argparse
+import logging
+import os
+import shutil
+
+import utils
```
DonggeLiu, line 22: Could you change this to a more specific import statement please?
E.g., `from experimental.jvm import utils`.
This statement is ambiguous for Python and will fail if there is a `util.py` in the dir with a higher search priority (e.g., the root of this repo).

PR #577, opened 2024-09-05
introspector: introduce harness source/executable retrieval via FI 

DavidKorczynski:
The goal is to phase out the project_src.search_source strategy
by pushing this to FI. I have set it as optional for now while
we're awaiting an OSS-Fuzz update on FI. However, this improves
performance, since we don't have to search through /src/ for
each benchmark, and we can also push much of the e.g.
bazel-recognition logic into FI.

### DIFF
--- a/data_prep/introspector.py
+++ b/data_prep/introspector.py
@@ -40,6 +40,8 @@ T = TypeVar('T', str, list, dict, int)  # Generic type.
 TIMEOUT = 45
 MAX_RETRY = 5
 
+USE_FI_TO_GET_TARGETS = bool(int(os.getenv('OSS_FI_TO_GET_TARGETS', '0')))
+
 # By default exclude static functions when identifying fuzz target candidates
 # to generate benchmarks.
 ORACLE_AVOID_STATIC_FUNCTIONS = bool(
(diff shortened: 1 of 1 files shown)


DavidKorczynski:
Locally the performance on this is nice, but I'll wait making this the default until we resolve the bazel issues. For example, the current approach gives us for e.g. abseil-cpp:

```
"target_name": "string_escape_fuzzer"
"target_path": "/proc/self/cwd/string_escape_fuzzer.cc"

```
where target path should really be `"target_path": "/src/string_escape_fuzzer.cc"`. This will come with the next FI bump on OSS-Fuzz

arthurscchan (approved):

oliverchang:
Are we ready to turn this on by default now?

DavidKorczynski:
Yeah -- there is still one limitation into our automated approach, which will require some heuristics to solve. The problem is when some harnesses has a broader set of linkage than other harnesses, we'd want to make sure our auto-generated harness uses the rules from the ones with a broad linkage.

Take `abseil-cpp`: https://github.com/google/oss-fuzz/blob/94bb58f8bae6e2bfb8a499a51ec66423ccfc6d32/projects/abseil-cpp/BUILD#L19-L41

We prefer to use string_utilities_fuzzer rather than string_escape_fuzzer for linking. I think, though, it will take longer to come up with a solution here.

PR #580, opened 2024-09-06
Agent enhancement 3

DonggeLiu:
Support cloud experiments:
1. Upload local pickle files for `agent` and `result_history`
2. In cloud build: Install dependecies and run `new_result = agent.execute(result_history)`
3. Download the pickle file of `new_result`
4. Save cloud exp logs
5. If cloud build fails, return a default `new_result` representing build failure.
6. Convert results into `status/**/result.json` for report generation.
7. Upload local OFG repo to cloud build, instead of asking cloud build to pull from repo.
8. Make new `Result` class compatible with report generation.

TODOs:
2. More exception handlings.
3. More intermediate files (local agent dialog)

### DIFF
--- a/Dockerfile.cloudbuild-agent
+++ b/Dockerfile.cloudbuild-agent
@@ -1,3 +1,5 @@
+# TODO(dongge): Automatically build and push this to registry daily:
+# us-central1-docker.pkg.dev/oss-fuzz/oss-fuzz-gen/agent-image
 FROM ubuntu:22.04
 
 ENV DEBIAN_FRONTEND=noninteractive
--- a/Dockerfile.cloudbuild-agent
+++ b/Dockerfile.cloudbuild-agent
@@ -1,0 +1,36 @@
+FROM ubuntu:22.04
+
+ENV DEBIAN_FRONTEND=noninteractive
+
+# Install Python 3.11 and pip
+RUN apt-get update && \
+    apt-get install -y software-properties-common curl && \
+    add-apt-repository ppa:deadsnakes/ppa && \
+    apt-get update && \
+    apt-get install -y python3.11 python3.11-dev python3.11-venv \
+    python3.11-distutils && \
+    curl -sS https://bootstrap.pypa.io/get-pip.py | python3.11
+
+# Install Docker
+RUN apt-get install -y ca-certificates gnupg lsb-release && \
+    mkdir -p /etc/apt/keyrings && \
+    curl -fsSL https://download.docker.com/linux/ubuntu/gpg | \
+    gpg --dearmor -o /etc/apt/keyrings/docker.gpg && \
+    echo "deb [arch=$(dpkg --print-architecture) \
+    signed-by=/etc/apt/keyrings/docker.gpg] \
+    https://download.docker.com/linux/ubuntu $(lsb_release -cs) stable" | \
+    tee /etc/apt/sources.list.d/docker.list > /dev/null && \
+    apt-get update && \
+    apt-get install -y docker-ce docker-ce-cli containerd.io \
+    docker-buildx-plugin docker-compose-plugin
+
+ENV DEBIAN_FRONTEND=dialog
+
+# Set the working directory
+WORKDIR /workspace/ofg
+
+# Copy the requirements file
+COPY requirements.txt /workspace/ofg/
+
+# Install Python dependencies
+RUN pip3.11 install --ignore-installed -r /workspace/ofg/requirements.txt
--- a/agent/base_agent.py
+++ b/agent/base_agent.py
@@ -30,13 +30,9 @@ class BaseAgent(ABC):
     self.name: str = name or self.__class__.__name__
     self.dialog: str = ''  # Communication history between LLM and tool.
 
-  def write_to_file(self, file_path: str, file_content: str):
-    with open(file_path, 'w') as file:
-      file.writelines(file_content)
-
   def log(self, msg: str, level=logging.INFO):
     """Method to log messages dynamically using the logger factory."""
-    agent_logger = logger.get_logger(self.name, self.trial, level)
+    agent_logger = logger.get_trial_logger()
     agent_logger.log(level, msg)
 
   def get_tool(self, tool_name: str) -> Optional[BaseTool]:
--- a/agent/base_agent.py
+++ b/agent/base_agent.py
@@ -88,33 +88,33 @@ class BaseAgent(ABC):
   def _parse_args(cls) -> argparse.Namespace:
     """Parses command line args."""
     parser = argparse.ArgumentParser(
-        description='Execute agent in cloud with pickle files.')
+        description='Execute agent in cloud with dill files.')
     parser.add_argument('-a',
                         '--agent',
-                        help='The pickle file path for the agent to execute.')
+                        help='The dill file path for the agent to execute.')
     parser.add_argument(
         '-rh',
         '--result-history',
-        help='The pickle file path for the agent input result history.')
+        help='The dill file path for the agent input result history.')
     parser.add_argument(
         '-rn',
         '--result-new',
-        help='The pickle file path to store the agent output new result.')
+        help='The dill file path to store the agent output new result.')
     return parser.parse_args()
 
   @classmethod
   def cloud_main(cls) -> None:
-    """Executes agent using pickle files. This is for cloud experiments launched
+    """Executes agent using dill files. This is for cloud experiments launched
     by cloud_builder.py. It runs `new_result = agent.execute(result_history)` in
     the same way as local experiments, except `agent` and `result_history` are
-    deserialized from pickle files and new_result will be pickled to share data
+    deserialized from dill files and new_result will be serialized to share data
     with the cloud experiment requester."""
     args = cls._parse_args()
 
-    agent = utils.deserialize_from_pickle(args.agent)
-    result_history = utils.deserialize_from_pickle(args.result_history)
+    agent = utils.deserialize_from_dill(args.agent)
+    result_history = utils.deserialize_from_dill(args.result_history)
     result = agent.execute(result_history)
-    utils.serialize_to_pickle(result, args.result_new)
+    utils.serialize_to_dill(result, args.result_new)
 
   @abstractmethod
   def _initial_prompt(self, results: list[Result]) -> Prompt:
--- a/agent/base_agent.py
+++ b/agent/base_agent.py
@@ -7,6 +7,7 @@ from abc import ABC, abstractmethod
 from typing import Optional
 
 import logger
+import utils
 from llm_toolkit.models import LLM
 from llm_toolkit.prompt_builder import DefaultTemplateBuilder
 from llm_toolkit.prompts import Prompt
@@ -79,6 +80,34 @@ class BaseAgent(ABC):
                      f'{tool.tutorial()}')
     return DefaultTemplateBuilder(self.llm, None, initial=prompt_text).build([])
 
+  @classmethod
+  def _parse_args(cls) -> argparse.Namespace:
+    """Parses command line args."""
+    parser = argparse.ArgumentParser(
+        description='Execute agent in cloud with pickle files.')
+    parser.add_argument('-a',
+                        '--agent',
+                        help='The pickle file path for the agent to execute.')
+    parser.add_argument(
+        '-rh',
+        '--result-history',
+        help='The pickle file path for the agent input result history.')
+    parser.add_argument(
+        '-rn',
+        '--result-new',
+        help='The pickle file path to store the agent output new result.')
+    return parser.parse_args()
+
+  @classmethod
+  def cloud_main(cls) -> None:
+    """Executes agent using pickle files. This is for cloud experiments."""
+    args = cls._parse_args()
+
+    agent = utils.deserialize_from_pickle(args.agent)
+    result_history = utils.deserialize_from_pickle(args.result_history)
+    result = agent.execute(result_history)
+    utils.serialize_to_pickle(result, args.result_new)
+
   @abstractmethod
   def _initial_prompt(self, results: list[Result]) -> Prompt:
     """The initial prompt of the agent."""
@@ -89,17 +118,5 @@ class BaseAgent(ABC):
 
 
 if __name__ == "__main__":
-  # Make this a class method so that child agents can inherit and reuse?
-  import sys
-
-  import utils
-
-  agent_pickle = sys.argv[1]
-  result_history_pickle = sys.argv[2]
-  new_result_pickle = sys.argv[3]
-
-  agent = utils.deserialize_from_pickle(agent_pickle)
-  result_history = utils.deserialize_from_pickle(result_history_pickle)
-
-  result = agent.execute(result_history)
-  utils.serialize_to_pickle(result, new_result_pickle)
+  # For cloud experiments.
+  BaseAgent.cloud_main()
--- a/agent/base_agent.py
+++ b/agent/base_agent.py
@@ -1,8 +1,10 @@
 """The abstract base class for LLM agents in stages."""
 import argparse
 import logging
+import random
 import re
 import subprocess as sp
+import time
 from abc import ABC, abstractmethod
 from typing import Optional
 
@@ -74,6 +76,13 @@ class BaseAgent(ABC):
                      'interaction protocols:\n'
                      f'{tool.tutorial()}')
     return DefaultTemplateBuilder(self.llm, None, initial=prompt_text).build([])
+
+  def _sleep_random_duration(self, min_sec: int = 1, max_sec: int = 60) -> None:
+    """Sleeps for a random duration between min_sec and max_sec. Agents uses
+    this to avoid exceeding quota limit (e.g., LLM query frequency)."""
+    duration = random.randint(min_sec, max_sec)
+    logger.debug('Sleeping for %d before the next query', duration)
+    time.sleep(duration)
 
   @classmethod
   def _parse_args(cls) -> argparse.Namespace:
--- a/agent/base_agent.py
+++ b/agent/base_agent.py
@@ -95,7 +95,11 @@ class BaseAgent(ABC):
 
   @classmethod
   def cloud_main(cls) -> None:
-    """Executes agent using pickle files. This is for cloud experiments."""
+    """Executes agent using pickle files. This is for cloud experiments launched
+    by cloud_builder.py. It runs `new_result = agent.execute(result_history)` in
+    the same way as local experiments, except `agent` and `result_history` are
+    deserialized from pickle files and new_result will be pickled to share data
+    with the cloud experiment requester."""
     args = cls._parse_args()
 
     agent = utils.deserialize_from_pickle(args.agent)
--- a/agent/base_agent.py
+++ b/agent/base_agent.py
@@ -9,6 +9,7 @@ from typing import Optional
 from llm_toolkit.models import LLM
 from llm_toolkit.prompt_builder import DefaultTemplateBuilder
 from llm_toolkit.prompts import Prompt
+from logger import Logger
 from results import Result
 from tool.base_tool import BaseTool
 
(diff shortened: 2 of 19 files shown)


agent/prototyper.py
```diff
@@ -129,14 +128,14 @@ def _container_handle_conclusion(
...
-                       f'{build_result.trial}.build_script'),
-          build_result.build_script_source)
+      # self.write_to_file(
+      #     os.path.join(build_result.work_dirs.fixed_targets,
+      #                  f'{build_result.trial}.fuzz_target'),
+      #     build_result.fuzz_target_source)
+      # self.write_to_file(
+      #     os.path.join(build_result.work_dirs.fixed_targets,
+      #                  f'{build_result.trial}.build_script'),
+      #     build_result.build_script_source)
```
DonggeLiu, line 138: Currently `write_to_file` only makes sense for local experiments, will address this in a better way soon.

DonggeLiu (commented):
Pretty sure the cloud experiment works now:
```
2024-09-06 18:00:50,409 [PID: 832750] INFO [run_all_experiments._print_experiment_results]: ================================================================================
*ada-url, bool ada_can_parse_with_base(const char *, size_t, const char *, size_t)*
build success rate: 1.0, crash rate: 0.0, found bug: 0, max coverage: 0.0, max line coverage diff: 0.0
```

While I am addressing the minor TODOs in this PR, could you please help me check if there is any structure/design issues? @oliverchang @mihaimaruseac 

Thanks!

agent/base_agent.py
```diff
@@ -93,3 +93,22 @@ def _initial_prompt(self, results: list[Result]) -> Prompt:
...
+if __name__ == "__main__":
+  # Make this a class method so that child agents can inherit and reuse?
+  import logging
+  import sys
+
+  import utils
+
+  agent_pickle = sys.argv[1]
+  result_history_pickle = sys.argv[2]
+  new_result_pickle = sys.argv[3]
```
mihaimaruseac, line 107: Is there a possibility we could avoid pickles?

agent/base_agent.py
```diff
@@ -93,3 +93,22 @@ def _initial_prompt(self, results: list[Result]) -> Prompt:
...
   def execute(self, result_history: list[Result]) -> Result:
     """Executes the agent based on previous result."""
+
+
+if __name__ == "__main__":
+  # Make this a class method so that child agents can inherit and reuse?
+  import logging
+  import sys
+
+  import utils
```
mihaimaruseac, line 103: Should we move the imports to the top?

agent/prototyper.py
mihaimaruseac: I really wish we had a library like [TF's gfile](https://www.tensorflow.org/api_docs/python/tf/io/gfile/GFile). Then we could use the same code path to write either locally or on GCP, etc. depending on how we run things.

But right now this requires importing the full machinery of TF :(

agent/base_agent.py
DonggeLiu: Yep! I will fix this.

agent/base_agent.py
DonggeLiu: Unfortunately, I cannot think of a good solution yet because those variables are pretty complex.
What's the downside of using pickles?

DonggeLiu:
Cloud experiment results looking good:

**** FINAL RESULTS: ****


2024-09-09 07:32:06.300 INFO run_all_experiments - _print_experiment_results: ================================================================================
*ada-url, struct url ada::url ada::parser::parse_url<ada::parse_url<ada::url>(string_view, const struct url *)*
build success rate: 1.0, crash rate: 0.0, found bug: 0, max coverage: 0.0, max line coverage diff: 0.0
max coverage sample:
max coverage diff sample:
max coverage diff report: None

2024-09-09 07:32:06.300 INFO run_all_experiments - _print_experiment_results: ================================================================================
*ada-url, bool ada_can_parse_with_base(const char *, size_t, const char *, size_t)*
build success rate: 0.7, crash rate: 0.0, found bug: 0, max coverage: 0.0, max line coverage diff: 0.0
max coverage sample:
max coverage diff sample:
max coverage diff report: None

2024-09-09 07:32:06.300 INFO run_all_experiments - _print_experiment_results: ================================================================================
*ada-url, struct ada_owned_string ada_idna_to_ascii(const char *, size_t)*
build success rate: 0.6, crash rate: 0.0, found bug: 0, max coverage: 0.0, max line coverage diff: 0.0
max coverage sample:
max coverage diff sample:
max coverage diff report: None

2024-09-09 07:32:06.300 INFO run_all_experiments - _print_experiment_results: ================================================================================
*ada-url, ada_url ada_parse_with_base(const char *, size_t, const char *, size_t)*
build success rate: 0.7, crash rate: 0.0, found bug: 0, max coverage: 0.0, max line coverage diff: 0.0
max coverage sample:
max coverage diff sample:
max coverage diff report: None

2024-09-09 07:32:06.300 INFO run_all_experiments - _print_experiment_results: ================================================================================
*ada-url, struct url ada::url_aggregator ada::parser::parse_url<ada::parse_url<ada::url>(string_view, const struct url *)*
build success rate: 0.8, crash rate: 0.0, found bug: 0, max coverage: 0.0, max line coverage diff: 0.0
max coverage sample:
max coverage diff sample:
max coverage diff report: None

2024-09-09 07:32:06.300 INFO run_all_experiments - _print_experiment_results: **** TOTAL COVERAGE GAIN: ****
2024-09-09 07:32:06.300 INFO run_all_experiments - _print_experiment_results: *ada-url: 0.0

agent/base_agent.py
oliverchang: +1. Pickles are not recommended for security reasons. See the big red box in https://docs.python.org/3/library/pickle.html.

logger.py
```diff
@@ -0,0 +1,70 @@
...
+import os
+
+from results import Result
+
+RESULT_JSON = 'result.json'
+
+_trial_logger = None
+
+
+class CustomLoggerAdapter(logging.LoggerAdapter):
```
oliverchang, line 13: Can you add some docstrings on what this is doing?

logger.py
```diff
@@ -0,0 +1,70 @@
+"""A module to write logs and files."""
```
oliverchang, line 1: Can you add some comments here to explain what this module is for, and why it's required?

agent/base_agent.py
```diff
@@ -29,18 +30,15 @@ def __init__(self,
...
-                                '[%(module)s.%(funcName)s]: %(message)s'))
-
-    self.logger = logging.getLogger(__name__)
-    self.logger.setLevel(logging.DEBUG)
-
   def write_to_file(self, file_path: str, file_content: str):
     with open(file_path, 'w') as file:
       file.writelines(file_content)
 
+  def log(self, msg: str, level=logging.INFO):
```
oliverchang, line 37: nit: can we keep the format string mechanism for logging? 

i.e. add `*args, **kwargs` here and let callers do `self.log('foo %s', 'arg')`.

agent/base_agent.py
```diff
@@ -93,3 +90,20 @@ def _initial_prompt(self, results: list[Result]) -> Prompt:
   @abstractmethod
   def execute(self, result_history: list[Result]) -> Result:
     """Executes the agent based on previous result."""
+
+
+if __name__ == "__main__":
```
oliverchang, line 95: nit: can you add some comments to explain what main is doing? 

Also, put main into its own function and call it from here.

common/cloud_builder.py
```diff
@@ -0,0 +1,258 @@
...
+from results import BuildResult, Result
+
+OFG_REPO = 'https://github.com/google/oss-fuzz-gen.git'
+OF_REPO = 'https://github.com/google/oss-fuzz.git'
+US_CENTRAL_CLIENT_OPTIONS = google.api_core.client_options.ClientOptions(
+    api_endpoint='https://us-central1-cloudbuild.googleapis.com/')
+
+
+class CloudBuilder:
+  """Encapsulate functions to request Google Cloud Builds to execute agents."""
```
oliverchang, line 29: nit: we should put a more descriptive text here, and talk about what this `CloudBuilder` represents, rather than about how this is encapsulating functions.

common/cloud_builder.py
```diff
@@ -0,0 +1,258 @@
...
+    #TODO(dongge): extra tags.
+    _, self.project_id = default()
+    assert self.project_id, 'Cloud experiment requires a Google cloud project.'
+
+    self.bucket_name = args.cloud_experiment_bucket
+    self.bucket = storage.Client().bucket(self.bucket_name)
+
+    # Service account.
+    self.service_account = os.getenv('GOOGLE_SERVICE_ACCOUNT', '')
+    service_account_file = os.getenv('GOOGLE_APPLICATION_CREDENTIALS', '')
```
oliverchang, line 41: Do we need this? Can we just use `google.auth.default()` ? https://google-auth.readthedocs.io/en/master/reference/google.auth.html#google.auth.default

This gets the most suitable credentials in each environment. This could be a service account key, a service account from GCE, or local gcloud creds.

common/cloud_builder.py
```diff
@@ -0,0 +1,258 @@
...
+                'name':
+                    'gcr.io/cloud-builders/docker',
+                'entrypoint':
+                    'bash',
+                'dir':
+                    '/workspace/ofg',
+                'args': [
+                    '-c',
+                    """
+                    # Install Python 3.11 and pip
```
oliverchang, line 127: This is very brittle, and will break if `gcr.io/cloud-builders/docker` updates. Can we create our own Docker image for this with all of these dependencies preinstalled?

agent/base_agent.py
DonggeLiu: Would signing the pickle file via `hmac` be safer than `json` (which may expose the data to humans)?

agent/base_agent.py
DonggeLiu: Yep, I pushed this as a temporary solution to show some quick results.
Ideally, I want to avoid manually defining this function and will think of a better solution.

agent/base_agent.py
oliverchang: I think we can just go with this for now to unblock this work, with a TODO look for a better replacement, or a disclaimer that there must never be untrusted data being passed in through this mechanism. 

>  (which may expose the data to humans)?

Can you elaborate on this concern?

DonggeLiu: > (which may expose the data to humans)?

> Can you elaborate on this concern?

This is unimportant for now, but given we cannot trust the file transfer channel, I was thinking about:
1. Hashing and signing the file to ensure it is not tampered with.
2. Encrypting the file so that others cannot view its content (which may contain new vuls and reproducers).

DonggeLiu:
OK now all structural and technical challenges are resolved, I will start fixing the style/linter/format/nit issues.

agent/base_agent.py
DonggeLiu: Done!

agent/base_agent.py
```diff
@@ -77,19 +67,51 @@ def _container_handle_bash_command(self, cur_round: int, response: str,
...
+
+  @classmethod
+  def cloud_main(cls) -> None:
+    """Executes agent using pickle files. This is for cloud experiments."""
+    args = cls._parse_args()
+
+    agent = utils.deserialize_from_pickle(args.agent)
+    result_history = utils.deserialize_from_pickle(args.result_history)
+    result = agent.execute(result_history)
+    utils.serialize_to_pickle(result, args.result_new)
```
mihaimaruseac, line 104: Let's not mention pickle in the public API, since we plan to replace them later.

`deserialize`/`serialize` (or `serialize_agent`, etc.) and `"Executes agent from saved config"?

agent/prototyper.py
```diff
@@ -179,22 +150,19 @@ def execute(self, result_history: list[Result]) -> BuildResult:
...
-                        self.inspect_tool.container_id,
-                        extra={'trial': self.trial})
+      logger.debug('Stopping and removing the inspect container %s',
+                   self.inspect_tool.container_id)
       self.inspect_tool.terminate()
     return build_result
+
+
+# if __name__ == "__main__":
+#   prototyper = Prototyper()
```
mihaimaruseac, line 168: Nit: this can be removed

common/cloud_builder.py
mihaimaruseac: +1

mihaimaruseac (commented):
Overall I like this, thank you! I'll ping you offline to discuss the pickle parts/plans

DonggeLiu:
The report shows a HUGE improvement:
https://llm-exp.oss-fuzz.com/Result-reports/ofg-pr/2024-09-12-580-ae-comparison/index.html

Baseline:
https://llm-exp.oss-fuzz.com/Result-reports/scheduled/2024-09-12-daily-comparison/index.html

oliverchang:
> The report shows a HUGE improvement: https://llm-exp.oss-fuzz.com/Result-reports/ofg-pr/2024-09-12-580-ae-comparison/index.html
> 
> Baseline: https://llm-exp.oss-fuzz.com/Result-reports/scheduled/2024-09-12-daily-comparison/index.html

Wow! Is there some way to check the generated targets to make sure they're actual proper fuzz targets?

DonggeLiu:
> Wow! Is there some way to check the generated targets to make sure they're actual proper fuzz targets?

Only via the bucket for now, because agents save fuzz targets (and build scripts) in a different way.
I am thinking of showing them (and other things) in the report in a separate PR because I plan to optimize the directory hierarchy, too. The current result dir is a bit confusing (raw_targets in one dir are used by reports for counting only, and all other fixed_targets are in another dir).

arthurscchan:
@DonggeLiu I have fixed the lint.

DonggeLiu:
> @DonggeLiu I have fixed the lint.

Could you please fix BOTH?
https://github.com/google/oss-fuzz-gen/pull/580#issuecomment-2342771523

also could you please do it in a separate PR, instead of pushing to this?

agent/base_agent.py
DonggeLiu: Using `dill` now : )

common/cloud_builder.py
DonggeLiu: Done!

common/cloud_builder.py
DonggeLiu: Added.

agent/base_agent.py
DonggeLiu: fixed.

common/cloud_builder.py
DonggeLiu: Done:
1. Using a separate image.
2. Moved the `Dockerfile` content to a file.
3. Manually pushed the image to our ofg artifact registry.
4. Cloud build now pulls from the registry.

mihaimaruseac (approved):

agent/prototyper.py
```diff
@@ -89,24 +80,18 @@ def _validate_fuzz_target_and_build_script(self, cur_round: int,
...
+                 compile_succeed)
 
     # Double-check binary.
     ls_result = compilation_tool.execute(f'ls /out/{benchmark.target_name}')
     binary_exists = ls_result.returncode == 0
-    self.logger.debug('ROUND %d Final fuzz target binary exists: %s',
-                      cur_round,
-                      binary_exists,
-                      extra={'trial': self.trial})
+    logger.debug('ROUND  Final fuzz target binary exists: %s', cur_round,
```
oliverchang, line 93: missing %d?

ci/request_pr_exp.py
```diff
@@ -152,6 +153,11 @@ def _parse_args(cmd) -> argparse.Namespace:
       default=VARY_TEMPERATURE,
       help=('Use different temperatures for each sample, default: '
             f'{VARY_TEMPERATURE}'))
+  parser.add_argument('-ae',
```
oliverchang, line 156: why not "ag". "ae" seems a bit confusing.

common/cloud_builder.py
```diff
@@ -0,0 +1,261 @@
...
+    self.credentials, self.project_id = default()
+    assert self.project_id, 'Cloud experiment requires a Google cloud project.'
+    assert hasattr(
+        self.credentials,
+        'refresh'), ('Cloud experiment requires a service account email')
+    assert hasattr(self.credentials, 'service_account_email'), (
+        'Cloud experiment requires a service account email')
+
+    try:
+      # TODO(dongge): Understand why this crashes in local experiments.
```
oliverchang, line 47: What's the exception that happens?

common/cloud_builder.py
```diff
@@ -0,0 +1,261 @@
...
+            {
+                'id':
+                    'agent-step',
+                'name':
+                    'gcr.io/cloud-builders/docker',
+                'args': [
+                    'run', '--rm', '-v', '/workspace:/workspace', '-v',
+                    '/var/run/docker.sock:/var/run/docker.sock',
+                    '--network=cloudbuild',
+                    ('us-central1-docker.pkg.dev/oss-fuzz/oss-fuzz-gen/'
```
oliverchang, line 141: Can you add a comment to link this to the relevant Dockerfile?

common/cloud_builder.py
```diff
@@ -0,0 +1,261 @@
...
+
+    # Convert to YAML string and submit the Cloud Build request
+    build_info = self.builds.create(projectId=self.project_id,
+                                    body=cloud_build_config).execute()
+    build_id = build_info.get('metadata', {}).get('build', {}).get('id', '')
+
+    logging.info('Cloud Build ID: %s', build_id)
+    return build_id
+
+  def _build_succeeds(self, build_id: str) -> bool:
```
oliverchang, line 182: nit: rename this to `_wait_for_build` to be more descriptive.

Dockerfile.cloudbuild-agent
```diff
@@ -0,0 +1,36 @@
+FROM ubuntu:22.04
```
oliverchang, line 1: We need to set up some kind of daily automated build for this. Can you add a TODO?

common/cloud_builder.py
```diff
@@ -0,0 +1,261 @@
...
+    """Uploads a file to Google Cloud Storage."""
+    dest_file_name = os.path.basename(local_file_path)
+    self.bucket.blob(dest_file_name).upload_from_filename(local_file_path)
+    bucket_file_url = f'gs://{self.bucket_name}/{dest_file_name}'
+    logging.info('Uploaded %s to %s', local_file_path, bucket_file_url)
+    return bucket_file_url
+
+  def _prepare_and_upload_archive(self) -> str:
+    """Archives and uploads local OFG repo to cloud build."""
+    ofg_repo = subprocess.check_output(['git', 'rev-parse', '--show-toplevel'],
```
oliverchang, line 73: Instead of shelling out to git for this, can we instead add this to `utils.py`: 

```
OFG_ROOT_DIR = os.path.abspath(os.path.dirname(os.path.dirname(__file__)))
```

(or something similar?) 

This also makes it more robust because it doesn't depend on the CWD.

oliverchang (approved):

common/cloud_builder.py
DonggeLiu: A fresh error:
```python
google.auth.exceptions.RefreshError: ('invalid_scope: Invalid OAuth scope or ID token audience provided.', {'error': 'invalid_scope', 'error_description': 'Invalid OAuth scope or ID token audience provided.'})
```
[Full log](https://paste.googleplex.com/6008336056320000#l=21).

This does not block any experiments because it only happens when requesting cloud builds from a local device, which does not require a refresh.

oliverchang (approved):

DonggeLiu:
OK the report seems good now:
https://llm-exp.oss-fuzz.com/Result-reports/ofg-pr/2024-09-17-580-ai-comparison/index.html

DavidKorczynski:
We have a regression in the cloud runs https://github.com/google/oss-fuzz-gen/pull/620 and https://github.com/google/oss-fuzz-gen/pull/618 -- considering the size of this PR I think maybe it was introduced here?

oliverchang:
+1 seeing:

```
usage: run_all_experiments.py [-h] [-n NUM_SAMPLES] [-t TEMPERATURE]
2024-09-18 09:11:28.322 AEST
[-tr [TEMPERATURE_LIST ...]]
2024-09-18 09:11:28.322 AEST
[-c CLOUD_EXPERIMENT_NAME]
2024-09-18 09:11:28.322 AEST
[-cb CLOUD_EXPERIMENT_BUCKET]
2024-09-18 09:11:28.322 AEST
[-b BENCHMARKS_DIRECTORY] [-y BENCHMARK_YAML]
2024-09-18 09:11:28.322 AEST
[-to RUN_TIMEOUT] [-a [AI_BINARY]] [-l MODEL]
2024-09-18 09:11:28.322 AEST
[-td TEMPLATE_DIRECTORY] [-w WORK_DIR]
2024-09-18 09:11:28.322 AEST
[--context] [-e INTROSPECTOR_ENDPOINT]
2024-09-18 09:11:28.322 AEST
[-lo LOG_LEVEL] [-of OSS_FUZZ_DIR]
2024-09-18 09:11:28.322 AEST
[-g GENERATE_BENCHMARKS]
2024-09-18 09:11:28.322 AEST
[-gp GENERATE_BENCHMARKS_PROJECTS]
2024-09-18 09:11:28.322 AEST
[-gm GENERATE_BENCHMARKS_MAX] [--delay DELAY]
2024-09-18 09:11:28.322 AEST
[-p PROMPT_BUILDER] [-ag]
```

oliverchang:
I think this is the fix: https://github.com/google/oss-fuzz-gen/pull/621

PR #595, opened 2024-09-10
Support dry run

fdt622:
This PR supports the function of dry run fuzz target, that is, only execute the empty seed once. Our insight is that the generated fuzz driver should not crash in any form when executing the empty seed in theory. Our idea is build => dry run with asan => short-term fuzz with asan. Failure of dry run is regarded as run failure, and the premise of entering short-term fuzz is that dry run is successful. The sanitizer used by dry run is ASAN. In addition, we have enabled the two [ASAN](https://github.com/google/sanitizers/wiki/AddressSanitizerFlags) options check_initialization_order and detect_stack_use_after_return (whenever dry run or fuzz). We created a dry_run folder under the logs folder. Currently, all dry run crash would be label with FP_NEAR_INIT_CRASH in _parse_libfuzzer_logs. For the time being, we do not support dry run of CloudBuilderRunner because this will involve modifications to OSS-Fuzz.


oliverchang:
Thanks for the contribution! We've actually found several legitimate bugs in various projects when an empty input is passed. 

e.g. cases where a function does not properly take into account zero length strings might look something like:

```c
void func(char *str) {
  if (str[strlen(str) - 1] == '\n') {
     ...
   }
}
```

Is there some better way to filter out fuzz targets in a way that accounts for this?

fdt622:
> Thanks for the contribution! We've actually found several legitimate bugs in various projects when an empty input is passed.
> 
> e.g. cases where a function does not properly take into account zero length strings might look something like:
> 
> ```c
> void func(char *str) {
>   if (str[strlen(str) - 1] == '\n') {
>      ...
>    }
> }
> ```
> 
> Is there some better way to filter out fuzz targets in a way that accounts for this?

Thank you for your review. After discussing with @DonggeLiu , we think that the result of "executing empty seed (dry run)" can be used as feedback information to the agent to assist it in generating a better driver, rather than as a criterion for judging whether there is a crash or a problem with the driver. I will add the dry run part after the crash analyzer agent. The crash analyzer agent is being developed intensively and a PR will be submitted soon.

fdt622:
I will temporarily close this pr.

PR #602, opened 2024-09-12
introspector: Fix bug in introspector response data retrieval

arthurscchan:
This PR fixes a bug in the `_get_data` function used for retrieving the necessary data from an introspector response. It was found that some introspector queries could return an empty list, empty dictionary, or boolean False as data. This causes the current check to be unable to distinguish between a missing field in the response, an empty list, or a boolean False value. This PR addresses the issue.

### DIFF
--- a/data_prep/introspector.py
+++ b/data_prep/introspector.py
@@ -209,8 +209,10 @@ def _get_data(resp: Optional[requests.Response], key: str,
         resp.content.decode('utf-8').strip())
     return default_value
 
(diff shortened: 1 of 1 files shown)


arthurscchan:
/gcbrun exp -n ascc-jvm-pr-602-1 -m vertex_ai_gemini-1-5 -i -b jvm-small

DavidKorczynski (approved):

PR #607, opened 2024-09-13
Agent enhancement 4

DonggeLiu:
1. Show more info in the report.
2. Save more files during the experiment.
3. Show more logs during the experiment.

### DIFF
--- a/experiment/workdir.py
+++ b/experiment/workdir.py
@@ -39,6 +39,7 @@ class WorkDirs:
     os.makedirs(self.build_logs, exist_ok=True)
     os.makedirs(self.run_logs, exist_ok=True)
     os.makedirs(self._corpus_base, exist_ok=True)
+    os.makedirs(self.pickles, exist_ok=True)
 
   @property
   def base(self):
@@ -79,6 +80,10 @@ class WorkDirs:
   @property
   def build_logs(self):
     return os.path.join(self._base_dir, 'logs', 'build')
+
+  @property
+  def pickles(self):
+    return os.path.join(self._base_dir, 'pickles')
 
   @property
   def run_logs(self):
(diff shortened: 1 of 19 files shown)


mihaimaruseac (approved):

experiment/workdir.py
```diff
@@ -40,6 +40,7 @@ def __init__(self, base_dir, keep: bool = False):
     os.makedirs(self.run_logs, exist_ok=True)
     os.makedirs(self._corpus_base, exist_ok=True)
     os.makedirs(self.dills, exist_ok=True)
+    os.makedirs(self.fuzz_target, exist_ok=True)
```
oliverchang, line 43: what's the difference between fuzz_target and raw_targets/fixed_targets? 

If we want to deprecate/replace the older ones, can you please add a comment?

report/common.py
```diff
@@ -340,6 +338,28 @@ def get_triage(self, benchmark: str, sample: str) -> Triage:
...
+        fuzz_target_code = f.read()
+
+    build_script_code = ''
+    if f'{trial:02s}.build_script' in files:
+      build_script_path = os.path.join(fuzz_target_dir,
+                                       f'{trial:02s}.build_script')
+      with FileSystem(build_script_path).open() as f:
+        build_script_code = f.read()
+
+    return [Target(code=fuzz_target_code, fixer_prompt=build_script_code)]
```
oliverchang, line 361: why does this have to be a list? can we just return a single pair? and have the caller construct any lists if necessary?

report/web.py
```diff
@@ -148,7 +148,9 @@ def generate(self):
       self._write_benchmark_crash(benchmark, samples)
 
       for sample in samples:
-        sample_targets = self._results.get_targets(benchmark.id, sample.id)
+        sample_targets = (self._results.get_targets(benchmark.id, sample.id) or
```
oliverchang, line 151: it's a bit confusing we have two very similar but different functions here. do we intend to get rid of the old one?

DonggeLiu:
OK the report seems good now:
https://llm-exp.oss-fuzz.com/Result-reports/ofg-pr/2024-09-17-580-ai-comparison/index.html

DonggeLiu:
Thanks for reviewing this.
This is not ready yet, let me close this for now to avoid confusion.
I will address the comments and create another more organized PR.

Issue #525, opened 2024-08-02
More robust and dynamic way to obtain fuzz target info


DonggeLiu:
1. Improve the existing logic of getting fuzz target binary name and source file path to update benchmark yaml.
2. Dynamically re-generate fuzz target info at runtime.
3. If dynamic generation fails, use the default one in benchmark yaml.
4. Record which binary and source file is used during the experiment to avoid confusion in debugging.

DavidKorczynski:
Ref: https://github.com/google/oss-fuzz-gen/pull/524#issuecomment-2264668798

DavidKorczynski:
I think we can also use FI data more here to increase reliability

oliverchang:
@DavidKorczynski (or @arthurscchan ), can you help us remove the hardcoded fuzz target binary names from our benchmark YAML configs? This is a recurring source of regressions as targets change names. 

e.g. https://llm-exp.oss-fuzz.com/Result-reports/scheduled/2024-08-31-weekly-all/sample/output-tinyxml2-_zn8tinyxml210xmlelement12setattributeepkcs2_/01.html

PR #524, opened 2024-08-02
Fix fuzz target binary name and source file path

DonggeLiu:
Some projects had [>0 build rate but 0 coverage](https://docs.google.com/spreadsheets/d/1bx1YeKc7cqJr9Qlt_Qrw06EV84p5OTbFQqbfVBPLKO4/edit?resourcekey=0-qvVvym2ZIUNIFn3jp-osLw&gid=1522358986#gid=1522358986) on all benchmarks, because of their incorrect fuzz target binary name (and/or fuzz target path).

This PR fixes that.

PR #611, opened 2024-09-14
introspector: use introspector to get harness/source pairs by default

DavidKorczynski:
Ref: https://github.com/google/oss-fuzz-gen/issues/525


DavidKorczynski:
This performs a lot better than our existing automated setup, but there are still three discrepancies between the manually generated ones from https://github.com/google/oss-fuzz-gen/pull/524:

```
"project": "librawspeed"
"target_name": "/src/librawspeed/fuzz/librawspeed/codes/PrefixCodeDecoder/Dual.cpp"
"target_path": "/src/librawspeed/fuzz/librawspeed/codes/PrefixCodeDecoder/Dual.cpp"


"target_name": "fuzztest"
"target_name": "fuzztest_proto2_static"
"target_path": "/src/nanopb/tests/fuzztest/fuzztest.c"


"project": "proftpd"
"target_name": "fuzzer"
"target_path": "/src/fuzzer.c"
```

These will be dealt with on the introspector side -- but this still performs a lot better than the current status quo -- this is good to go.

Finally, there are some benign discrepancies to https://github.com/google/oss-fuzz-gen/pull/524 in that sometime the current heuristic will select other harnesses than the manually selected ones. In general, this is no problem. However, there are some potential cases where we may want to select a specific harness within a given project if the given harness has a wider (or different) linkage. An example here is abseil-cpp where the better harness to taget is `string_utilities_fuzzer` as it includes more sources from the source tree: https://github.com/google/oss-fuzz/blob/master/projects/abseil-cpp/BUILD

AdamKorcz (approved):

PR #621, opened 2024-09-17
Don't quote `$AGENT_ARG` in docker_run.sh

oliverchang:
Otherwise we can pass an empty string which breaks the argument passing.


DonggeLiu (commented):
LGTM

PR #623, opened 2024-09-18
Agent enhancement 4

DonggeLiu:
Add execution stage


stage/execution_stage.py
```diff
@@ -0,0 +1,80 @@
...
+from experiment.evaluator import Evaluator
+from results import Result, RunResult
+from stage.base_stage import BaseStage
+
+
+class ExecutionStage(BaseStage):
+  """Executes fuzz targets and build scripts. This stage takes a fuzz target
+  and its build script, runs them locally or on the cloud with OSS-Fuzz infra,
+  and outputs code coverage report and run-time crash information for later
+  stages to analyze and improve on. It OSS-Fuzz infra to perform these tasks."""
```
oliverchang, line 18: It **uses** OSS-Fuzz infra?

oliverchang (approved):

stage/execution_stage.py
DonggeLiu: Yep, let me fix that.

DonggeLiu:
This PR is not ready for review yet (many minor things to fix, e.g., a better design of `RunResult`) but should be ready for a preliminary exp : )
The report from a local exp already shows `Program Counter Coverage` correctly:
image

The 0 build rate is incorrect due to current data structure of `RunResult`, I will fix it.

DonggeLiu:
The report looks better now:
https://llm-exp.oss-fuzz.com/Result-reports/ofg-pr/2024-09-20-623-dg6-comparison/index.html

results.py
```diff
@@ -48,41 +51,87 @@ def __init__(self,
...
-        'compile_log': self.full_log,
-        'compile_insight': self.insight,
+        'compiles': self.compiles,
+        'compile_error': self.compile_error,
+        'compile_log': self.compile_log,
     }
 
 
-class RunResult(Result):
+class RunResult(BuildResult):
```
DonggeLiu, line 75: Would it be better if we use a single `Result` class that contains all attributes, instead of multiple subclasses like this?

DonggeLiu:
Hmm, Build status has some false negatives:
1. [The report](https://llm-exp.oss-fuzz.com/Result-reports/ofg-pr/2024-09-20-623-dg6-comparison/sample/output-abseil-cpp-_zzn4absl19str_format_internal12_global__n_122formatfnegativeexpslowens_7uint128eirkns1_11formatstatee/03.html) says this fuzz target failed.
2. But its [original cloud build](https://pantheon.corp.google.com/cloud-build/builds;region=us-central1/d6ffd145-329e-4ccf-b33b-8056f6530dc2;step=5?e=-13802955&mods=logs_tg_prod&project=oss-fuzz) says the same fuzz target succeeded.

utils.py
```diff
@@ -29,3 +32,65 @@ def deserialize_from_dill(dill_path: Any) -> Any:
...
+            delay = delay_fn(exc_name, retry_count)
+          else:
+            delay = random.uniform(0, 60)
+          time.sleep(delay)
+
+          if retry_count >= current_retries:
+            logging.error('Max retries reached for %s. Exiting.', exc_name)
+            break
+        except Exception as e:
+          logging.error('Unhandled exception: %s. Exiting.', e)
```
oliverchang, line 86: shouldn't we re-raise the exception? Or just not catch these? 

It seems dangerous to let things proceed if an exception happens that we don't expect.

utils.py
```diff
@@ -29,3 +32,65 @@ def deserialize_from_dill(dill_path: Any) -> Any:
...
+          break
+
+      # Final return after all retries failed or unhandled exception
+      logging.warning('All retries failed for %s. Returning default value.',
+                      func.__name__)
+      return default_return
```
oliverchang, line 92: same comment here: it seems pretty dangerous to have a pattern to let things fail silently and return a default value. do we need it ?

utils.py
```diff
@@ -29,3 +32,65 @@ def deserialize_from_dill(dill_path: Any) -> Any:
...
+          exc_name = type(e).__name__
+          retry_count += 1
+          logging.info(
+              'Retrying %s due to %s. Attempt %d with args=%s, kwargs=%s',
+              func.__name__, exc_name, retry_count, args, kwargs)
+
+          # Get the number of retries for this exception, or use the default
+          current_retries = exception_config.get(type(e), default_retries)
+          if delay_fn:
+            delay = delay_fn(exc_name, retry_count)
```
oliverchang, line 77: Probably better to pass in the exception object rather than just the name?

utils.py
```diff
@@ -29,3 +32,65 @@ def deserialize_from_dill(dill_path: Any) -> Any:
...
+          else:
+            delay = random.uniform(0, 60)
+          time.sleep(delay)
+
+          if retry_count >= current_retries:
+            logging.error('Max retries reached for %s. Exiting.', exc_name)
+            break
```
oliverchang, line 84: re-raise?

utils.py
```diff
@@ -29,3 +32,65 @@ def deserialize_from_dill(dill_path: Any) -> Any:
...
+    Decorator that retries the function on specified exceptions.
+    :param exceptions: List/Set of exceptions or a dictionary of exceptions with
+      custom retry counts.
+    :param default_retries: Number of retries if no custom count is provided.
+    :param delay_fn: Function to determine the delay between retries. Default
+      is random between 0-60 seconds.
+    :param default_return: The value returned if all retries fail.
+    """
+  exception_config = {
+      exc: default_retries for exc in exceptions or {}
```
oliverchang, line 52: Do we need the `dict` support here to support different retry configs per exception? Would be good to follow YAGNI here and keep things simple until we actually need it.

pipeline.py
```diff
@@ -49,6 +49,14 @@ def _execute_one_cycle(self, result_history: list[Result],
                      result_history[-1])
     result_history.append(
         self.writing_stage.execute(result_history=result_history))
+    if (not isinstance(result_history[-1], BuildResult) or
+        not result_history[-1].compiles):
```
oliverchang, line 53: Why single out "compiles" ? That seems to be tying specific stage logic too much into this Pipeline code. 

Can we instead generalize this some other way to avoid encoding low level logic here? i.e. `result_history[-1].success`. ?

report/common.py
```diff
@@ -358,6 +364,28 @@ def get_targets(self, benchmark: str, sample: str) -> list[Target]:
...
+        fuzz_target_code = f.read()
+
+    build_script_code = ''
+    if f'{trial:02s}.build_script' in files:
+      build_script_path = os.path.join(fuzz_target_dir,
+                                       f'{trial:02s}.build_script')
+      with FileSystem(build_script_path).open() as f:
+        build_script_code = f.read()
+
+    return Target(code=fuzz_target_code, fixer_prompt=build_script_code)
```
oliverchang, line 387: why `fixer_prompt` ?

results.py
```diff
@@ -38,6 +39,8 @@ def to_dict(self) -> dict:
         'trial': self.trial,
         'fuzz_target_source': self.fuzz_target_source,
         'build_script_source': self.build_script_source,
+        'author': str(self.author),
+        'agent_dialogs': self.agent_dialogs,
```
oliverchang, line 43: how about just chat_history ?

results.py
```diff
@@ -48,41 +51,87 @@ def __init__(self,
...
                build_script_source: str = '',
                author: Any = None,
                agent_dialogs: Optional[dict] = None) -> None:
     super().__init__(benchmark, trial, work_dirs, fuzz_target_source,
                      build_script_source, author, agent_dialogs)
-    self.status: bool = status  # Build success/failure.
-    self.error: str = error  # Build error message.
-    self.full_log: str = full_log  # Build full output.
-    self.insight: str = insight  # Reason and fixes for build failure.
+    self.compiles: bool = compiles  # Build success/failure.
```
oliverchang, line 63: do we need the type annotations on these lines?

results.py
oliverchang: yes possibly, but this huge flat class is getting a bit too large and difficult to understand. Would it make sense to have a single `Result` that contains substructures ?

ci/k8s/pr-exp.yaml
```diff
@@ -41,6 +41,8 @@ spec:
         - mountPath: "/experiment/results"
           name: results-volume
         env:
+        - name: GCB_BUILDPOOL_NAME
```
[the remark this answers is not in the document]
DonggeLiu: Yep, I will remove this when no more experiment is required.

DonggeLiu:
Report looking good:
https://llm-exp.oss-fuzz.com/Result-reports/ofg-pr/2024-09-20-623-dg8-comparison/index.html

mihaimaruseac (approved):

experiment/evaluator.py
```diff
@@ -256,6 +259,20 @@ def create_ossfuzz_project(self, name: str, target_file: str) -> str:
       f.write(f'\nCOPY {os.path.basename(target_file)} '
               f'{self.benchmark.target_path}\n')
 
+    if os.path.getsize(build_script_path) == 0:
+      return name
+
+    logger.info('Build script: %s', build_script_path)
```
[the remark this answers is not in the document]
DonggeLiu: No, but let me keep it a bit longer for debugging purpose.
I will remove it before merging.

utils.py
DonggeLiu: I added this mainly for errors related to exceeding the quota limit.
We can retry them a large number of times (e.g., 100), while only retry others (e.g., the VertexAI internal API error we saw earlier) for max 5 times.

report/common.py
DonggeLiu: It's a temporary hacky way to show the new build script in the report.
I plan to organize it, but maybe better in a separate PR?

results.py
DonggeLiu: Yes! I already named that step, will find a way to extract it later.

results.py
DonggeLiu: I'm not sure if I understand this correctly, but I think these annotations will be helpful to the linters when the variables are used elsewhere?
For local variables, they are not necessary.

results.py
DonggeLiu: Yep, this is a very good suggestion, I can have:
```python
class Result:
  build_status = BuildStatus()
  run_status = RunStatus()
  ...
```

Let me do it in the next PR, because this will involve refactoring a lot of code that are not directly related to this PR (e.g., the writing stage, functions to save results)

DonggeLiu:
TODO:
1. https://github.com/google/oss-fuzz-gen/pull/623#discussion_r1768072147
2. https://github.com/google/oss-fuzz-gen/pull/623#discussion_r1772385856
3. https://github.com/google/oss-fuzz-gen/pull/623#discussion_r1772453454

report/common.py
oliverchang: Is it a lot of work to do that ? 

If so, please add TODO with an explanation here.

mihaimaruseac (approved):

mihaimaruseac (approved):

DonggeLiu:
The last commit should fix the "0-coverage" in report.

I started to see more invalid fuzz targets, e.g., did not call function under test. I will create a new PR to add the previous checker into prototyper.

As for the crashes caused by semantic errors, they should be fixed in the next iteration (e.g., enhancer) in the current design.

DonggeLiu:
Will do [this](https://github.com/google/oss-fuzz-gen/pull/623#issuecomment-2370057440) once the exp reports confirms the '0-coverage' issue is resolved.

mihaimaruseac (approved):

ci/k8s/pr-exp.yaml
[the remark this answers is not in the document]
oliverchang: Please remember to remove this before merging

common/cloud_builder.py
```diff
@@ -210,14 +214,29 @@ def _cancel_build(self, build_id: str) -> None:
     """Cancel a GCB build"""
     self.builds.cancel(projectId=self.project_id, id=build_id).execute()
 
+  def _extract_chat_history(self, full_log: str) -> str:
+    """Extracts the agent chat history from cloud build log."""
+    in_chat = False
+    pattern = r'^Step\s+#(\d+)\s+-\s+"agent-step":\s+'
```
oliverchang, line 220: nit: can you put this in `_CHAT_HISTORY_PATTERN = re.compile(...)` somewhere?

results.py
oliverchang: Sorry I meant: would it be better to define these at a class level?

i.e.

```python

class BuildResult(Result):
  compiles: bool 
  compile_error: str
```

etc. IMO it's a bit easier to find this way.

oliverchang (approved):

common/cloud_builder.py
DonggeLiu: Done, thanks

results.py
DonggeLiu: Like this?
https://github.com/google/oss-fuzz-gen/pull/623/commits/c00fbdbaf48008cbb68110bd17aa0eb9547ab3aa

ci/k8s/pr-exp.yaml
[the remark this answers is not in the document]
DonggeLiu: Yep, done!

PR #629, opened 2024-09-18
Add another trophy from today.

oliverchang:


DonggeLiu (commented):
Wow, so many findings recently!

PR #634, opened 2024-09-20
Use different GCB pool for large PR experiments

oliverchang:


DonggeLiu (approved):

DonggeLiu:
IIUC, we also need these commits to use the customized pool:
https://github.com/google/oss-fuzz-gen/pull/623/commits/a508021934810f9716157546b59f0f0f2bf6e7c8
https://github.com/google/oss-fuzz-gen/pull/623/commits/934ba2af62ffea15c011bf952932ecf78f6844be

oliverchang:
> IIUC, we also need these commits to use the customized pool: [a508021](https://github.com/google/oss-fuzz-gen/commit/a508021934810f9716157546b59f0f0f2bf6e7c8) [934ba2a](https://github.com/google/oss-fuzz-gen/commit/934ba2af62ffea15c011bf952932ecf78f6844be)
> 
> We can merge them in #623

Right, that's only for the agent workflow right? Let's merge it in a separate PR as you mention.

PR #645, opened 2024-09-24
run_all_experiments: fix coverage sync and wrong aggregate coverage gains measure

DavidKorczynski:
- Adjusts coverage gain to show the actual coverage gain with respect to the existing coverage report. Previously the aggregated data simply showed the total coverage achieved.
- Makes the reading independent of the results array. The goal here is to make it asynchronously since we experienced issues where some benchmark runs would finish but not show up in the aggregate table.

### DIFF
--- a/run_all_experiments.py
+++ b/run_all_experiments.py
@@ -132,8 +132,7 @@ def prepare_experiment_targets(
   return experiment_configs
 
 
-def run_experiments(benchmark: benchmarklib.Benchmark,
-                    args: argparse.Namespace) -> Result:
+def run_experiments(benchmark: benchmarklib.Benchmark, args) -> Result:
   """Runs an experiment based on the |benchmark| config."""
   try:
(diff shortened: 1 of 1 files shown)


DavidKorczynski:
/gcbrun exp -n d-cov-40 -b comparison -ns 3 -rd

DavidKorczynski:
/gcbrun exp -n d-cov-41 -b comparison -ns 3 -rd

DavidKorczynski:
Experiment is looking good: https://llm-exp.oss-fuzz.com/Result-reports/ofg-pr/2024-09-25-645-d-cov-41-comparison/index.html

There is a minor discrepancy because the textcov in the comparison includes some of the harness source code, which makes the number slightly off. I'll fix that in a later PR as we need the benchmark to be around in the workdir -- which would have some separate benefits in and of itself in the form of making the `results` folder self-contained.

AdamKorcz (approved):

PR #650, opened 2024-09-25
save benchmarks in results and use for aggregate coverage calculation

DavidKorczynski:
This fixes the discrepancy mentioned here https://github.com/google/oss-fuzz-gen/pull/645#issuecomment-2373714631. Saving the benchmark for each result has additional benefits than supporting the aggregate. It makes it easier to save results when auto-generation of benchmarks are used as part of the experiment (I use this a lot).

Notice not the full benchmark `.yaml` from `benchmarks-set/...` is set, but only the actual (one function or one test file) is saved per result dir.

### DIFF
--- a/data_prep/introspector.py
+++ b/data_prep/introspector.py
@@ -1049,7 +1049,7 @@ if __name__ == '__main__':
                                                       args.max_functions,
                                                       args.target_oracle)
   if benchmarks:
-    benchmarklib.Benchmark.to_yaml(benchmarks, args.out)
+    benchmarklib.Benchmark.to_yaml(benchmarks, outdir=args.out)
   else:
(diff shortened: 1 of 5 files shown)


DavidKorczynski:
/gcbrun exp -n d-cov-44 -b comparison -ns 4 -rd

DavidKorczynski:
/gcbrun exp -n d-cov-45 -b comparison -ns 4 -rd

DavidKorczynski:
Experiment is looking good, although there are some minor discrepancies still, but will look at those further down the line: https://llm-exp.oss-fuzz.com/Result-reports/ofg-pr/2024-09-26-650-d-cov-45-comparison/index.html

AdamKorcz (approved):

PR #661, opened 2024-10-02
Prototyper improvements

### LINKED ISSUE #658: Detect if the fuzz-under-test is invoked by the fuzz target
Instrument compiler to tell us if a target is actually calling the target function:
Jonathan: [wrap function](https://cppcodetips.wordpress.com/2020/08/18/mocking-a-function-in-c-using-wl-wrap-compiler-flag-in-gcc/) to get it to print something, like [this](https://github.com/google/fuzzbench/pull/838/files).
* Mihai: A wrapper calls the function-under test, link wrapper during linking.

DonggeLiu:
## Functionality
1. Validate if the function-under-test is referenced by the fuzz target (fixes #658)
2. Use cached image from `chronos`.
3. Use exact exception matching in decorator `@retryable` to ensure the correct config (e.g., #retries) is applied.
4. Compile the new fuzz target with the old build script once even if LLM proposes a new build script, in case the new one is wrong.
5. Show chat history of each trail for simpler debugging.
6. Show exception (e.g., cloud build expiry) in report.


DonggeLiu:
The reports includes the chat history, which shows the reference-check is in place:

https://llm-exp.oss-fuzz.com/Result-reports/ofg-pr/2024-10-02-661-dg-comparison/index.html

DonggeLiu:
https://llm-exp.oss-fuzz.com/Result-reports/ofg-pr/2024-10-14-661-dg-comparison/index.html

DonggeLiu:
21 successes:
https://llm-exp.oss-fuzz.com/Result-reports/ofg-pr/2024-10-16-661-dg-comparison/index.html

tool/container_tool.py
```diff
@@ -34,23 +34,30 @@ def _execute_command(self,
                        command: list[str],
                        in_container: bool = False) -> sp.CompletedProcess:
```
oliverchang, line 35: nit: we should just have two separate functions for the "in container" vs not case. It's a bit too confusing to have one function that handles both via a boolean flag.

agent/prototyper.py
```diff
@@ -93,11 +114,24 @@ def _validate_fuzz_target_and_build_script(self, cur_round: int,
     binary_exists = ls_result.returncode == 0
     logger.debug('ROUND %02d Final fuzz target binary exists: %s', cur_round,
                  binary_exists)
-    compilation_tool.terminate()
 
+    # Validate if function-under-test is referenced by the fuzz target.
+    disassemble_result = compilation_tool.execute(
```
oliverchang, line 119: can we move this into its own function?

common/cloud_builder.py
```diff
@@ -23,8 +24,10 @@
 
 OF_REPO = 'https://github.com/google/oss-fuzz.git'
 OFG_ROOT_DIR = os.path.abspath(os.path.dirname(os.path.dirname(__file__)))
-US_CENTRAL_CLIENT_OPTIONS = google.api_core.client_options.ClientOptions(
-    api_endpoint='https://us-central1-cloudbuild.googleapis.com/')
+REGION = random.choice(
+    os.getenv('CLOUD_BUILD_LOCATIONS', 'us-west2').split(','))
```
oliverchang, line 28: singular CLOUD_BUILD_LOCATION?

llm_toolkit/models.py
```diff
@@ -607,24 +616,82 @@ class GeminiV1D5(GeminiModel):
...
+  def truncate_prompt(self,
+                      raw_prompt_text: Any,
+                      extra_text: Any = None) -> Any:
+    """Truncates the prompt text to fit in MAX_INPUT_TOKEN."""
+    raw_prompt_text = raw_prompt_text or ' '
+
+    # A rough estimation in case count_tokens failed on very large prompt.
+    # Gemini suggest 4 char per token, using 3 here to be safer. Normally,
+    # prompts should not be this large anyway.
+    if len(raw_prompt_text) > 3 * self.MAX_INPUT_TOKEN:
```
oliverchang, line 659: Can we use the existing `estimate_token_num` instead for all of these token calculations? We can modify it if needed if it's too slow on large text. e.g. fallback to a simpler calculation if len is greater than some threshold.

llm_toolkit/models.py
```diff
@@ -607,24 +616,82 @@ class GeminiV1D5(GeminiModel):
...
+    try:
+      original_token_count = self.get_model().count_tokens(
+          raw_prompt_text).total_tokens
+      if extra_text:
+        extra_token_count = self.get_model().count_tokens(
+            extra_text).total_tokens
+      else:
+        extra_token_count = 0
+      token_count = original_token_count
+    except Exception as e:
```
oliverchang, line 671: we should avoid this error handling if we can. what exception do we see here? If we absolutely have to, let's have this all be handled inside `estimate_token_num` ?

experiment/oss_fuzz_checkout.py
```diff
@@ -375,23 +375,58 @@ def prepare_build(project_name, sanitizer, generated_project):
...
+                  image_name)
+      return True
   except sp.CalledProcessError:
-    logger.info('Unable to list all docker images')
+    logger.warning('Unable to use local cached image of %s: %s', project_name,
+                   image_name)
+  return False
+
+
+def _image_exists_online(image_name: str, project_name: str) -> bool:
```
oliverchang, line 395: Can you re-use existing functionality in this file? see e.g. https://github.com/google/oss-fuzz-gen/blob/0d80846b50a5d2dbb98e9dc154edb5ab34d06da5/experiment/oss_fuzz_checkout.py#L189C5-L189C34

oliverchang (changes requested):

experiment/oss_fuzz_checkout.py
DonggeLiu: Done. I also thought about reusing other functions, but they appear to be coupled with cached build scripts etc.
I want to avoid adding params to them for my case, but please let me know if you think that's better.

mihaimaruseac (approved):

llm_toolkit/models.py
DonggeLiu: I've replaced this code with `estimate_token_num()` for simplicity.
The `count_tokens()` API has some drawbacks, (e.g., unavailable in `us-west2`) and we do not need the exact token number in this case, particularly given we have largely reduce max context token size.

llm_toolkit/models.py
DonggeLiu: done.

common/cloud_builder.py
DonggeLiu: Done.

agent/prototyper.py
DonggeLiu: Yep, done

tool/container_tool.py
DonggeLiu: Sure, done

DonggeLiu:
24 successes:
https://llm-exp.oss-fuzz.com/Result-reports/ofg-pr/2024-10-17-661-dg1-comparison/index.html

llm_toolkit/models.py
```diff
@@ -56,6 +56,8 @@ class LLM:
   # TODO(mihaimaruseac): Should this be MAX_TOKENS or a different global?
   context_window: int = 2000  # Default token size.
 
+  MAX_INPUT_TOKEN: int = 0
```
oliverchang, line 59: should this be infinity by default? to avoid surprising behaviour otherwise. It seems to me that the default should be to have no limit, rather than limit to 0?

agent/prototyper.py
```diff
@@ -43,26 +43,68 @@ def _update_fuzz_target_and_build_script(self, cur_round: int, response: str,
...
+    self._validate_fuzz_target_and_build_script_via_recompile(
+        cur_round, build_result)
+
+    if not build_result.success and build_script_source:
+      logger.info('Then compile fuzz target with modified build script.')
+      build_result.build_script_source = build_script_source
+      self._validate_fuzz_target_and_build_script_via_recompile(
+          cur_round, build_result, use_recompile=False)
+
+  def _vadlidate_fuzz_target_references_function(
```
oliverchang, line 85: nit: fix spelling of validate.

agent/prototyper.py
```diff
@@ -109,18 +155,34 @@ def _container_handle_conclusion(
...
+          f'{build_result.fuzz_target_source}\n</fuzz target>\n'
+          f'<build script>\n{build_result.build_script_source}\n'
+          f'</build script>\n<compilation log>\n{compile_log}\n'
+          '</compilation log>\n')
+    elif not build_result.is_function_referenced:
+      logger.info(
+          '***** Fuzz target does not reference function-under-test in %02d '
+          'rounds *****', cur_round)
+      prompt_text = (
+          'The fuzz target builds succeessfully, but the target function '
```
oliverchang, line 178: nit: fix spelling of successfully.

oliverchang (approved):

llm_toolkit/models.py
DonggeLiu: Thanks, using `sys.maxsize` now.

agent/prototyper.py
DonggeLiu: Done, thanks

agent/prototyper.py
DonggeLiu: Done.

DonggeLiu:
Final report looking good:
https://llm-exp.oss-fuzz.com/Result-reports/ofg-pr/2024-10-21-661-dg-comparison/index.html

Merging now.

PR #664, opened 2024-10-03
Use fuzz target binary and source file from FI APIs

DonggeLiu:
So that we can use the cached images from OSS-Fuzz's `chronos`.
TODO:
Update other sets (e.g., `comparison`).


DonggeLiu:
Report:
https://llm-exp.oss-fuzz.com/Result-reports/ofg-pr/2024-10-03-664-dg-comparison/index.html

DonggeLiu:
[Dynamically collecting fuzz target binary & source file info](https://github.com/google/oss-fuzz-gen/issues/525#issuecomment-2325616151) will need to re-struct `introspectory.py` to avoid circular-dependency.
I will leave it for next time.

oliverchang (approved):
LGTM, but I wonder if we can just remove these hardcoded values completely from the yaml files by default? 

i.e. always call FI dynamically to get these if they don't exist in the yaml? There may still be some edge cases where we do need to hardcode them for certain projects, and  we can still prioritize these ones when they exist.

DonggeLiu:
> LGTM, but I wonder if we can just remove these hardcoded values completely from the yaml files by default?
> 
> i.e. always call FI dynamically to get these if they don't exist in the yaml? There may still be some edge cases where we do need to hardcode them for certain projects, and we can still prioritize these ones when they exist.

Yep, good idea. That will address [this issue](https://github.com/google/oss-fuzz-gen/issues/525#issuecomment-2325616151).
Would it be OK to leave that to another PR later?
I wanted to quickly merge this so that we can maybe use #661 in this week's weekly exp.

PR #670, opened 2024-10-22
prompts: extend test-to-harness prompting

DavidKorczynski:
- Fixes some issues there were with header file inclusion in the prompt (it used the wrong variable, so the string was wrong)
- adds a sample harness from the actual repository
- adds a list of all headers


DavidKorczynski:
This has been tested extensive locally, but will not run an experiment here since the OSS-Fuzz build status seems to be a bit off with introspector

AdamKorcz (approved):

Issue #676, opened 2024-10-29
Improve report output for agent flow


oliverchang:
Right now, our logs show a flattened interaction log between LLM and our prompts. e.g. https://llm-exp.oss-fuzz.com/Result-reports/ofg-pr/2024-10-21-661-dg-comparison/sample/output-hoextdown-hoedown_document_render_inline/01.html

We should format this a bit better, in a way that let's us format the HTML report to clearly see the prompts we are sending vs what the LLM is returning.

PR #677, opened 2024-10-29
Better format agent chat history in reports.

oliverchang:
To make the LLM interactions clearer. Prompts are coloured with a slight pink background, and responses are moved to the right side with a slight green background. 

Part of #676.

Also stop propagating chat logging to the root logger (to avoid duplicate output). 

This PR also adds some workarounds for when `fixed_targets`, `raw_targets` don't exist in the results dir for agent runs. This can happen when we download a results dir from GCS, because empty dirs are not uploaded. 

Example: 

image

### DIFF
--- a/agent/base_agent.py
+++ b/agent/base_agent.py
@@ -5,7 +5,7 @@ import re
 import subprocess as sp
 import time
 from abc import ABC, abstractmethod
-from typing import Optional
+from typing import Any, Optional
 
 import logger
 import utils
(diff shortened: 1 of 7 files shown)


oliverchang:
Agent report: https://llm-exp.oss-fuzz.com/Result-reports/ofg-pr/2024-10-30-677-oc-comparison/index.html

DonggeLiu (approved):
Thanks soo much!

PR #694, opened 2024-11-06
oss_fuzz_checkout: support local ccache builds

DavidKorczynski:
This works but once we support the replayable scripts we need further adjustments.


DavidKorczynski:
We could potentially reduce more logic in the current cache rewriting logic, however, I'll leave it there until we have the replayable scripts integrated.

PR #699, opened 2024-11-07
Fix Dockerfile rewrite for using cached builds.

oliverchang:
We weren't keeping the ARG, WORKDIR lines.

ARG is necessary for obvious reasons since we are relying on injecting
the base image at runtime.

WORKDIR is required for: https://github.com/google/oss-fuzz/blob/edfa6730fa9f1dca77dd27c0b3a643171e082950/infra/build/functions/build_project.py#L229

### DIFF
--- a/experiment/oss_fuzz_checkout.py
+++ b/experiment/oss_fuzz_checkout.py
@@ -68,7 +68,7 @@ def _set_temp_oss_fuzz_repo():
 def _clone_oss_fuzz_repo():
   """Clones OSS-Fuzz to |OSS_FUZZ_DIR|."""
   clone_command = [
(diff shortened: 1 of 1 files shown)


DonggeLiu (approved):

DonggeLiu:
report:
https://llm-exp.oss-fuzz.com/Result-reports/ofg-pr/2024-11-07-699-ochang-test-comparison/index.html

oliverchang:
I think this works, but there needs to be another change to the oss-fuzz build infra side to use the cached coverage image as well. I'll do that in a followup.

DavidKorczynski (commented):
LGTM

PR #704, opened 2024-11-07
Build image in cloudbuild

DonggeLiu:
This allows `/gcbrun exp` to build and push images before requesting experiments so that maintainers can run PR experiments too.


ci/request_pr_exp.py
```diff
@@ -294,6 +294,36 @@ def _get_gke_credential(args: argparse.Namespace):
     logging.error('Failed to authenticate gcloud: %s', e)
 
 
+def _build_image(args: argparse.Namespace):
```
oliverchang, line 297: I think we can just do this in https://github.com/google/oss-fuzz-gen/blob/main/ci/cloudbuild.yaml pretty easily. 

see e.g. https://github.com/google/oss-fuzz/blob/7f138b90225a76beb5ee47ebfc5cf12d624ca47e/infra/experimental/chronos/cloudbuild.yaml#L30

DonggeLiu: Yep I did that in the first two commits.
Later commits moved these steps to the python script because I am unsure if we want to build an image for every commit/PR:
1. Some commits pushed to PR are not yet ready for experiments. Having those steps in `cloudbuild.yaml` automatically [builds images](https://github.com/google/oss-fuzz-gen/runs/32642001417) for them and wastes computation power.
2. Some PRs from attackers may contain security threats. It might not be a good idea to automatically build and push the image and show the output in CI page.

Moving the steps to `ci/request_pr_exp.py` ensures an image won't be built unless a maintainer+ role manually executes the `gcbrun` command.

oliverchang: `cloudbuild.yaml` only runs with /gcbrun right? So this should never be a concern?

oliverchang: i.e. we just move the docker building into `cloudbuild.yaml`, and then the image only gets build prior to the epxeriment running if someone with privileges runs `/gcbrun`.

ci/cloudbuild.yaml
```diff
@@ -1,4 +1,21 @@
...
+  args:
+  - build
+  - '-t'
+  - us-central1-docker.pkg.dev/oss-fuzz-base/testing/oss-fuzz-gen-pull-request:pr-${_PR_NUMBER}
+  - .
+- name: 'gcr.io/google.com/cloudsdktool/cloud-sdk'
+  entrypoint: 'gcloud'
+  args:
+    - auth
+    - configure-docker
```
oliverchang, line 12: is this required? cloud build should have all the creds set up correctly.

DonggeLiu:
Apparently `oss-fuzz-gen-exp` won't be automatically triggered in other PRs even with the same commits.
See https://github.com/google/oss-fuzz-gen/pull/708

oliverchang (approved):

PR #692, opened 2024-11-06
Large experiment.

oliverchang:
With all oracles and 6 targets each per project.

PR #707, opened 2024-11-08
Try setting maxtasksperchild=1

oliverchang:
This makes us create fresh instances for every experiment. 

Speculative fix for some issue we're seeing with hanging experiments.

### DIFF
--- a/run_all_experiments.py
+++ b/run_all_experiments.py
@@ -536,7 +536,7 @@ def main():
       _print_and_dump_experiment_result(result)
   else:
     experiment_tasks = []
-    with Pool(NUM_EXP) as p:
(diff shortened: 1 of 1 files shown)


DonggeLiu (commented):
LGTM, is this causing memory leak?

oliverchang:
> LGTM, is this causing memory leak?

Not sure, I wasn't able to debug the existing experiment in https://github.com/google/oss-fuzz-gen/pull/692 with gdb prior to  https://github.com/google/oss-fuzz-gen/pull/705. 

If this happens again, we'll be able to gdb inside the containers to see what's happening with these child processes.

PR #708, opened 2024-11-08
Test oss fuzz gen exp trigger

DonggeLiu:
Testing if `oss-fuzz-gen-exp` will be automatically triggered as in #704.


DonggeLiu:
It won't.

PR #709, opened 2024-11-08
Don't call extend_report_with_coverage_gains in apply_async callback.

oliverchang:
Per
https://docs.python.org/3/library/multiprocessing.html#multiprocessing.pool.Pool.apply_async, callbacks should return immediately or they will otherwise block the entire Pool from making progress.

For large experiments, this is likely causing problems causing our throughput to slow to a crawl as the experiment runs, as every single benchmark experiment finishing requires this expensive calculation. 

From debugging with GDB on
https://github.com/google/oss-fuzz-gen/pull/692, it looks like a large number of worker processes are stuck waiting to report results:

```
(gdb) py-bt
Traceback (most recent call first):
  File "/usr/lib/python3.11/multiprocessing/synchronize.py", line 95, in __enter__
    return self._semlock.__enter__()
  File "/usr/lib/python3.11/multiprocessing/queues.py", line 376, in put
    with self._wlock:
  File "/usr/lib/python3.11/multiprocessing/pool.py", line 131, in worker
    put((job, i, result))
  File "/usr/lib/python3.11/multiprocessing/process.py", line 108, in run
    self._target(*self._args, **self._kwargs)
  File "/usr/lib/python3.11/multiprocessing/process.py", line 314, in _bootstrap
    self.run()
```

This partially reverts https://github.com/google/oss-fuzz-gen/pull/566. We instead just create a new sub-process to periodically call this in the background to avoid blocking anything.

### DIFF
--- a/run_all_experiments.py
+++ b/run_all_experiments.py
@@ -337,7 +337,7 @@ def extend_report_with_coverage_gains_process():
   """A process that continuously runs to update coverage gains in the
   background."""
   while True:
-    time.sleep(120)  # 2 minutes.
+    time.sleep(300)  # 5 minutes.
     try:
       extend_report_with_coverage_gains()
     except Exception:
--- a/run_all_experiments.py
+++ b/run_all_experiments.py
@@ -334,7 +334,8 @@ def extend_report_with_coverage_gains() -> None:
 
 
 def extend_report_with_coverage_gains_process():
-  """A process that continuously runs to update coverage gains in the background."""
+  """A process that continuously runs to update coverage gains in the
+  background."""
   while True:
     time.sleep(120)  # 2 minutes.
     try:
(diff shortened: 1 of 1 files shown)


oliverchang:
Report: https://llm-exp.oss-fuzz.com/Result-reports/ofg-pr/2024-11-08-709-ochang-mp-comparison/index.html (triggered before the formatting fix)

oliverchang:
Report generation seems to be working. I'm going to merge this to unblock the large experimet.

DavidKorczynski:
> This partially reverts https://github.com/google/oss-fuzz-gen/pull/566. We instead just create a new sub-process to periodically call this in the background to avoid blocking anything.

Sounds good, I'll take a look at this

PR #711, opened 2024-11-10
Agent prioritize bash

DonggeLiu:
Previously, when the LLM response included both bash commands and a conclusion, we ignored the bash command and compiled the fuzz target in the conclusion alone. This approach is suboptimal, as conclusions may be premature, and the bash command often provides useful context to refine them.

To improve accuracy, we should prioritize the bash command when both are present in a response.

A more robust solution would involve structuring tasks to prevent mixed responses (e.g., using task-focused agents). But until then, prioritizing bash commands in mixed responses provides a fallback/default solution.


DonggeLiu:
Something weird with gcb exps causing exps not requested.

PR #716, opened 2024-11-10
Prioritize Bash if LLM responses both Bash and Conclusion

DonggeLiu:
Previously, when the LLM response included both bash commands and a conclusion, we ignored the bash command and compiled the fuzz target in the conclusion alone. This approach is suboptimal, as conclusions may be premature, and the bash command often provides useful context to refine them.

To improve accuracy, we should prioritize the bash command when both are present in a response.

A more robust solution would involve structuring tasks to prevent mixed responses (e.g., using task-focused agents). But until then, prioritizing bash commands in mixed responses provides a fallback/default solution.

### DIFF
--- a/agent/base_agent.py
+++ b/agent/base_agent.py
@@ -68,6 +68,7 @@ class BaseAgent(ABC):
     return filtered_code_block
 
   def _format_bash_execution_result(self, process: sp.CompletedProcess) -> str:
+    """Formats a prompt based on bash execution result."""
     stdout = self.llm.truncate_prompt(process.stdout)
     # TODO(dongge) Share input limit evenly if both stdout and stderr overlong.
     stderr = self.llm.truncate_prompt(process.stderr, stdout)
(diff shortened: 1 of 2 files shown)


DonggeLiu:
https://llm-exp.oss-fuzz.com/Result-reports/ofg-pr/2024-11-10-716-dg-comparison/index.html

mihaimaruseac (approved):

DonggeLiu:
Thanks @mihaimaruseac : )
I plan to break [these steps](https://github.com/google/oss-fuzz-gen/blob/de6497de587bd4846173e3583f1d2bbc10cf47b8/prompts/agent/prototyper-priming.txt#L6-L112) into individual agents in the next step, do you have any advices/suggestions?

mihaimaruseac:
Oh, that's great, the prompt is quite long right now and I think that splitting it over multiple agents might bring more benefits.

I think one best rule of thumb is to have each agent use only one tool. So we should not have an agent use both bash and have access to the compiler, for example

oliverchang (approved):

PR #722, opened 2024-11-21
Fix incorrect binary names

DonggeLiu:
... and a script to find mismatched binary names between benchmark sets.


oliverchang (approved):
can we just migrate to using https://introspector.oss-fuzz.com/api/harness-source-and-executable by default?

@DavidKorczynski

DonggeLiu:
> can we just migrate to using https://introspector.oss-fuzz.com/api/harness-source-and-executable by default?

Yep that can solve the problems with outdated benchmark yamls, e.g., [sleuthkit](https://introspector.oss-fuzz.com/api/harness-source-and-executable?project=sleuthkit). 
Some projects may need a bit more work, e.g., [hermes](https://introspector.oss-fuzz.com/api/harness-source-and-executable?project=hermes) has no result. I presume some projects have incorrect binary names too.

During my next exp, I will collect the projects whose fuzz target binaries cannot be found after compiling the human-written fuzz target. This step should be simple to add, but can help us rule out false negatives and inform us which projects to fix.

DonggeLiu:
Testing it a bit more, we may need to further confirm the correctness of API query result before making it default.
For example, `proftpd`'s fuzz target source file is at `/src/proftpd/tests/fuzzing/json_fuzzer.c`, but API result is `/src/fuzzer.c`.

PR #725, opened 2024-11-22
Fix granting permissions only if there is a commit

maoyixie:
It seems that permissions should be granted even without commit.


DonggeLiu (approved):
Yep nice catch, thanks!

PR #733, opened 2024-11-28
Fix that specific log levels are not displayed when using --help

maoyixie:
The original result of using '--hep' is as follows:
```
  -lo LOG_LEVEL, --log-level LOG_LEVEL
                        Sets the logging level. Options available: [{LOG_LEVELS}]
```
The result of using '--hep' after the repair is as follows:
```
  -lo LOG_LEVEL, --log-level LOG_LEVEL
                        Sets the logging level. Options available: debug, info
```


DonggeLiu (approved):
Thanks for noticing and fixing this, @maoyixie!

PR #748, opened 2024-12-10
Show successful benchmarks per project

DonggeLiu:
This shows which projects cannot benefit from OFG.

### DIFF
--- a/report/common.py
+++ b/report/common.py
@@ -79,6 +79,7 @@ class Project:
   """Results for a project entire."""
   name: str
   count: int = 0
+  success: int = 0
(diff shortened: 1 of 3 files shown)


DonggeLiu:
The new column looks good:
https://llm-exp.oss-fuzz.com/Result-reports/ofg-pr/2024-12-10-748-dg-comparison/index.html
image

I will change the column names to be more accurate, because each benchmark may have multiple (successful) harnesses generated:
* `Total generated harnesses` to `Total benchmarks`
* `Total successful harnesses` to `Successful benchmarks`

PR #753, opened 2024-12-16
Update fuzz target source file path and binary name

DonggeLiu:
`nss` failed FI build, found the new pair by improving the script in #752:

```bash
#!/usr/bin/env bash

PREFIX="OFG_UNIQUE_PREFIX_"

# Find all source files containing 'LLVMFuzzerTestOneInput' and exclude specific paths
FILES=$(find /src \
    -type f \( -name '*.c' -o -name '*.cc' -o -name '*.cpp' -o -name '*.cxx' \) \
    -not -path '*/aflplusplus/*' \
    -not -path '*/fuzztest/*' \
    -not -path '*/honggfuzz/*' \
    -not -path '*/libfuzzer/*' \
    -exec grep -l 'LLVMFuzzerTestOneInput' {} \;)

count=1
for file in $FILES; do
    echo "Inserting into $file"

    # Detect whether the file is C or C++ based on its extension
    if [[ "$file" == *.c ]]; then
        # Code for C files
        if ! grep -q "build_id_$count" "$file"; then
            cat <<EOF >>"$file"

#include <stdio.h>

const char build_id_$count[] __attribute__((used)) = "$PREFIX$file";

__attribute__((constructor))
static void reference_build_id_$count(void) {
    fprintf(stderr, "%s\\n", build_id_$count);
}

EOF
        fi
    else
        # Code for C++ files
        if ! grep -q "struct OFGBuildIdReference" "$file"; then
            cat <<EOF >>"$file"

#include <cstdio>

const char build_id_$count[] __attribute__((used)) = "$PREFIX$file";

struct OFGBuildIdReference {
    ~OFGBuildIdReference() {
        fprintf(stderr, "%s\\n", build_id_$count);
    }
};

static OFGBuildIdReference ref;

EOF
        fi
    fi

    count=$((count+1))
done

echo "Unique build_id lines inserted in all matched files."

```

```bash
#!/usr/bin/env bash
PREFIX="OFG_UNIQUE_PREFIX_"

for bin in /out/*; do
    [ -f "$bin" -a -x "$bin" ] || continue
    binary_name=$(basename "$bin")

    # Extract lines containing PREFIX, remove prefix, then check each filepath
    while IFS= read -r filepath; do
        [ -z "$filepath" ] && continue
        if [ -f "$filepath" ]; then
            export TARGET_NAME="$binary_name" TARGET_PATH="$filepath"
            echo "Binary: $TARGET_NAME"
            echo "Source file: $TARGET_PATH"
            echo
        else
            echo "Binary: $binary_name"
            echo "Source file (not found): $filepath"
            echo
        fi
    done < <(strings "$bin" | grep "$PREFIX" | sed "s/^$PREFIX//")
done

```

### DIFF
--- a/benchmark-sets/all/neomutt.yaml
+++ b/benchmark-sets/all/neomutt.yaml
@@ -57,5 +57,5 @@
   "signature": "int mutt_send_message(SendFlags, struct Email *, const char *, struct Mailbox *, struct EmailArray *, struct ConfigSubset *)"
 "language": "c++"
 "project": "neomutt"
-"target_name": "fuzz/date.c"
-"target_path": "/src/neomutt/fuzz/date.c"
+"target_name": "address-fuzz"
+"target_path": "/src/neomutt/fuzz/address.c"
--- a/benchmark-sets/all/nss.yaml
+++ b/benchmark-sets/all/nss.yaml
@@ -79,5 +79,5 @@
   "signature": "CK_RV s_open(const char *, const char *, const char *, int, int, int, SDB **, SDB **, int *)"
 "language": "c++"
 "project": "nss"
-"target_name": "mpi-mod"
-"target_path": "/src/nss/fuzz/mpi_mod_target.cc"
+"target_name": "certDN"
+"target_path": "OFG_UNIQUE_PREFIX_/src/nss/fuzz/targets/certDN.cc"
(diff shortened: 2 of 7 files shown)


DonggeLiu:
`tinyusb` is a particularly tricky case because it optimizes out unused code, so the previous method (declaring a string) does not work.
The new method above works.

DonggeLiu:
`wuffs` is another corner case, where each fuzz target is built with another file `/src/wuffs/fuzz/c/fuzzlib/fuzzlib.c` that also contains `LLVMFuzzerTestOneInput`.
This is probably why FI thought `fuzzlib.c` is a fuzz target.
We need to capture and handle this case.

DonggeLiu:
All done!

DavidKorczynski (approved):

PR #798, opened 2025-02-18
Add timestamp suffix to avoid conflict

arthurscchan:
The docker container created during local source copy from local is only used within the function. But it will result in docker container collision when it is running across multiple pipelines. This PR fixes the naming of the one-use docker container to add a timestamp suffix (and retry with random sleep when collision occur) to avoid docker container collision.

### DIFF
--- a/data_prep/project_src.py
+++ b/data_prep/project_src.py
@@ -269,10 +269,11 @@ def _copy_project_src_from_local(project: str, out: str, language: str):
     logger.info('Done copying %s /src to %s.', project, out)
   finally:
     # Shut down the container that was just started.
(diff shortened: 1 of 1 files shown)


arthurscchan:
/gcbrun exp -n ascc-pr-798-1 -m vertex_ai_gemini-1-5 -i -b comparison

DonggeLiu:
The report looks good:
https://llm-exp.oss-fuzz.com/Result-reports/ofg-pr/2025-02-18-798-ascc-pr-798-1-comparison/index.html

DonggeLiu (approved):

DavidKorczynski (approved):

PR #785, opened 2025-02-10
Merge the default one-prompt method into the new agentic workflow (Analysis+Enhancement)

DonggeLiu:
* [x]  Merge run step of check_targets() into ExecutionStage, to collect runtime stats.
* [x] Merge triage_crash() + the following steps into AnalysisStage()->Triager
* [x] Test build-ability with the new `container_tool` from agentic workflow instead of the old `target_experiment.py` script.
* [x] Split build_and_run_cloud into 3 functions:
    * [ ] ~~Build and collect results in 'OnePromptPrototyper'~~ 
    * [x] Use `container_tool` in OnePromptPrototyper/OnePromptEnhancer.
    * [x] Build+run and collect results in ExecutionStage.
* [x] Triage results in the Triager
* [x] Split _fix_generated_fuzz_target() into 2 functions
    * [x] Fix build failures in Protyper
    * [x] Fix runtime failure in Refiner.
    * [x] Refactor the function.

Next:
* [ ] Ensure GKE uses service account.
* [ ] Remove `JCC`.
* [ ] Add `cycle` in results.
* [ ] Use `cycle` when saying logs.
* [ ] Replace old `Run/BuildResult`s with the new one.
* [ ] Clean up repeated code (e.g., parse_libfuzzer_log)

PR #800, opened 2025-02-18
models: add ask_llm to GPT

DavidKorczynski:
Following https://github.com/google/oss-fuzz-gen/pull/785 we need to use a `ask_llm` in the GPT models. We should propagate this further through but will do that in follow-ups.


DonggeLiu (approved):
Thanks for addressing this so quickly!

PR #802, opened 2025-02-19
Disable JCC and the reliance of its err.log

DonggeLiu:
Reasons:
1. We don't need `JCC` to parse and pass fuzz target build error messages (into `err.log`), because agents can capture and fix build errors in project's containers natively, without relying on `target_experiment.py`. Some err.log were empty anyway.
2. Agents can modify build script, hence we don't need JCC to add flags, add header files, or switch compiler (C/C++).
3. `JCC`'s heuristics cause build errors in some projects (e.g., `brpc`).

### DIFF
--- a/experiment/builder_runner.py
+++ b/experiment/builder_runner.py
@@ -42,7 +42,6 @@ logger = logging.getLogger(__name__)
 logger.setLevel(logging.DEBUG)
 
 # The directory in the oss-fuzz image
-JCC_DIR = '/usr/local/bin'
 RUN_TIMEOUT: int = 30
 CLOUD_EXP_MAX_ATTEMPT = 5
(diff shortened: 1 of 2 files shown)


oliverchang (approved):
LGTM, but did you mean "reliance" in your PR title?

DonggeLiu:
I reckon this is ready to merge.
Cloud log has the following error, due to missing enhancer, which is unrelated to this PR.
image

It is on the top of my TODO list (right after adding the SemanticAnalyzer), I will fix this in another PR.

jonathanmetzman:
LGTM and RIP.
Why don't we delete it?

jonathanmetzman:
Ah OK.

PR #804, opened 2025-02-19
Fix Agent prompt and infra

DonggeLiu:
Fixing some issues revealed by full agent experiments earlier:
1. [x] LLM-generated build scripts do not save fuzz target binary into the correct path.
2. [x] Use default build script in the code fixing prompt in this scenario:
    1. The default build scripts builds successfully but failed other checks (i.e., reference), and
    2. The LLM-generated build script does not work.
3. [x] Selectively use the default built script and the LLM-generated built script, depending which is better.
4. [x] Use different code-fixing prompts based on which built script and which result it is:
    * default or LLM built script
    * No reference, no binary, or compilation failure
5. [x] Backup human-writtent `/src/build.sh` to `/src/build.bk.sh` in agent's containers in case LLM wants to reuse it in the new build script.
    * Create the same copy for fuzzing execution.
6. [x] Hide the compile command to prevent LLM from reusing it in the inspection tool and be distracted by irrelevant errors. E.g.:
    * The inspection container always runs compile before LLM analysis. Rerunning it may fail in some projects due to an existing /src/<project>/build directory.
7. [x] Prompt use example fuzz target in the language the same as the generated fuzz target, (not the project).
    * Also dynamically adjust instructions in priming. Do not leave LLM to judge which language the fuzz target is.
8. [x] Remove the agent log when receiving fuzz targets.
9. [x] Do not restrict LLM to send one bash command per query.

Also need to:
1. [ ] Use SemanticAnalyzer in agent workflow, at least to ensure the last Result is Analysis Result.
2. [ ] Add an Enhancer in agent workflow.
3. [ ] Use service account in GKE, hopefully this will solve the [`Service Unavailable` problem](https://github.com/google/oss-fuzz/pull/13042).

### DIFF
--- a/agent/prototyper.py
+++ b/agent/prototyper.py
@@ -251,8 +251,8 @@ class Prototyper(BaseAgent):
       # binary to expected path, but does not reference function-under-test.
       prompt_text.replace(
           '{BUILD_TEXT}',
-          'Althoug `/src/build.sh` compiles and saves the binary to the correct'
-          ' path:')
+          'Althoug `/src/build.bk.sh` compiles and saves the binary to the '
+          'correct path:')
       # NOTE: Unsafe to say the following, because /src/build.sh may miss a
       # library required by the function-under-test, and the fuzz target did not
       # invoke the function-under-test either.
@@ -310,8 +310,8 @@ class Prototyper(BaseAgent):
           build_result.benchmark.target_name,
           trial=build_result.trial)
       prompt_text = (
-          'The fuzz target compiles successfully with /src/build.sh, but the '
-          'final fuzz target binary was not saved to the expected path at '
+          'The fuzz target compiles successfully with /src/build.bk.sh, but the'
+          ' final fuzz target binary was not saved to the expected path at '
           f'`{binary_path}`.\n'
           f'<fuzz target>\n{fuzz_target_source}\n</fuzz target>\n'
           f'<compilation log>\n{compile_log}\n</compilation log>\n'
--- a/agent/prototyper.py
+++ b/agent/prototyper.py
@@ -238,18 +238,18 @@ class Prototyper(BaseAgent):
                                            extra_text=prompt.get()).strip()
     prompt_text = (
         "The fuzz target's `LLVMFuzzerTestOneInput` did not invoke the "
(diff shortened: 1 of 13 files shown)


DonggeLiu:
Report looking OK:
https://llm-exp.oss-fuzz.com/Result-reports/ofg-pr/2025-02-19-804-dg1-comparison/index.html

mihaimaruseac (approved):

mihaimaruseac (approved):

DonggeLiu:
Report:
https://llm-exp.oss-fuzz.com/Result-reports/ofg-pr/2025-02-23-804-dg-comparison/index.html

It's the first time seeing LLM generated a compilable fuzz target for a `Bazel` project:
https://llm-exp.oss-fuzz.com/Result-reports/ofg-pr/2025-02-23-804-dg-comparison/sample/output-abseil-cpp-_zzn4absl19str_format_internal12_global__n_115binarytodecimal13runconversionens_7uint128eins_11functio/03.html

DavidKorczynski:
I'm experiencing regressions in the traditional workflow in https://github.com/google/oss-fuzz-gen/pull/815 Am not sure but where the regression comes from, but considering the size of this PR I suspect it may come from here. Is there a reason not to run experiments without `-ag` for this PR, and do you have ideas if this is causing the regression?

DavidKorczynski:
I fixed the regression in https://github.com/google/oss-fuzz-gen/pull/815 It doesn't seem to come from this PR but rather https://github.com/google/oss-fuzz-gen/pull/785 which is a bit surprising since we've run successful experiments since that landed. So I'm not fully sure that it's not this PR

DonggeLiu:
> I fixed the regression in #815 It doesn't seem to come from this PR but rather #785 which is a bit surprising since we've run successful experiments since that landed. So I'm not fully sure that it's not this PR

Thanks for the information.

> Is there a reason not to run experiments without -ag for this PR, and do you have ideas if this is causing the regression?

I did not run experiments without `-ag` because 
1. I expect all problems with the traditional workflow have been addressed in #785.
2. This PR should not affect the performance of transitional workflow (which uses `OnePrompt*` agents). For example, changes like introducing `build.bk.sh` should not matter.

PR #805, opened 2025-02-19
Fix validation for function reference

arthurscchan:
The new approach force the function reference check on the built binary for LLVMTestOneInput and this is only valid for C/C++ projects. This make the prototyper check always failed for other language. This PR temporary skip that validation for other languages which the validation is not valid.

### DIFF
--- a/agent/one_prompt_prototyper.py
+++ b/agent/one_prompt_prototyper.py
@@ -236,6 +236,13 @@ class OnePromptPrototyper(BaseAgent):
       cur_round: int, trial: int) -> bool:
     """Validates if the LLM generated fuzz target assembly code references
     function-under-test."""
+
+    # LLVMFuzzerTestOneInput and binary dumps are only valid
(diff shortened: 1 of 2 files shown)


arthurscchan:
/gcbrun exp -n ascc-pr-805-1 -m vertex_ai_gemini-1-5 -i -b comparison

DavidKorczynski:
hmm, the experiment seems to be stuck in an infinite loop. It's not terminated and the logs keep going. Will need to investigate

DavidKorczynski:
> hmm, the experiment seems to be stuck in an infinite loop. It's not terminated and the logs keep going. Will need to investigate

This is not true, it was just a case of still running. For some reason I thought you had started the experiment 10+ hours ago, but it was only start 1 hour or so before I wrote the comment. Nevermind -- things are looking well. Will land once the experiment is completed.

DavidKorczynski (approved):

PR #808, opened 2025-02-20
Add handling for empty response

arthurscchan:
In some cases, the llm may fail to generate any response (maybe timeout or connection issues), it will result in None response and that creates an error when it is passed in for regular expression tag parsing. This PR fixes that by ensuring response is not None before passing in to process.


arthurscchan:
/gcbrun exp -n ascc-pr-808-1 -m vertex_ai_gemini-1-5 -i -b comparison

arthurscchan:
/gcbrun exp -n ascc-pr-808-2 -m vertex_ai_gemini-1-5 -i -b comparison

DavidKorczynski (changes requested):
I feel like if there is an error, wouldn't we want to at least log this somewhere or so? Can it lead to confusion just passing over it in this way?

arthurscchan:
/gcbrun exp -n ascc-pr-808-3 -m vertex_ai_gemini-1-5 -i -b comparison

DavidKorczynski (approved):

PR #823, opened 2025-02-28
Fix prompt text return bug

arthurscchan:
OpenAIPrompt's get function will return a list of prompts with role and content instead of pure string. This will cause error in later bash result handling. This PR adds a new gettext() function in OpenAiPrompt to combine content string from multiple prompt to a single string for processing.

### DIFF
--- a/agent/base_agent.py
+++ b/agent/base_agent.py
@@ -23,7 +23,7 @@ from typing import Any, Optional
 import logger
 import utils
 from llm_toolkit.models import LLM
-from llm_toolkit.prompts import OpenAIPrompt, Prompt
+from llm_toolkit.prompts import Prompt
 from results import Result
 from tool.base_tool import BaseTool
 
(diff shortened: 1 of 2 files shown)


arthurscchan:
/gcbrun exp -n ascc-pr-823-1 -m vertex_ai_gemini-1-5 -i -b comparison

arthurscchan:
/gcbrun exp -n ascc-pr-823-2 -m vertex_ai_gemini-1-5 -i -b comparison

agent/base_agent.py
```diff
@@ -116,7 +116,10 @@ def _format_bash_execution_result(
       previous_prompt: Optional[Prompt] = None) -> str:
     """Formats a prompt based on bash execution result."""
     if previous_prompt:
-      previous_prompt_text = previous_prompt.get()
+      if isinstance(previous_prompt, OpenAIPrompt):
+        previous_prompt_text = previous_prompt.gettext()
+      else:
+        previous_prompt_text = previous_prompt.get()
```
DonggeLiu, line 122: Thanks for noticing and fixing this, @arthurscchan.
Would you think me could add a new `gettext()` method for `TextPrompt` (and `Prompt`) too? The new method returns the same value as `get()` in `TextPrompt`.
This allows us to call `previous_prompt.gettext()` transparently for both `TextPrompt` and `OpenAIPrompt`.

The main intention is to avoid special cases (i.e., `if isinstance...`) because over time more and more special cases will become harder to maintain.

DonggeLiu: Thanks @arthurscchan  : )

arthurscchan:
/gcbrun exp -n ascc-pr-823-3 -m vertex_ai_gemini-1-5 -i -b comparison

DavidKorczynski:
https://llm-exp.oss-fuzz.com/Result-reports/ofg-pr/2025-03-03-823-ascc-pr-823-3-comparison/index.html

DavidKorczynski (approved):

PR #824, opened 2025-02-28
OpenAiPrompt: Ensure no empty content

arthurscchan:
OpenAI prompt are generally a list of prompts with role and content. Sometimes, prompt with empty content will confuse the OpenAI API and cause unexpected responses. This PR fixes the functions that handle prompt content addition to avoid adding new prompt if the content is empty.


arthurscchan:
/gcbrun exp -n ascc-pr-824-1 -m vertex_ai_gemini-1-5 -i -b comparison

arthurscchan:
/gcbrun exp -n ascc-pr-824-2 -m vertex_ai_gemini-1-5 -i -b comparison

DavidKorczynski (changes requested):
if we call any of these methods with an empty string, isn't there a problem already? I.e. a fix to silent return doesn't seem a smart choice here? In what contexts do we end up with empty strings?

We should at least log the cases IMO

DavidKorczynski:
https://llm-exp.oss-fuzz.com/Result-reports/ofg-pr/2025-03-04-824-ascc-pr-824-3-comparison/index.html

DavidKorczynski (approved):

Issue #780, opened 2025-02-06
Add missing licence headers and a corresponding CI check.


DonggeLiu:
Example CI from OSS-Fuzz:
https://github.com/google/oss-fuzz/blob/abb18716350b24db913d642e85e87fc621059f9f/.github/header-checker-lint.yml

DavidKorczynski:
Solved by https://github.com/google/oss-fuzz-gen/pull/806

PR #806, opened 2025-02-19
ci: add license checking

DavidKorczynski:

PR #828, opened 2025-03-02
Add CI workflow to check license headers

HarshitNagpal29:
Solves #780


DavidKorczynski (changes requested):
Ah, thanks for the effort. This was done here https://github.com/google/oss-fuzz-gen/pull/806 following a similar style to OSS-Fuzz. Apologies for not closing the issue.

Issue #743, opened 2024-12-05
Interact with LLM in JSON


DonggeLiu:
Currently we use XML-tag, it would be nice to [use JSON](https://github.com/google/oss-fuzz-gen/pull/736#discussion_r1870449044) as that seems to be more common, e.g., in Tool API.

One potential issue is escaping special characters. We have seen cases where LLM tend to add redundant `\`s, which may have performance impact.

nugget-cloud:
@DonggeLiu Hi can you assign this issue to me

DonggeLiu:
Hi @nugget-cloud Thanks for willing to help.
#744 could be a good starting point for you : )

nugget-cloud:
Hi, @DonggeLiu 
I am trying to understand the code changes I need to make to interact with the LLM in JSON 
- in the prompts directory, change the XML to JSON 
- make changes in the llm_toolkit/prompts_builder and prompts and output_parser.py to handle JSON output 

Please correct me if I am wrong here

DonggeLiu:
Thanks @nugget-cloud!
I suggest breaking this task into stages, starting with a simpler change before moving to more complex ones.

## Stage 1: Converting XML prompts to JSON
The goal here is to improve clarity in our requests to the LLM by switching from XML to JSON, which is a more commonly used format. The core logic remains the same.

TODOs:
* This is essentially [#744](https://github.com/google/oss-fuzz-gen/pull/744).
* You can fix the conflict and may also need to check for places (e.g., in `agents/`) where prompts are constructed dynamically instead of using templates in `prompts/`.
* No changes are needed for `prompt_builder.py` or `prompts.py` since we still expect the LLM to return plain text responses.

We can stop here if Stage 2 turns out to be too complex—your help is greatly appreciated regardless!


## Stage 2: Structuring LLM Responses in JSON
This stage requires deeper modifications, ensuring the LLM directly returns structured JSON responses rather than plain text.

TODOs:
1. Use [GenerationConfig](https://cloud.google.com/vertex-ai/generative-ai/docs/multimodal/control-generated-output) to enforce JSON responses. This involves defining a response_schema, which varies based on the expected output.
2. To simplify schema management, we will likely need to modify `prompt_builder.py` and `prompts.py`, potentially introducing a new class that extends existing text-based ones but is tailored for JSON-based responses.


[...]

PR #843, opened 2025-03-08
conversion from xml to json

nugget-cloud:
#743 
Hi, I made some changes. I removed XML tags from all the prompts in the prompts directory. 
Can you share some feedback on the changes if I am doing it correctly?


run-auto.sh
```diff
@@ -0,0 +1,21 @@
+BASE_OFG=$PWD
```
DonggeLiu, line 1: I don't think you meant to push this and the `fuzz-introspector` module below?

DonggeLiu (commented):
Thanks @nugget-cloud.
Given this is an exploration work, we can start from some small experiments using agent-related templates (e.g., in `prompts/agent`, `prompts/tool`).
If they prove to be useful, we can extend the JSON structure to more prompts.

Some minor comments below:

DonggeLiu:
> I attempted to run run_all_experiments.py, but I stopped midway because I was concerned it was making too many calls and that the API might be expensive.

Yep, it will be expensive to run it locally.
We have PR experiments which runs it on our cloud and won't cost you anything.

However, PR experiments can only be run by the maintainer of this repo.
The simplest solution is:
1. I create a new PR for us to run experiments (Done here: #844).
2. You rebase your PR on to the new experiment PR.
3. I will merge your PR to run experiments from there.

For simplicity, let's revert the conflict files for now (we need to start with some small experiments using agent-related templates anyway) so that you don't have to repeatedly fixing them.
But you do have to solve the CLA error in CI.

run-auto.sh
nugget-cloud: Yes , Sorry i completely forgot about this 
I will fix this

nugget-cloud:
Hi , @DonggeLiu I signed the CLA but the error still persists , I am not sure what is the issue

DonggeLiu:
> Hi , @DonggeLiu I signed the CLA but the error still persists , I am not sure what is the issue

It's because of this:
image

image

DonggeLiu:
Converting this to a draft for now, but please let me know when you are ready. @nugget-cloud

PR #847, opened 2025-03-10
Fix benchmark status reporting to accurately reflect trial completion

### LINKED ISSUE #721: Incorrect Benchmark Status on report's Index Page

ay-bh:
### Changes
Modified the benchmark status determination logic in `common.py` to:

1. Added a new helper method `_get_expected_trials_count()` that scans benchmark directories to determine the total expected number of trials
2. Updated `match_benchmark()` to mark benchmarks as "Done" only when all expected trials are complete

### Testing

Tested the solution by creating benchmarks in various states using a local script:

- Complete benchmark with all trials present and successful
- Partial benchmark with only some trials having results
- Failed benchmark with all trials present but some failing

Verified the status was correctly reported as "Running" for partial and failed benchmarks, and "Done" only for the complete benchmark

Fixes #721

### DIFF
--- a/report/common.py
+++ b/report/common.py
@@ -269,7 +269,10 @@ class Results:
   def match_benchmark(self, benchmark_id: str, results: list[evaluator.Result],
                       targets: list[str]) -> Benchmark:
     """Returns a benchmark class based on |benchmark_id|."""
-    status = 'Done' if results and all(results) else 'Running'
+    expected_trials = self._get_expected_trials_count(benchmark_id)
+    status = 'Done' if (results and len(results) == expected_trials and
+                        all(results)) else 'Running'
+
     filtered_results = [(i, stat) for i, stat in enumerate(results) if stat]
 
     if filtered_results:
@@ -278,6 +281,35 @@ class Results:
       result = run_one_experiment.AggregatedResult()
 
     return self._create_benchmark(benchmark_id, status, result)
+
+  def _get_expected_trials_count(self, benchmark_id: str) -> int:
+    """Returns the expected number of trials for a benchmark."""
+    fuzz_targets_dir = os.path.join(self._results_dir, benchmark_id,
+                                    'fuzz_targets')
+    if not FileSystem(fuzz_targets_dir).exists():
+      # Counting files in raw_targets directory for older experiments
+      raw_targets_dir = os.path.join(self._results_dir, benchmark_id,
+                                     'raw_targets')
+      if not FileSystem(raw_targets_dir).exists():
+        return 0
+      targets = [
+          f for f in FileSystem(raw_targets_dir).listdir()
+          if os.path.splitext(f)[1] in TARGET_EXTS
+      ]
+      return len(targets)
+
+    # Count unique trial IDs in the fuzz_targets directory
+    trial_ids = set()
+    for filename in FileSystem(fuzz_targets_dir).listdir():
+      if os.path.splitext(filename)[1] in TARGET_EXTS:
+        # Extract trial ID from filenames like "01.fuzz_target"
+        try:
+          trial_id = os.path.splitext(filename)[0]
+          trial_ids.add(trial_id)
+        except (ValueError, IndexError):
+          continue
+
+    return len(trial_ids)
 
   def get_final_target_code(self, benchmark: str, sample: str) -> str:
     """Gets the targets of benchmark |benchmark| with sample ID |sample|."""
--- a/report/common.py
+++ b/report/common.py
@@ -302,17 +302,13 @@ class Results:
       for filename in FileSystem(fuzz_targets_dir).listdir():
         if os.path.splitext(filename)[1] in TARGET_EXTS:
           # Extract trial ID from filenames like "01.fuzz_target"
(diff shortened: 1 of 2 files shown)


report/common.py
```diff
@@ -279,6 +282,35 @@ def match_benchmark(self, benchmark_id: str, results: list[evaluator.Result],
 
     return self._create_benchmark(benchmark_id, status, result)
 
+  def _get_expected_trials_count(self, benchmark_id: str) -> int:
+    """Returns the expected number of trials for a benchmark."""
+    fuzz_targets_dir = os.path.join(self._results_dir, benchmark_id,
+                                    'fuzz_targets')
+    if not FileSystem(fuzz_targets_dir).exists():
+      # Counting files in raw_targets directory for older experiments
```
DonggeLiu, line 290: It's great to see you've noticed the difference ways to store fuzz targets between new/old experiments.

How about organize the logic to:
```
if FileSystem(fuzz_targets_dir).exists():
  # Check and return the trial count in new experiments.

# Check and return the trial count in old experiments.
```

There is a catch in counting the number of trials in the new experiment: Not all fuzz targets will be returned immediately at once. For example, we may have fuzz targets from trial 8 and 10 first, then from trial 2 and 5, and later the rest.
There are 2 solutions:
1. In most cases, we can assume `run_all_experiments.py` (which takes the trial count from `--num-samples`) is run under the same fresh environment as report generation. Hence we can add a line in `run_all_experiments.py` to record `num-samples` in an ENV VAR, and reuse it in report generation.
2. In rare cases where the ENV VAR is not set, we can assume the max trial ID so far is the trial count as a temp solution. E.g., if we have trial 01, 05, 10, then we assume 10 trials in total.

The workflow would be:
1. Check and use ENV VAR.
2. If not available, check `fuzz_targets` dir and count trials in new experiments.
3. If not available, check `raw_targets` dir and count trials in old experiments.
4. return a default value (0 or 1).

I reckon this is the minimum changes required to solve this, but please feel free to let me know if you can think of  a more elegant/complete/sound solution : )

DonggeLiu: BTW, thanks for addressing this so quickly!

ay-bh: @DonggeLiu Thanks, I find the work you are doing really interesting. 

We can also use a metadata approach to write experiment parameters (like the expected trial count) to a dedicated file during initialization, ensuring a persistent source for reporting. File-based record decouples reporting from runtime quirks such as environment variables, etc. But both approaches work well, while the metadata method feels more sophisticated, I believe they are equally effective overall, what do you think?

ay-bh: @DonggeLiu I have a doubt: can a trial ID be '00', or do they always start at '01'? If '00' is possible, then in a case with trials '01', '05', and '10', the total trial count would be 11 rather than 10.

DonggeLiu: > We can also use a metadata approach to write experiment parameters (like the expected trial count) to a dedicated file during initialization,

Yep, similar to #105. But let's start simple in this PR before others accidentally modify the same code and result in a code conflict. I also want to leave you more time writing the proposal, which is main thing we will consider when reviewing applications.

> can a trial ID be '00'? 

Not that I know of (under this use case).
If you find evidence that proves me wrong, please let me know :)

ay-bh: @DonggeLiu I have accommodated the changes. Please review now. Thank you.

DonggeLiu (commented):
Thanks @ay-bh!
Just a final nit.

DonggeLiu:
Since the PR is mostly ready, I created #859 so that we can run experiments on PRs and see the results before merging.

report/common.py
```diff
@@ -279,6 +282,56 @@ def match_benchmark(self, benchmark_id: str, results: list[evaluator.Result],
...
+            trial_id = os.path.splitext(filename)[0]
+            trial_ids.add(trial_id)
+
+            try:
+              trial_num = int(trial_id)
+              max_trial_id = max(max_trial_id, trial_num)
+            except ValueError:
+              pass
+          except (ValueError, IndexError):
+            continue
```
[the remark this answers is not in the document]
ay-bh: @DonggeLiu on a closer look, the outer try-except seems redundant. `os.path.splitext(filename)[0]` will likely never raise an`IndexError` (it always returns a 2-tuple, even for filenames without an extension). A `ValueError` probably wouldn’t occur here either, since string splitting doesn’t trigger it. I’ll fix this in a couple of hours, currently away from my desk.

Thanks for pointing it out.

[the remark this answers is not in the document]
ay-bh: Done

DonggeLiu (approved):

DonggeLiu:
Thanks @ay-bh, merged this into #859 for expriements.

Issue #366, opened 2024-06-21
Assert temperature in argparser


DonggeLiu:
Different models have different temperature ranges.
It's better to assert its value dynamically based on the model used:
https://github.com/google/oss-fuzz-gen/blob/main/run_all_experiments.py#L246

dberardi99:
Hi @DonggeLiu !
I'd like to help with this issue. Could it be assigned to me?

DonggeLiu:
> Hi [@DonggeLiu](https://github.com/DonggeLiu) ! I'd like to help with this issue. Could it be assigned to me?

Sure, do you know the temperature range of the models we use?

dberardi99:
I've found the following temperature ranges:

| Model name | Vertex AI Model Name (if applicable) | Temperature Range | Default Temperature
| --- | --- | :---: | :---: |
| vertex_ai_code-bison | code-bison | 0.0 - 1.0 | 0.2 |
| vertex_ai_code-bison-32k | code-bison-32k | 0.0 - 1.0 | 0.2 |
| vertex_ai_gemini-pro | gemini-1.0-pro | 0.0 - 1.0 | 0.9 |
| vertex_ai_gemini-ultra | gemini-ultra | ? | ? |
| vertex_ai_gemini-experimental | gemini-experimental | 0.0 - 2.0 | 1.0 |
| vertex_ai_gemini-1-5 <br> vertex_ai_gemini-1-5-chat | gemini-1.5-pro-002 | 0.0 - 2.0 | 1.0 |
| vertex_ai_gemini-2-flash <br> vertex_ai_gemini-2-flash-chat | gemini-2.0-flash-001 | 0.0 - 2.0 | 1.0 |
| vertex_ai_gemini-2 <br> vertex_ai_gemini-2-chat | gemini-2.0-pro-exp-02-05 | 0.0 - 2.0 | 1.0 |
| vertex_ai_gemini-2-think <br> vertex_ai_gemini-2-think-chat | gemini-2.0-flash-thinking-exp-01-21 | 0.0 - 2.0 | 0.7 |
| vertex_ai_claude-3-haiku | claude-3-haiku@20240307 | 0.0 - 1.0 | 0.5 |
| vertex_ai_claude-3-opus | claude-3-opus@20240229 | 0.0 - 1.0 | 0.5 |
| vertex_ai_claude-3-5-sonnet | claude-3-5-sonnet@20240620 | 0.0 - 1.0 | 0.5 |
[...]

PR #856, opened 2025-03-12
Dynamically assert model temperature value in argparser

dberardi99:
A new method `get_model_temperature(args: argparse.Namespace)` has been added to hardcode the model temperature value based on the model name. The chat session models (namely the ones ending with _"-chat"_ or _"-azure"_) have been treated as the corresponding base models. The temperature values can be found in the following data table https://github.com/google/oss-fuzz-gen/issues/366#issuecomment-2708184530

The temperature value will be automatically aligned to the one of the model chosen only if no temperature has been set (`args.temperature == TEMPERATURE`)

In addition, the default temperature has been changed to 1.0 since it is the one relative to the default model (_vertex_ai_gemini-1-5_)

Fix https://github.com/google/oss-fuzz-gen/issues/366

### DIFF
--- a/llm_toolkit/models.py
+++ b/llm_toolkit/models.py
@@ -46,7 +46,7 @@ logger = logging.getLogger(__name__)
 # Model hyper-parameters.
 MAX_TOKENS: int = 2000
 NUM_SAMPLES: int = 1
-TEMPERATURE: float = 0.4
+TEMPERATURE: float = 1.0
 
 
 class LLM:
--- a/run_all_experiments.py
+++ b/run_all_experiments.py
@@ -279,6 +279,9 @@ def parse_args() -> argparse.Namespace:
 
   if args.temperature:
     assert 2 >= args.temperature >= 0, '--temperature must be within 0 and 2.'
+    
+  if args.temperature == TEMPERATURE and args.model in models.LLM.all_llm_names():
+    args.temperature = run_one_experiment.get_model_temperature(args)
 
   benchmark_yaml = args.benchmark_yaml
   if benchmark_yaml:
--- a/run_one_experiment.py
+++ b/run_one_experiment.py
@@ -54,7 +54,7 @@ NUM_EVA = int(os.getenv('LLM_NUM_EVA', '3'))
 NUM_SAMPLES = 2
 MAX_TOKENS: int = 4096
 RUN_TIMEOUT: int = 30
-TEMPERATURE: float = 0.4
+TEMPERATURE: float = 1.0
 
 RESULTS_DIR = './results'
 
@@ -311,3 +311,30 @@ def run(benchmark: Benchmark, model: models.LLM, args: argparse.Namespace,
 
   return AggregatedResult.from_benchmark_result(
       _fuzzing_pipelines(benchmark, model, args, work_dirs))
+
+
+def get_model_temperature(args: argparse.Namespace) -> float:
+  """Retrieves model temperature default value."""
+  default_temperatures = {models.VertexAICodeBisonModel.name: 0.2,
+                          models.VertexAICodeBison32KModel.name: 0.2,
+                          models.GeminiPro.name: 0.9,
+                          models.GeminiUltra.name: 0.2,
+                          models.GeminiExperimental.name: 1.0,
+                          models.GeminiV1D5.name: 1.0,
+                          models.GeminiV2Flash.name: 1.0,
+                          models.GeminiV2.name: 1.0,
+                          models.GeminiV2Think.name: 0.7,
+                          models.ClaudeHaikuV3.name: 0.5,
+                          models.ClaudeOpusV3.name: 0.5,
+                          models.ClaudeSonnetV3D5.name: 0.5,
+                          models.GPT.name: 1.0,
+                          models.GPT4.name: 1.0,
+                          models.GPT4o.name: 1.0,
+                          models.GPT4oMini.name: 1.0,
+                          models.GPT4Turbo.name: 1.0}
+  if args.model.endswith('-chat') or args.model.endswith('-azure'):
+    model_name = '-'.join(args.model.split('-')[:-1])
+  else:
+    model_name = args.model
+  
+  return default_temperatures[model_name]


dberardi99:
Hi @DonggeLiu Could you please review it when you get a chance? Many thanks in advance!

llm_toolkit/models.py
```diff
@@ -46,7 +46,7 @@
 # Model hyper-parameters.
 MAX_TOKENS: int = 2000
 NUM_SAMPLES: int = 1
-TEMPERATURE: float = 0.4
+TEMPERATURE: float = 1.0
```
DonggeLiu, line 49: Ops let's keep the default temperature the same for now to avoid causing surprising results in other people's recent experiments.
We can grid search for the best default values for our use case later.

run_all_experiments.py
```diff
@@ -279,6 +279,9 @@ def parse_args() -> argparse.Namespace:
 
   if args.temperature:
     assert 2 >= args.temperature >= 0, '--temperature must be within 0 and 2.'
+    
+  if args.temperature == TEMPERATURE and args.model in models.LLM.all_llm_names():
+    args.temperature = run_one_experiment.get_model_temperature(args)
```
DonggeLiu, line 284: I reckon the issue we solve has 2 tasks:
1. Main: Different models have different temperature ranges, we want to assert that if user specified a temperature in args, then it should fall into the corresponding models range.
2. Minor: Define a default temperature for each model.

let's solve the main task first:
1.1. Define the temperature rate for each model class in https://github.com/google/oss-fuzz-gen/blob/main/llm_toolkit/models.py. Use inheritance to minimize the changes needed.
1.2. Replace this hardcoded assertion with dynamic assertion based on the model name:
https://github.com/google/oss-fuzz-gen/blob/33bddff778cd3af258c90ca32c4ed2d1f671699b/run_all_experiments.py#L280-L281

Then we can work on the minor task:
1. Add default temperatures under each class.
2. Set the temperature as the default value here, if user did not specify it.

DonggeLiu (changes requested):
Thanks @dberardi99.
I left a comment to clarify a bit more about the task.
Please let me know if that makes sense : )

llm_toolkit/models.py
dberardi99: Perfect, I got it. I'll reset it to its prior value

run_all_experiments.py
dberardi99: Everything's clear! Just one thing, do you prefer to solve only the different temperature ranges in this PR or to implement both temperature ranges and default temperatures here?

llm_toolkit/models.py
```diff
@@ -61,6 +61,8 @@ class LLM:
 
   _max_attempts = 5  # Maximum number of attempts to get prediction response
 
+  temperature_range: list[float] = [0.0, 2.0]  # Default model temperature range
```
DonggeLiu, line 64: Let's keep this attribute but make the value a constant  like `TEMPERATURE` defined at the top of the file:
```python
temperature_range: list[float] = TEMPERATURE_RANGE
```

llm_toolkit/models.py
```diff
@@ -125,6 +127,16 @@ def all_llm_names(cls):
...
+  
+  @classmethod
+  def all_llm_temperature_ranges(cls):
+    """Returns the current model and all child temperature ranges."""
+    ranges = {}
+    for subcls in cls.all_llm_subclasses():
+      if (hasattr(subcls, 'temperature_range') and hasattr(subcls, 'name')
+          and subcls.name != AIBinaryModel.name):
+        ranges[subcls.name] = subcls.temperature_range
+    return ranges
```
DonggeLiu, line 139: IIUC, you are replicating `all_llm_names`.
Would it be more maintainable and extensible to extract and reuse the repeating logic of these two functions? E.g., a function to return all models so that we can reuse it to acquire other attributes in the future.

```python
    @classmethod
    def _all_llm_models(cls):
        """
        Returns a list of LLM model classes that have a `name` attribute
        and are not `AIBinaryModel`.
        """
        models = []
        for subcls in cls.all_llm_subclasses():
            # May need a different filter logic here.
            if subcls.name != AIBinaryModel.name:
                models.append(subcls)
        return models

    @classmethod
    def all_llm_names(cls) -> list[str]:
        """Returns the current model name and all child model names."""
        return [m.name for m in cls._all_llm_models()]

    @classmethod
    def all_llm_temperature_ranges(cls) -> dict[str, list[float, float]]:
        """Returns the current model name and all child model temperature ranges."""
        return {
            m.name: m.temperature_range
            for m in cls._all_llm_models()
            if hasattr(m, 'temperature_range')
        }
```

run_all_experiments.py
```diff
@@ -278,7 +278,10 @@ def parse_args() -> argparse.Namespace:
     assert args.num_samples > 0, '--num-samples must take a positive integer.'
 
   if args.temperature:
-    assert 2 >= args.temperature >= 0, '--temperature must be within 0 and 2.'
+    ranges = models.LLM.all_llm_temperature_ranges()
+    assert ranges[args.model][1] >= args.temperature >= ranges[args.model][0], (
+      f'--temperature must be within {ranges[args.model][0]} and '
+      f'{ranges[args.model][1]}.')
```
DonggeLiu, line 284: Add `... for model args.model` specify the model name we parsed.

DonggeLiu (commented):
Thanks @dberardi99, some nits.

llm_toolkit/models.py
DonggeLiu: Feel free to adjust/simplify these functions as you see fit, particular the filtering logic.
These above are just examples.

dberardi99: Good idea! I've slightly modified it by introducing the `all_llm_search` method, which allows to output the desired attribute for all LLMs. If you don't like the idea, I can switch back to your implementation.

dberardi99:
@DonggeLiu I'm so sorry for the great delay... If you like this final implementation I can open a new PR creating a new branch for this fix in order to leave the modification history clearer

llm_toolkit/models.py
```diff
@@ -116,15 +119,31 @@ def all_llm_subclasses(cls):
...
+    return cls.all_llm_search('name')
+  
+  @classmethod
+  def all_llm_temperature_ranges(cls) -> dict[str, list[float, float]]:
+    """Returns the current model and all child temperature ranges."""
+    out = {}
+    names = cls.all_llm_search('name')
+    tr = cls.all_llm_search('temperature_range')
+    for i in range(len(names)):
+      out[names[i]] = tr[i]
```
DonggeLiu, line 145: `tr` and `names` may mismatch.
This code assumes that `all_llm_subclasses` will always return the subclasses in the same order, but I don't think Python's `__subclasses__` can guarantee that.

DonggeLiu (commented):
Thanks @dberardi99.
This is already pretty good, just a final question regarding reliability.

I don't think you have to create a new PR, as the modification history is pretty clear. But please let me know if you prefer otherwise.

Also, as your code is almost ready to merge, I will create a new base branch for your PR so that we can experiment with it before merging into main.

llm_toolkit/models.py
dberardi99: I had the same your doubt regarding the order in which Python returns subclasses. I can make some research and in case we are unsure about it propose a small fix to increase reliability

dberardi99:
> Thanks @dberardi99. This is already pretty good, just a final question regarding reliability.
> 
> I don't think you have to create a new PR, as the modification history is pretty clear. But please let me know if you prefer otherwise.
> 
> Also, as your code is almost ready to merge, I will create a new base (#986) for your PR so that we can experiment with it before merging into main.

Awesome! We can continue working in this PR with no problem in this case

llm_toolkit/models.py
DonggeLiu: How about using dictionaries like we discussed earlier?
https://github.com/google/oss-fuzz-gen/pull/856#discussion_r1997339660

dberardi99: I've slightly adapted the filtering function to return also all the models names, so that now we guarantee the correct assignment between model name and temperature range

PR #859, opened 2025-03-13
@ay-bh: Fix benchmark status reporting to accurately reflect trial completion

### LINKED ISSUE #721: Incorrect Benchmark Status on report's Index Page

DonggeLiu:
Run experiments for #847, contributed by @ay-bh.
Fix #721.


DonggeLiu:
https://llm-exp.oss-fuzz.com/Result-reports/ofg-pr/2025-03-14-859-ab-comparison/index.html

DonggeLiu:
@erfanio Does this look good to you?
Thanks!

run_all_experiments.py
```diff
@@ -535,6 +535,9 @@ def main():
   _setup_logging(args.log_level, is_cloud=args.cloud_experiment_name != '')
   logger.info('Starting experiments on PR branch')
 
+  # Store the trial count in an environment variable
+  os.environ['BENCHMARK_TRIAL_COUNT'] = str(args.num_samples)
```
erfanio, line 539: tbh I'm not a fan of using environment variables to store this.

I have two alternate suggestions:
* Pass num-samples to web.py via cli flags. Modify docker_run and upload_report accordingly.
* Add num_samples  to report.json and retrieve it in web.py just like timing data.

temp
erfanio, line 1: Not sure if this file is intentional.

erfanio (commented):
I'm not super opposed to using environment variables so if you'd like, I'm happy to submit this but I think the other ways are clearer for everyone trying to trace the code.

run_all_experiments.py
DonggeLiu: Oh these two are definitely better, particularly adding `num_samples` to `report.json`.
@ay-bh could you please add that? Thanks!

temp
DonggeLiu: Nope, it's just help me to create this PR.

run_all_experiments.py
ay-bh: @DonggeLiu sure!

ay-bh: @erfanio we actually need num_samples in `common.py` for the `_get_expected_trials_count` method, why do we need to retrieve it in `web.py`?
Thanks

DonggeLiu:
This works as expected now : )

DonggeLiu:
TBR @erfanio

PR #864, opened 2025-03-14
Unify cached images

DonggeLiu:
Unify the way to use cached images when building the fuzz targets in agents and Execution stage.

### DIFF
--- a/experiment/oss_fuzz_checkout.py
+++ b/experiment/oss_fuzz_checkout.py
@@ -489,7 +489,22 @@ def prepare_project_image(project: str) -> str:
     rewrite_project_to_cached_project(project, project, 'address')
     # Prepare build
     prepare_build(project, 'address', project)
-    logger.info('Using cached project image for %s', project)
-    return image_name
+    # Build the image
+    command = [
+        'docker', 'build', '-t', image_name,
+        os.path.join('projects', project)
+    ]
+    try:
+      sp.run(command,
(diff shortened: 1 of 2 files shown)


experiment/oss_fuzz_checkout.py
```diff
@@ -483,10 +483,28 @@ def _build_image(project_name: str) -> str:
 def prepare_project_image(project: str) -> str:
   """Prepares original image of the |project|'s fuzz target build container."""
   image_name = f'gcr.io/oss-fuzz/{project}'
-  if (ENABLE_CACHING and is_image_cached(project, 'address') and
-      (_image_exists_locally(image_name, project_name=project) or
-       _image_exists_online(image_name, project_name=project))):
-    logger.info('Using cached project image for %s', project)
-    return image_name
+  if ENABLE_CACHING and is_image_cached(project, 'address'):
```
oliverchang, line 486: should we move `prepare_project_image` into `container_tool`? container_tool seems to be the only user, and it's very specific to that.

oliverchang (approved):

experiment/oss_fuzz_checkout.py
DonggeLiu: Logically I consider projects images as a feature provided by OSS-Fuzz, `contain_tool` uses this feature and focuses on executing commands / parsing outputs instead of handling images.
We can refine this and find the best solution when working on the GSoC project : )

PR #868, opened 2025-03-14
fix aggregate coverage analysis

DavidKorczynski:
Several fixes:

- coverage process was not started in local runs
- typo causing no coverage collection to be done `== rust` should be `!= rust` (see diff)
- we were deleting `benchmark.yaml` from the workdirs in the new prototypers, which is used by the coverage collection to identify projects/benchmarks. The `benchmark.yaml` is created early in the experimental process and should not be deleted in proceeding logic.


DavidKorczynski:
/gcbrun exp -n davidzxx12 -m vertex_ai_gemini-1-5 -b comparison

DavidKorczynski:
/gcbrun exp -n davidavid-1 -ag -m vertex_ai_gemini-2-think-chat

DonggeLiu:
Noice! Thanks David!

DavidKorczynski:
The agent workflow is having some issues, will need to debug

DavidKorczynski:
/gcbrun exp -n dg-actually-david-1 -ag -mr 100

DavidKorczynski:
agent is looking good as well. This is good to go.

PR #878, opened 2025-03-15
logger: make sure to overwrite harness instead of appending

DavidKorczynski:
This is causing issues where we have multiple harnesses in a single *.fuzz_target

### DIFF
--- a/logger.py
+++ b/logger.py
@@ -37,22 +37,25 @@ class CustomLoggerAdapter(logging.LoggerAdapter):
     kwargs['extra'] = {**(self.extra or {}), **(kwargs.get('extra') or {})}
     return msg, kwargs
 
-  def write_to_file(self, file_path: str, file_content: str) -> None:
+  def write_to_file(self,
+                    file_path: str,
+                    file_content: str,
+                    mode: str = 'a') -> None:
     """Writes the |file_content| into a local |file_path|."""
-    with open(file_path, 'a') as file:
+    with open(file_path, mode) as file:
       file.writelines(file_content)
 
   def write_fuzz_target(self, result: Result) -> None:
     """Writes fuzz target."""
     fuzz_target_path = os.path.join(result.work_dirs.fuzz_targets,
                                     f'{result.trial:02d}.fuzz_target')
-    self.write_to_file(fuzz_target_path, result.fuzz_target_source)
+    self.write_to_file(fuzz_target_path, result.fuzz_target_source, 'w')
 
   def write_build_script(self, result: Result) -> None:
     """Writes build script."""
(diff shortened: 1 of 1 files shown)


DavidKorczynski:
Using the one prompt prototyper, before this:

Screenshot 2025-03-15 172732


After this:
Screenshot 2025-03-15 172838

DavidKorczynski:
/gcbrun exp -n davidzxx16 -m vertex_ai_gemini-1-5 -b comparison

DavidKorczynski:
Experiment looks good: https://llm-exp.oss-fuzz.com/Result-reports/ofg-pr/2025-03-16-878-davidzxx16-comparison/index.html

DavidKorczynski:
/gcbrun exp -n davidzxx17 -ag -m vertex_ai_gemini-1-5 -b comparison

DavidKorczynski:
Experiment is looking good:  https://llm-exp.oss-fuzz.com/Result-reports/ofg-pr/2025-03-16-878-davidzxx17-comparison/index.html

DonggeLiu (approved):
Good catch, thanks David!

DonggeLiu:
In future PRs, we could save all fuzz targets and their corresponding build scripts in separate files, such as `Trial-01.Cycle-01.fuzz_target`.

This approach ensures that we retain every iteration of the fuzz target, as the final version may not always be the most interesting or valuable. For example, a previous fuzz target might have triggered a crash that revealed a real vulnerability in the project. However, if the Analyzer agent incorrectly classified the crash as a false positive, the Enhancer agent could have "fixed" the fuzz target in the next cycle, accidently suppressing the crash.

By storing each iteration separately, we can preserve valuable insights, track changes over time, and revisit earlier versions for debugging or further analysis.

DavidKorczynski:
> In future PRs, we could save all fuzz targets and their corresponding build scripts in separate files, such as `Trial-01.Cycle-01.fuzz_target`.
> 
> This approach ensures that we retain every iteration of the fuzz target, as the final version may not always be the most interesting or valuable. For example, a previous fuzz target might have triggered a crash that revealed a real vulnerability in the project. However, if the Analyzer agent incorrectly classified the crash as a false positive, the Enhancer agent could have "fixed" the fuzz target in the next cycle, accidently suppressing the crash.
> 
> By storing each iteration separately, we can preserve valuable insights, track changes over time, and revisit earlier versions for debugging or further analysis.

Yeah, I did think this approach makes more sense as well when hunting down this issue. It would enable us also to have some post processing analysis that looks at the entire history to determine which harness was actually the best one, as opposed to having a need for an approach that determined at a given stage if "this is a good enough harness".

PR #867, opened 2025-03-14
Add line numbers to harness code in reports

Kartikayy007:
# Add line numbers to harness code in reports

Fixes #579

## Description
This PR adds line numbers to code blocks in the OSS-Fuzz-gen reports, making it easier to reference specific lines when triaging issues. When debugging crashes, stack traces often reference line numbers, and this change allows for easier correlation between stack traces and the displayed code.

## Implementation Details
- Added a new `_add_line_numbers` filter function to `JinjaEnv` class in `report/web.py`
- Applied the filter to code blocks in `sample.html` template
- Line numbers are right-aligned with the proper width based on total number of lines

## Testing
Verified the implementation works with different code lengths and content types
Ensured line numbers are properly aligned and formatted

## Screenshots
Earlier: 
Screenshot 2025-03-14 at 4 48 26 PM

Updated:

Screenshot 2025-03-14 at 4 48 43 PM

PR #881, opened 2025-03-16
Add syntax highlighting to code blocks in reports

Kartikayy007:
This PR enhances the report UI by adding syntax highlighting to both fuzz target code and build script code:
reference: https://github.com/google/oss-fuzz-gen/pull/867#issuecomment-2727001909
CC: @DonggeLiu 

- **Fuzz Target Code**: Applies language-specific highlighting based on benchmark YAML language field.
- **Build Scripts**: Used Bash syntax highlighting for all build scripts.
- **Separation of Code Sections**: Properly separates build script code from fixer prompts.

## Implementation Details

- Added `highlight.js` library to `base.html`
- Enhanced `Benchmark` class with language field and extraction from YAML files
- Updated `Target` class to separate build script code from fixer prompts
- Modified `templates` to apply appropriate syntax highlighting

## Testing

- Changes have been tested with a standalone test page that demonstrates syntax highlighting for different programming languages.
- The syntax highlighting applies correctly to all dynamically generated report pages.

## Screenshots
Screenshot 2025-03-17 at 2 57 32 AM
Screenshot 2025-03-17 at 2 57 40 AM
Screenshot 2025-03-17 at 2 57 50 AM

### DIFF
--- a/report/common.py
+++ b/report/common.py
@@ -24,6 +24,7 @@ from typing import List, Optional
 
 import yaml
 from google.cloud import storage
+import glob
 
 import run_one_experiment
 from data_prep import project_src
@@ -73,6 +74,7 @@ class Benchmark:
   signature: str = ''
   project: str = ''
   function: str = ''
+  language: str = ''
 
 
 @dataclasses.dataclass
@@ -129,6 +131,7 @@ class Sample:
 class Target:
   code: str
   fixer_prompt: Optional[str] = None
+  build_script_code: Optional[str] = None
 
 
 @dataclasses.dataclass
@@ -411,7 +414,7 @@ class Results:
         build_script_code = f.read()
 
     # TODO(dongge): Properly show build script code in reports.
-    return Target(code=fuzz_target_code, fixer_prompt=build_script_code)
+    return Target(code=fuzz_target_code, fixer_prompt=None, build_script_code=build_script_code)
 
   def get_samples(self, results: list[evaluator.Result],
                   targets: list[str]) -> list[Sample]:
@@ -638,7 +641,8 @@ class Results:
     function = benchmark_id.split('-')[-1]
     signature = self._find_benchmark_signature(project,
                                                function) or benchmark_id
-    return Benchmark(benchmark_id, status, result, signature, project, function)
+    language = self._find_benchmark_language(project, function)
+    return Benchmark(benchmark_id, status, result, signature, project, function, language)
 
   def _find_benchmark_signature(self, project: str,
                                 target_function: str) -> str:
@@ -670,6 +674,25 @@ class Results:
 
     return matched_prefix_signature
 
+  def _find_benchmark_language(self, project: str, target_function: str) -> str:
+    """Finds the programming language of the benchmark."""
+    if not self._benchmark_dir:
+      return ''
+    
+    # looks for a YAML file with the project name
+    benchmark_glob = os.path.join(self._benchmark_dir, '*', f'{project}.yaml')
+    benchmark_files = glob.glob(benchmark_glob)
+    
+    if not benchmark_files:
+      return ''
+    
+    try:
+      with open(benchmark_files[0], 'r') as f:
+        benchmark_data = yaml.safe_load(f)
+        return benchmark_data.get('language', '')
+    except Exception as e:
+      logging.error('Failed to read benchmark file %s: %s', benchmark_files[0], e)
+      return ''
 
 def _parse_log_parts(log: str) -> list[LogPart]:
   """Parse log into parts."""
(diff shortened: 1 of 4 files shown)


Kartikayy007:
@DonggeLiu please review whenever you get time, Thank You!

report/common.py
```diff
@@ -670,6 +674,25 @@ def _find_benchmark_signature(self, project: str,
...
     return matched_prefix_signature
 
+  def _find_benchmark_language(self, project: str, target_function: str) -> str:
+    """Finds the programming language of the benchmark."""
+    if not self._benchmark_dir:
+      return ''
+    
+    # looks for a YAML file with the project name
+    benchmark_glob = os.path.join(self._benchmark_dir, '*', f'{project}.yaml')
+    benchmark_files = glob.glob(benchmark_glob)
```
DonggeLiu, line 684: Could this be simplified?
From a few lines above:
https://github.com/google/oss-fuzz-gen/blob/2c539eb617e876bd8b28432aa27604b13e41ceef/report/common.py#L647-L652

DonggeLiu: BTW, another way to determine the fuzz target language is by checking the target_path in the YAML file.
For example:
https://github.com/google/oss-fuzz-gen/blob/d505a1fa536fb97c361adcaae0494986804796d1/benchmark-sets/all/abseil-cpp.yaml#L51
This function can be useful too:
https://github.com/google/oss-fuzz-gen/blob/d505a1fa536fb97c361adcaae0494986804796d1/experiment/benchmark.py#L231-L234


Feel free to choose the approach that you find most suitable.

DonggeLiu (commented):
Thanks for addressing this so quickly, @Kartikayy007!
You have done an amazing job.

Some minor nits before we merge and experiment more with your PR:

DonggeLiu:
Once you are ready, we can merge your PR to #883 for some final experiments.

Kartikayy007:
Hey, @DonggeLiu  its ready for another review

report/common.py
Kartikayy007: I Thought to use  combined approach that incorporates both suggestions, firstly use _find_benchmark_signature to simplify the path lookup and use checking target_path in the YAML as a fallback system, but I didn't used the fallback system because it would add significant complexity to the code with numerous conditional statements

```
# Fallback to target_path
target_path = benchmark_data.get('target_path', '')
if target_path:
    file_ext = os.path.splitext(target_path)[1].lower()
    if file_ext == '.c':
        return 'c'
    elif file_ext in ['.cc', '.cpp', '.cxx', '.c++', '.h', '.hpp']:
        return 'c++'
    elif file_ext == '.java':
        return 'jvm'
    elif file_ext == '.rs':
        return 'rust'
    elif file_ext == '.py':
        return 'python'
    # Add Further mapping........
```
ill proceed further as you prefer Thank you!

report/common.py
```diff
@@ -24,6 +24,7 @@
 
 import yaml
 from google.cloud import storage
+import glob
```
DonggeLiu, line 27: I presume this is no longer needed?
The [presubmit script](https://github.com/google/oss-fuzz-gen/blob/df3e87c9b139f7a17a0a4ca237752961f5f32223/USAGE.md#contribution-process) may help you finding issues like this : )

Kartikayy007: yup that's not needed, I'll surely use [presubmit script](https://github.com/google/oss-fuzz-gen/blob/df3e87c9b139f7a17a0a4ca237752961f5f32223/USAGE.md#contribution-process) in future before raising PR

Kartikayy007:
@DonggeLiu I've updated as per changes.

experimental/c-cpp/manager.py
```diff
@@ -1084,6 +1084,9 @@ def run_introspector_on_dir(build_worker, test_dir,
   modified_env['FUZZ_INTROSPECTOR_AUTO_FUZZ'] = '1'
   modified_env['PROJECT_NAME'] = 'auto-fuzz-proj'
   modified_env['FUZZINTRO_OUTDIR'] = test_dir
+
+  # Disable FI light because we want to make sure we can compile as well.
+  modified_env['FI_DISABLE_LIGHT'] = "1"
```
DonggeLiu, line 1089: Weird that this shows up in your PR, but it should not bother us after we merge yours into mine.

DonggeLiu (approved):

experimental/c-cpp/manager.py
Kartikayy007: that was not by me It was by @DavidKorczynski in [commit](https://github.com/google/oss-fuzz-gen/pull/881/commits/df3e87c9b139f7a17a0a4ca237752961f5f32223) as the message suggest: Will make it faster to generate empty builds. was done to make generate fast empty builds

DonggeLiu: I know : )

Issue #862, opened 2025-03-13
Add local development hot-reloading for experiment reports


myanvoos:
To my understanding, right now while an experiment is running locally with `./run-all-experiments.py`, the results go in the `results` directory (or as specified by the user) whenever they're available. The `results-report` directory is only updated after we manually run `python -m report.web -r results -o results-report`. This means that currently there's no automatic refresh/update of the report while the experiment is still running locally.

It might be good to add a functionality that watches the `results` directory for any new finished experiment, then automatically updates `results-report` (or equivalent) and hot-reloads the HTTP server accordingly. So you can run the local HTTP server alongside experiments. Perhaps this can be triggered with a `--watch` flag on `./run_all_experiments.py` or on `python -m report.web -r results -o results-report`. 

Discussion much appreciated :)  I'll have an experimental PR for demonstration up later this week.

myanvoos:
A small note: There's a `upload_report.sh` script that runs as a subprocess of `docker_run.sh` that basically does a similar thing, but these scripts don't seem to be triggered in a basic local workflow like

```
./run_all_experiments.py \
    --model=<model-name> \
    --benchmarks-directory='./benchmark-sets/comparison'
```

Karanjot786:
Hi @myanvoos, I'm very interested in working on this feature. I agree that having automatic hot-reloading of the experiment reports would greatly improve the local development workflow, especially when running experiments with `./run_all_experiments.py`. My plan is to add a `--watch flag `(or similar) that watches the results and report directories and automatically updates the results-report along with hot-reloading the HTTP server.

Could you please assign this issue to me so I can start working on it? Thanks!

Vinay152003:
Hi everyone,

I've submitted PR [#869](https://github.com/google/oss-fuzz-gen/pull/869) which implements the hot-reloading functionality for experiment reports. This PR introduces a new --watch flag to run_all_experiments.py that uses Python's watchdog library to monitor the results directory. When changes are detected, it automatically triggers an update of the results-report directory by running the report generation command.

Thank you for the discussion and feedback so far. I'm looking forward to your review and am happy to make further adjustments if needed!

myanvoos:
Hi @Vinay152003 and @Karanjot786 ,

[...]

PR #884, opened 2025-03-17
Extend web.py and add local hot reload

myanvoos:
Related discussion #862 

This PR extends the command line parser in [`web.py`](https://github.com/google/oss-fuzz-gen/blob/main/report/web.py) to take in some additional inputs as outlined [here](https://github.com/google/oss-fuzz-gen/issues/862#issuecomment-2727160204). It also adds an optional server-side hot-reloading functionality with the `watchdog` library. 

Happy to iterate upon this @DonggeLiu!

### DIFF
--- a/USAGE.md
+++ b/USAGE.md
@@ -171,10 +171,13 @@ python -m report.web -r <results-dir> -o <output-dir>
 python -m http.server <port> -d <output-dir>
 ```
 Where `<results-dir>` is the directory passed to `--work-dir` in your
-experiments (default value `./results`).
+experiments (default value `./results`). 
 
 Then navigate to `http://localhost:<port>` to view the result in [a table](#result-table).
 
+> Tip:
+You can also pass `-w` to watch for changes in `results-dir` and automatically
+update the results report. Refresh the page to see the updated report.
 
 ## Detailed workflows
 Configure and use framework in the following steps:
--- a/report/web.py
+++ b/report/web.py
@@ -28,7 +28,6 @@ from typing import Any, Dict, List, Optional
 
 from watchdog.observers import Observer
 from watchdog.events import FileSystemEventHandler
-import subprocess
 import threading
 
 
@@ -263,10 +262,12 @@ class ReportWatcher(FileSystemEventHandler):
     self.server = None  
     
     if args.watch_filesystem:
+      logging.info(f"Watching filesystem for changes in {args.results_dir}")
       self.observer.schedule(self, args.results_dir, recursive=True)
         
     if args.watch_template:
-      self.observer.schedule(self, "report/", recursive=True)
+      logging.info(f"DEV: Watching for changes in the report folder")
+      self.observer.schedule(self, "report", recursive=True)
 
     if args.serve:
       self.server_thread = threading.Thread(target=self._start_server)
@@ -292,6 +293,7 @@ class ReportWatcher(FileSystemEventHandler):
     self.observer.join()
 
   def on_modified(self, event):
+    """Restart the server when the watcher detects a change"""
     logging.info(f"{event.src_path} has been modified. Regenerating report...")
     generate_report(self.args)
     
--- a/report/web.py
+++ b/report/web.py
@@ -26,6 +26,12 @@ from functools import partial
 from http.server import SimpleHTTPRequestHandler, ThreadingHTTPServer
 from typing import Any, Dict, List, Optional
 
+from watchdog.observers import Observer
+from watchdog.events import FileSystemEventHandler
+import subprocess
+import threading
+
+
 import jinja2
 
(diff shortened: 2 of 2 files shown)


USAGE.md
```diff
@@ -171,10 +171,13 @@ python -m report.web -r <results-dir> -o <output-dir>
 python -m http.server <port> -d <output-dir>
 ```
 Where `<results-dir>` is the directory passed to `--work-dir` in your
-experiments (default value `./results`).
+experiments (default value `./results`).
```
DonggeLiu, line 174: nit: trailing space.

report/web.py
```diff
@@ -301,27 +361,46 @@ def _parse_arguments() -> argparse.Namespace:
...
   return parser.parse_args()
 
 
 def main():
-  args = _parse_arguments()
-
-  if not args.serve:
-    generate_report(args)
-  else:
+    args = _parse_arguments()
```
DonggeLiu, line 382: nit: Mismatched indentation. 2-spaces indentation is used by default in this repo.
https://github.com/google/oss-fuzz-gen/blob/main/USAGE.md#contribution-process
This can help you discover similar problems : )

report/web.py
```diff
@@ -246,6 +251,61 @@ def _write_benchmark_sample(self, benchmark: Benchmark, sample: Sample,
...
+    
+    if args.watch_filesystem:
+      logging.info(f"Watching filesystem for changes in {args.results_dir}")
+      self.observer.schedule(self, args.results_dir, recursive=True)
+        
+    if args.watch_template:
+      logging.info(f"DEV: Watching for changes in the report folder")
+      self.observer.schedule(self, "report", recursive=True)
+
+    if args.serve:
```
DonggeLiu, line 272: Where is `serve` set?

report/web.py
```diff
@@ -301,27 +361,46 @@ def _parse_arguments() -> argparse.Namespace:
                       help='Port to launch webserver on.',
                       type=int,
                       default=8012)
+  parser.add_argument('--interval-seconds',
+                      '-i',
+                      help='Interval in seconds to generate report.',
+                      type=int,
+                      default=600)
+  parser.add_argument('--watch-filesystem',
+                      '-w',
```
DonggeLiu, line 370: nit: `-wf` for `--watch-filesystem` and `-wt` for `--watch-template`?

report/web.py
```diff
@@ -301,27 +361,46 @@ def _parse_arguments() -> argparse.Namespace:
...
+        should_continue = args.serve or args.watch_filesystem or args.watch_template
+        
+        while should_continue:
+            generate_report(args)
+            # If interval is specified, wait and regenerate
+            if args.interval_seconds > 0:
+                time.sleep(args.interval_seconds)
+            # If only watching filesystem and no interval, just wait
+            else:
+                time.sleep(1)
```
DonggeLiu, line 398: 1. I reckon we can simplify
```python
            if args.interval_seconds > 0:
                time.sleep(args.interval_seconds)
            # If only watching filesystem and no interval, just wait
            else:
                time.sleep(1)
```
to
```python
time.sleep(args.interval_seconds)
```
if we set
```python
  parser.add_argument('--interval-seconds',
                      '-i',
                      help='Interval in seconds to generate report.',
                      type=int,
                      default=90)
```

2. 1 second is too short, many report generation takes longer. let's use the old default value 90

DonggeLiu (commented):
Thanks @myanvoos!

report/web.py
myanvoos: It was predefined in [`_parse_arguments()`](https://github.com/google/oss-fuzz-gen/blob/main/report/web.py#L295) as a handy flag to trigger the HTTP server right away:

```py
parser.add_argument('--serve',
                      '-s',
                      help='Will launch a web server if set.',
                      action='store_true')
```

It was then passed to `ReportWatcher` in `main()` via `args`:

```py
def main():
  args = _parse_arguments()
  logging.getLogger().setLevel(os.environ.get('LOGLEVEL', 'INFO').upper())

  watcher = ReportWatcher(args)
```

report/web.py
myanvoos: Yep, to be honest that would make more sense. On it!

report/web.py
myanvoos: Agreed, I guess that's a bit redundant. And 90 it is :)

report/web.py
myanvoos: This as well as the string formatting in logs should be resolved now :)

myanvoos:
Whenever you're available @DonggeLiu!

DonggeLiu:
LGTM now.
I noticed some legacy `"` that may have confused you (not your fault) so I unified them to use `'` by default.

I will create another PR to run some experiments, just to be safe.

PR #887, opened 2025-03-17
Store trial count in report.json instead of environment variable

ay-bh:
### Changes

These changes are based on the discussion in #859.

- Removed environment variable for storing `num_samples` from `run_all_experiments.py`
- Added trial count directly to `report.json` using existing `add_to_json_report` function
- Modified `_get_expected_trials_count()` to first check `report.json` for `num_samples`

### DIFF
--- a/report/common.py
+++ b/report/common.py
@@ -284,11 +284,16 @@ class Results:
 
   def _get_expected_trials_count(self, benchmark_id: str) -> int:
     """Returns the expected number of trials for a benchmark."""
-    env_trial_count = os.environ.get('BENCHMARK_TRIAL_COUNT')
-    if env_trial_count:
+    # First try to get the value from report.json
+    report_path = os.path.join(self._results_dir, 'report.json')
+    if os.path.exists(report_path):
       try:
-        return int(env_trial_count)
-      except ValueError:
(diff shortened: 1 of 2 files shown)


report/common.py
```diff
@@ -284,11 +284,16 @@ def match_benchmark(self, benchmark_id: str, results: list[evaluator.Result],
...
       try:
-        return int(env_trial_count)
-      except ValueError:
+        with open(report_path, 'r') as f:
+          report_data = json.load(f)
+          if 'num_samples' in report_data:
+            return report_data['num_samples']
+      except (json.JSONDecodeError, OSError):
+        # If there's an error reading the file, continue to fallbacks
         pass
```
DonggeLiu, line 297: nit: avoid too many indentation levels
```python
try:
    with open(report_path) as f:
        num_samples = json.load(f).get('num_samples')
    if num_samples is not None:
        return num_samples
except (FileNotFoundError, json.JSONDecodeError, OSError):
    pass
```

DonggeLiu (approved):

report/common.py
ay-bh: Thanks @DonggeLiu. Done.

PR #899, opened 2025-03-20
JVM: Fix wrong use of OpenAI prompt priming and problem 

arthurscchan:
This PR fixes the wrong use of OpenAI prompt priming and problem role for JVM projects. It is understood that, in a chat session of OpenAi, role `System` is used for setting the base priming, role `user` is used to store the user prompt and role `assistant` is used to store the previous reply from OpenAi. Each dictionary in the messages list is act as the full trace of chat history. This PR fixes the JVM (and Rust/Python in later PR) approach on the wrong use of these roles in the OpenAi prompts.


arthurscchan:
/gcbrun exp -n ascc-pr-899-1 -m vertex_ai_gemini-1-5 -i -b jvm-small

DavidKorczynski:
Experiment looks good: https://llm-exp.oss-fuzz.com/Result-reports/ofg-pr/2025-03-20-899-ascc-pr-899-1-jvm-small/index.html

DavidKorczynski (approved):

Issue #908, opened 2025-03-21
Add Logging and Error Handling for get_benchmark_final_target_code


Ekam219:
#### **Description**
@DonggeLiu 
While PR [#905](link) addresses JSON syntax issues in `crash.json`, the root cause of missing `source_code` fields—such as silent failures in `get_benchmark_final_target_code()`—remains unresolved. This issue focuses on adding logging and improving error handling to diagnose and mitigate such failures effectively.

---

#### 📝 **Enhanced Logging**
- Add logs to identify when `source_code` is missing and the reason behind it.
- Log invalid `sample_id` formats (e.g., cases where `sample_id` cannot be split into `benchmark/sample`).
- Capture errors related to missing source code files.

#### ⚡ **Improved Error Handling**
- Ensure `get_benchmark_final_target_code()` handles malformed or empty `sample_id` gracefully.
- Return `null` explicitly instead of empty strings in JSON outputs to improve clarity.
- Implement fallback mechanisms for missing or malformed data.

---

### **Why This Matters**
✅ **Improved Debuggability** – Logs will help identify whether `source_code` is missing due to experiment failures or internal code issues.  
✅ **Increased Robustness** – Ensures graceful handling of errors instead of silent failures, reducing invalid JSON outputs.

---

Issue #625, opened 2024-09-18
missing error logs for failed build


DavidKorczynski:
Some fixer prompts dont have anything in the error tag: https://llm-exp.oss-fuzz.com/Result-reports/ofg-pr/2024-09-18-618-ochang-extra-large-all-oracles-many/sample/output-mdbtools-mdb_insert_row/03.html

PR #910, opened 2025-03-21
Fix Logging and Error Handling for get_benchmark_final_target_code

Ekam219:
## Description
This PR addresses the issue of missing source code by adding enhanced logging and improving error handling in the `get_benchmark_final_target_code()` method. Previously, silent failures could occur, making it difficult to diagnose the root cause of missing source code.

## Changes
- **Enhanced Logging**: Logs are added to track when the `sample_id` is invalid or the source code is missing.  
- **Error Handling**: Ensures that malformed or empty `sample_id` values are handled gracefully and logged appropriately.  
- **Return Null for Missing Source Code**: In case of an error, the function explicitly returns `None` rather than an empty string for clarity in JSON outputs.  
#908
It may also refer to https://github.com/google/oss-fuzz-gen/issues/625
## Benefits
- Improved **debuggability** by making it easier to identify the reasons behind missing source code.  
- Increased **robustness** by preventing silent failures and ensuring more informative logs and error handling.

### DIFF
--- a/report/common.py
+++ b/report/common.py
@@ -310,7 +310,7 @@ class Results:
     targets_dir = os.path.join(self._results_dir, benchmark, 'fixed_targets')
     # TODO(donggeliu): Make this consistent with agent output.
     if not os.path.exists(targets_dir):
-      return ''
+      return None
 
     for name in sorted(FileSystem(targets_dir).listdir()):
       path = os.path.join(targets_dir, name)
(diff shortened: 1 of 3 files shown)


DonggeLiu:
+@Kartikayy007 for visibility.
I will create another experiment PR so that we can compare the result of these two methods.
Thanks agin to both of you!

DonggeLiu:
Experiment launched in #919

PR #916, opened 2025-03-22
oss_fuzz_checkout: relax constraints on venv check

DavidKorczynski:
Sometimes it's useful to use a virtual environment that is not used in the given path. If the VIRTUAL_ENV environment is set, then we can assume we are in a correct virtual environment.

PR #921, opened 2025-03-23
Revert "oss_fuzz_checkout: relax constraints on venv check (#916)"

DonggeLiu:
This reverts commit aaaf0f8ef1e5202d0e8190089a6d7219b9ce82ac to identify root cause of recent bugs that affects all experiments.

### DIFF
--- a/experiment/oss_fuzz_checkout.py
+++ b/experiment/oss_fuzz_checkout.py
@@ -119,7 +119,7 @@ def postprocess_oss_fuzz() -> None:
 
   # If already in a virtualenv environment assume all is set up
   venv_path = os.path.split(os.environ.get('VIRTUAL_ENV', ''))
-  if venv_path:
+  if venv_path and venv_path[0].endswith(os.path.split(OSS_FUZZ_DIR)[-1]):
     return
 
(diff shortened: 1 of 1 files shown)


DonggeLiu:
Report is back to normal now:
https://llm-exp.oss-fuzz.com/Result-reports/ofg-pr/2025-03-24-921-dg-comparison/index.html

DonggeLiu:
TBR.
I will merge this for now to unblock some PRs waiting for experiments.

DavidKorczynski:
hmm, it's a bit suboptimal `and venv_path[0].endswith(os.path.split(OSS_FUZZ_DIR)[-1]` needs to be satisfied. Do you know the rootcause behind this?

DonggeLiu:
> thanks for fixing.
> 
> Hmm, it's a bit suboptimal `and venv_path[0].endswith(os.path.split(OSS_FUZZ_DIR)[-1]` needs to be satisfied. Do you know the rootcause behind this?

I reckon we need this specific condition because we might be in one of two virtual environments: OSS-Fuzz or OSS-Fuzz-Gen.
If I recall correctly, the failure happened when an agent running in the OSS-Fuzz-Gen venv tried to execute OSS-Fuzz code, which led to an error.

PR #883, opened 2025-03-17
@Kartikayy007: Add syntax highlighting to code blocks in reports and fix and the Outdated section name

DonggeLiu:
Run experiment for #881 before merging.

PR #926, opened 2025-03-24
Fix exp 881

Kartikayy007:
# Fix Sample Page Rendering Issues

## Problem
mentioned in #883 
Sample pages were failing to render with these errors:
image

The template was using a filter called `splitlines` which wasn't registered in the Jinja2 environment. Additionally, the template also lacking robustness against undefined variables.

## Files Changed for #883 
1. `report/web.py` - Added missing filter and improved error handling
2. `report/templates/sample.html` - Added robustness checks 

## Changes Made
1. **Added missing filter:**
   - Implemented `_splitlines` method in `web.py`
   - Registered it with Jinja2 environment

2. **Improved template robustness:**
   - Added `is defined` checks for variables that might be missing
   - Added default values with `|default()` filter
   - Made `remove_trailing_empty_lines` filter handle edge cases better

3. **Enhanced error handling:**
   - Added detailed error logging with full traceback
   - Created fallback error pages when rendering fails
   - Provided better default values for missing data

These improvements should resolve all the sample page rendering issues.


Kartikayy007:
CC: @DonggeLiu

DonggeLiu:
Thanks! BTW there is no need to replicate everything in a new PR:
If you set the base branch to be `exp-881`, then we can merge this back to exp-881 and run experiments there : )

Issue #134, opened 2024-03-04
Save example function signature in cloud buckets.


DonggeLiu:
Currently, we only save existing human-written fuzz targets in[ the cloud bucket](https://pantheon.corp.google.com/storage/browser/oss-fuzz-llm-public/human_written_targets?project=oss-fuzz) as example solutions.
The corresponding example questions (i.e., signature of the function used in the fuzz targets) are parsed from FI report everytime by the user.

This is problematic because:
1. We cannot access them when the FI report was not available, and 
2. It reports distracting errors.

Ideally solution:
1. Add a JSON that maps each fuzz target to a list of signatures of the functions used in the fuzz target.
2. Restructure the code to take function signatures from the bucket and only generate them at runtime if not in the bucket.

sudhin05:
Hi, are there any updates on this? If possible, I'd like to work on this issue.

sudhin05:
Here's my broader understanding of the issue so far @DonggeLiu, 

The framework adds existing human-written fuzz targets as examples into the prompt to get more project specific context for the LLM. 

The generation of the examples is handled using the `generate_data()` function in the `project_targets.py` file which downloads the human-written fuzz targets from Google Cloud Bucket.

(Here on a side note, the current design indirectly validates the Cloud Bucket downloads via `target_content_signature_dict`, which may confuse post-download processing failures with download failures. Would it be better to verify downloads explicitly in `_get_fuzz_target_dir()`? This would help to isolate download & processing issues)

The signatures are retrieved from the FuzzIntrospector Report through a querying function `query_introspector_function_signature()` that queries the FI API for the signature for a given function name. It gets called in the `_bucket_match_target_content_signatures()` function. Here the possibility arises, that due to unavailability of FI Report, parsing issues or timeouts we may get unexpected errors and delays.

For the solution to this, I intend to create helper functions to load and update the signature mapping from GC Bucket and add caching logic for the function signatures in a JSON file at the beginning of `_bucket_match_target_content_signatures()` so that it first attempts to load the cached mapping from the bucket. Then for each target content (the corresponding key in the dictionary), check if there is already a list of signatures cached. If so, use it and if not, compute it using the existing introspector calls and then update the mapping. As for the fallback logic in case if the function returns an empty dict, it already falls back to _match_target_content_signatures(), so it will remain as it is.

DonggeLiu:
Thanks @sudhin05 !
Your plan and explanation are solid overall.

[...]

PR #940, opened 2025-03-27
Update and Retrieval Scripts for Cloud Buckets

sudhin05:
### **Description**

Introduces two key scripts, **`bucket_update.py`** and **`bucket_retrieval.py`** that handle retrieving, updating, and caching for projects in the cloud bucket.  The changes in this PR are a direct continuation of the discussions with @DonggeLiu  in #134  suggesting the need for re-structuring and storing project information in the bucket

#### **`bucket_update.py`**
This script is responsible for updating project information in the bucket by merging existing data from the cloud bucket along with data from the fuzz-target under test. Current Workflow:
1.  Retrieves all fuzz target files and their content from the cloud bucket
2. Matches fuzz targets under test with those in the bucket. For matched targets, it merges the `signatures` into the bucket data
3. Includes fuzz targets from the bucket that are not in the benchmark YAML, leaving their `signatures` empty for now. These can later be populated using FuzzIntrospector or other tools
4. Currently saves the final metadata to a local file (`data_prep/uploads/<project_name>.yaml`) for testing purposes. Can be extended to the bucket with appropriate permissions

#### **2. `bucket_retrieval.py`**
This script complements bucket_update.py by providing a mechanism to retrieve project metadata from either the cloud bucket or a local cache. Current Workflow:

1.  Checks for a locally cached YAML file before querying the bucket. This reduces redundant bucket queries 
2. If the Projectis not available in the local cache, it fetches the data from the cloud bucket and caches it locally for future use. For testing purposes, it reads either from cached file or local file (`data_prep/uploads/<project_name>.yaml`), extensible to cloud bucket with permissions

### **Work Under Development**
  - Populating `signatures`  for unmatched targets  and optional metadata using FuzzIntrospector or other tools.
  - Replacing the simulated upload logic in bucket_update.py with actual bucket uploads with due permission
  - Implementing querying functions to retrieve different parts of the metadata separately
  - Improving the retrieval and merging logic for larger scales

### DIFF
--- a/data_prep/bucket_update.py
+++ b/data_prep/bucket_update.py
@@ -1,0 +1,189 @@
+import sys
+import argparse
+import os
+import yaml
+from google.cloud import storage
+import logging
+from datetime import datetime, timezone
+sys.path.append(os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
+from experiment.benchmark import Benchmark  
+from introspector import get_project_funcs, query_introspector_function_signature
+from datetime import datetime
+
+
+logger = logging.getLogger(__name__)
+logging.basicConfig(level=logging.INFO)
+
+OSS_FUZZ_EXP_BUCKET = 'oss-fuzz-llm-public'
+YAML_DIR = 'benchmark-sets/all'
+
+def extract_from_bucket(bucket, project_name: str) -> list:
+    """Retrieve all fuzz target files and their content from the bucket."""
+    fuzz_targets = []
+    project_prefix = f"human_written_targets/{project_name}/"
+    blobs = bucket.list_blobs(prefix=project_prefix)
+
+    for blob in blobs:
+        # Change current language logic
+        logger.info(f"Retrieving content for fuzz target: {blob.name}")
+        content = blob.download_as_text()
+
+        fuzz_targets.append({
+            "full_path": f"/src/{blob.name.split('/')[-1]}",
+            "content": content,
+            "signatures": [],  
+            "metadata": {
+                "language": "c++" if blob.name.endswith(".cpp") else "c",
+                "dependencies": [],
+                "notes": f"Extracted from {blob.name}"
+            },
+        })
+    return fuzz_targets
+
+def extract_from_benchmark(project_name: str) -> list:
+    """Extract fuzz targets and Project Info from benchmark YAML files."""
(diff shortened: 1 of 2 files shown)


DonggeLiu:
Thanks @sudhin05!
The overall design looks good to me.

I've converted this into a draft because it is under development, please feel free to convert them back once they are ready : )

Also please remember to mention this PR in your GSoC proposal, and ensure you have sufficient time to write/refine the proposal before the deadline. The proposal is the most important evaluation criterion for applications.

sudhin05:
Hi @DonggeLiu , I've incorporated the changes in the update script and validated it's functionality locally. I think we can start together with it's testing to see if it works as intended and may add on / update things along the way

sudhin05:
Hi @DonggeLiu I have added the necessary changes to the update script and validated the same locally. I think we can proceed with testing the same, and see if any more additions/modifications to it are required

data_prep/bucket_update.py
```diff
@@ -0,0 +1,189 @@
...
+            for function in functions:
+                function_name = function.get("function-name")
+                if function_name:
+                    signature = query_introspector_function_signature(project_name, function_name)
+                    if signature:
+                        target_entry["signatures"].append(signature)
+                    else:
+                        logger.info(f"Error: No signature found for function '{function_name}' in project '{project_name}'.")
+            fuzz_targets.append(target_entry)
+        return fuzz_targets
```
DonggeLiu, line 90: IIUC, we might have a misunderstanding here.
The functions in the YAML file should be the functions that are both:
1. Used in the fuzz target.
2. Defined in the project under test.

I reckon your code satisfies 2, but not 1.

We need both because the function and fuzz target will be used as examples in the prompt, like:
```
Example Query: Generate a fuzz target for <function> in <project>
Example Response: <fuzz-target>
```
As a result:
1. If a function is not used by the fuzz target, then it cannot be used in the example query.
2. If a function is not defined in the project-under-test, then it is not testing the <project>.

[`project_targets.py`](https://github.com/google/oss-fuzz-gen/blob/d12ae0e2a9a51dcbc1e2655288b059954a25dc48/data_prep/project_targets.py#L157) has the code that finds the project functions used by fuzz targets. Here we need to save that info in the YAML file, like you proposed at https://github.com/google/oss-fuzz-gen/issues/134#issuecomment-2745195802

DonggeLiu (changes requested):

Issue #931, opened 2025-03-25
Missing activeDeadlineSeconds in Kubernetes Job in pr-exp.yaml


Ekam219:
#### Description:
@DonggeLiu  
In the pr-exp.yaml file, the first Kubernetes job includes the `activeDeadlineSeconds: 604800` (7 days) field to prevent indefinite execution. However, the second job is missing this field, which could result in:
- The job running indefinitely.
- Excessive resource consumption.
- Unpredictable behavior.

#### Suggested Fix:
Add `activeDeadlineSeconds: 604800` to the second job in pr-exp.yaml to match the first job's configuration.

Kartikayy007:
Hello @Ekam219 I think there are some flaws adding it like if there is a job which already have more than 7 days (604800) of job it'll force fully terminate the job on 7th day

Ekam219:
Hello @Kartikayy007,

The 7-day activeDeadlineSeconds in large-pr-exp.yaml was likely chosen as a safe upper bound for large experiments. For smaller jobs in pr-exp.yaml, we could use a shorter time limit based on their expected duration. Aligning with this approach or creating a tiered system would ensure efficient resource management without prematurely terminating jobs with longer durations.

DonggeLiu:
Sorry I missed this earlier, it's a good point.

As far as I know, all jobs should complete within 7 days at most.

In particular, PR experiments are meant for quick testing and should ideally finish within a much shorter window, e.g., 12 hours maximum.

> we can also make it configurable through environment variable
The default timeout is set to 7 days (604800 seconds), but it can be overridden by setting the GKE_EXP_TIMEOUT environment variable.
activeDeadlineSeconds: ${GKE_EXP_TIMEOUT:-604800}

This is certainly fine. We could also hardcode a shorter timeout for PRs for now and make it configurable later if needed.

Ekam219:
CC @DonggeLiu

PR #943, opened 2025-03-31
Added activeDeadlineSeconds for Kubernetes Job Resource Management  Description:

Ekam219:
Description:
This pull request addresses an issue in the pr-exp.yaml file, where the second Kubernetes job was missing the activeDeadlineSeconds field. The absence of this field could have caused the job to run indefinitely, consuming excessive resources and leading to unpredictable behavior. This fix ensures that both jobs have the same resource management strategy.

Changes:
Added activeDeadlineSeconds: 43200 (12 hours) to the second Kubernetes job configuration. This prevents the job from running indefinitely and reduces resource consumption.

Fix:
By adding this field, we align the second job’s configuration with the first job, which already has a similar field, ensuring consistent resource management across both jobs.

Additional Notes:
The change aims to prevent excessive resource consumption by limiting the runtime of the job to 12 hours. This is a more reasonable timeframe compared to the original 7 days, ensuring better resource allocation for quicker jobs.
#931

### DIFF
--- a/ci/k8s/pr-exp.yaml
+++ b/ci/k8s/pr-exp.yaml
@@ -20,6 +20,7 @@ metadata:
     customLabel: "myCustomValue"
 spec:
   template:
+     activeDeadlineSeconds: 43200  # 12 hours. We don't want large experiments to run indefinitely and eat up resources.
     spec:
       nodeSelector:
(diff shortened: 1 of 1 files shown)


DonggeLiu:
Thanks @Ekam219.
I will merge this to an experiment PR so that we can double-check it before merging into main.

Issue #946, opened 2025-03-31
Address TODO: Implement Success Property for RunResult Class


abohoss:
Hello @maintainers,

I've identified a TODO in the `RunResult` class within results.py that reads: 

`# TODO(dongge): Define success property to show if the fuzz target was run.`
I propose to implement `success` property based on 3 attributes:

1. **No Crashes**: Ensuring crashes is False indicates that the fuzz target did not crash during execution.
2. **No Runtime Errors:** Checking that` run_error` is an empty string confirms there were no runtime errors.
3. **Coverage Threshold**: Including a coverage check with a threshold of 0.1% ensures that the fuzzing run was not only executed but also provided minimal level of code exploration, which aligns with the goals of fuzzing.
I will link a pull request with the implementation. As I would appreciate your feedback on this approach, Are there any additional criteria or considerations that should be included in the success property?

DonggeLiu:
Hi @abohoss 
The main idea looks good to me: We do need a way to classify if a fuzzing run succeeds or not.
But I am not sure if this is best time to do so, given we have multiple PRs modifying related files, including crash/coverage analyzer and the result.py refactoring.

I will close this for now, but we can reopen it once these PRs are merged.

abohoss:
Hey @DonggeLiu 
Understood. Also, if there are any other related issues, I would be happy to work on them.

PR #947, opened 2025-03-31
Add success property to RunResult class to address TODO

abohoss:
### Description

defined success property in `RunResult` class in `results.py` to address TODO , which was initially mentioned in #946


DonggeLiu:
Thanks @abohoss!
I believe this has already been addressed in #948 : )

abohoss:
Ah I see , Thanks @DonggeLiu

PR #951, opened 2025-04-01
update priming instruction in priming.txt

siddhesh0705:
**Summary of Changes:**

This pull request includes optimizations made to the priming.txt file, which is used to guide the LLM in generating fuzz targets for the oss-open-fuzz project. The key improvements aim to enhance the efficiency, clarity, and correctness of the fuzz targets generated in the initial run.

**Changes Made:**

- Refined Goal Wording: Updated the fuzz target generation goal to emphasize triggering diverse behaviors rather than focusing on 100% line coverage.

- Explicit Input Source: Added clear instructions for deriving all input variables from the fuzzer input (data and size parameters) or FuzzedDataProvider, instead of using hardcoded or random values.

- Simplicity Emphasis: Added instructions to simplify the fuzz target logic, focusing on converting input data into valid parameters for the function-under-test.

- Error Handling Guidance: Included best practices for handling errors without introducing excessive logic after the function-under-test is called.

- Resource Management: Added instructions to ensure that any allocated memory within the fuzz target is properly freed to avoid memory leaks.

- Goto Warning: Placed an early warning about avoiding goto statements to prevent common pitfalls in C/C++ fuzz target generation.

**Impact:**

These changes are expected to improve build success rates, reduce fix attempts, and increase pipeline efficiency by simplifying the fuzz target logic.

 The optimizations will enhance the semantic correctness of the fuzz targets and improve code coverage during fuzzing.

 Overall, the goal is to create more reliable, efficient, and scalable fuzz targets for testing.


DonggeLiu:
Hi @siddhesh0705.
Unfortunately, rephrasing wording in prompts is not our current priority.

If you like to suggest some contextual information or agent tools that we do not support yet, we would love to evaluate and merge them.
Thanks!

PR #976, opened 2025-04-09
agent: one-prompt*: use gettext for logs

DavidKorczynski:
The logs when shown in the report a not formatted well for openai models, as they are dictionaries not plaintext. We want only the plaintext in the logs.


DavidKorczynski:
Experiment looks good locally, and also https://llm-exp.oss-fuzz.com/Result-reports/ofg-pr/2025-04-10-976-dgk-abz1-1-comparison/index.html

PR #979, opened 2025-04-10
build_generation: Add LLM approach

arthurscchan:
This PR initialis the LLM approach for automatic build script generation of C/C++ projects. This is done by adding a separate path for the new LLM agent. Adding an extra flag --agent could trigger the new LLM agent flow for automatic build script generation.


DavidKorczynski (approved):
This looks great for a start. I think we should make sure to give the agent more control of the environment in the near future.

Issue #969, opened 2025-04-07
Chat history not being properly recorded


myanvoos:
I ran a few experiments in agentic mode today (with the `--ag` flag) and found that the samples' `log.txt` are not showing the agents' interactions that were logged to the console.  At the end of the experiments the `log.txt` for every single sample just looks like:

```
Prototyper

SemanticAnalyzer
{'has_err': False, 'err_type': 'NO_SEMANTIC_ERR', 'crash_symptom': '', 'crash_stacks': [], 'crash_func': {}}
```

Experiments without agentic mode similarly seem to have a blank chat history. They only show the initial prompt in `log.txt`.

After some digging into the codebase and adding
```python
def _container_tool_reaction(self, cur_round: int, response: str,
                               build_result: BuildResult) -> Optional[Prompt]:
    """Validates LLM conclusion or executes its command."""
    prompt = prompt_builder.DefaultTemplateBuilder(self.llm, None).build([])

    if response:
      # Adding this
      build_result.chat_history[self.name] += f"\nRound {cur_round}:\n{response}"
```

to `prototyper.py`, this now works as expected and all the `log.txt` are showing the Prototyper's responses in the newest batch of agentic experiments. However I'm not sure if this is the most idiomatic way to go about it. 

It would be great if you can check the `log.txt` for cloud experiments @DonggeLiu, just to see if you're seeing this same behaviour as well.

myanvoos:
> Thanks for reporting this, [@myanvoos](https://github.com/myanvoos) !
> 
> This is a good catch, most of our experiments are run on cloud so we do need more help maintain local exps. Chat history works fine in cloud exps, it is set here:
> 
> [oss-fuzz-gen/common/cloud_builder.py](https://github.com/google/oss-fuzz-gen/blob/358aff99ac5a4560acc537b726b0c581eb1df1bd/common/cloud_builder.py#L417)
> 
> Line 417 in [358aff9](/google/oss-fuzz-gen/commit/358aff99ac5a4560acc537b726b0c581eb1df1bd)
>  result.chat_history = {agent.name: cloud_build_log} 
> 
> > Experiments without agentic mode similarly seem to have a blank chat history. They only show the initial prompt in log.txt.
> 
> Prototyper is an agent, we indeed missed its chat history, your fix should work. SemanticAnalyzer is not really an agent (LLM does not interact with tools), hence its log is set manually. We will likely replace it with a crash analyzer.

[...]

PR #981, opened 2025-04-11
Add chat history logging to Prototyper

myanvoos:
Related issue #969 

This PR adds chat history logging to the Prototyper agent (and Enhancer as well, since Enhancers inherit from Prototypers).

### DIFF
--- a/agent/prototyper.py
+++ b/agent/prototyper.py
@@ -438,10 +438,16 @@ class Prototyper(BaseAgent):
     try:
       client = self.llm.get_chat_client(model=self.llm.get_model())
       while prompt and cur_round < self.max_round:
+
+        build_result.chat_history[self.name] += f'Step #{cur_round} - "agent-step": <CHAT PROMPT:ROUND {cur_round:02d}>{prompt.get()}</CHAT PROMPT:ROUND {cur_round:02d}>\n'
+        
         response = self.chat_llm(cur_round,
                                  client=client,
                                  prompt=prompt,
                                  trial=last_result.trial)
+        
+        build_result.chat_history[self.name] += f'Step #{cur_round} - "agent-step": <CHAT RESPONSE:ROUND {cur_round:02d}>{response}</CHAT RESPONSE:ROUND {cur_round:02d}>\n'
+        
         prompt = self._container_tool_reaction(cur_round, response,
                                                build_result)
         cur_round += 1
--- a/agent/prototyper.py
+++ b/agent/prototyper.py
@@ -439,15 +439,19 @@ class Prototyper(BaseAgent):
       client = self.llm.get_chat_client(model=self.llm.get_model())
       while prompt and cur_round < self.max_round:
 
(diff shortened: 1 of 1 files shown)


agent/prototyper.py
```diff
@@ -438,10 +438,20 @@ def execute(self, result_history: list[Result]) -> BuildResult:
     try:
       client = self.llm.get_chat_client(model=self.llm.get_model())
       while prompt and cur_round < self.max_round:
+
+        build_result.chat_history[
+            self.
+            name] += f'Step #{cur_round} - "agent-step": <CHAT PROMPT:ROUND {cur_round:02d}>{prompt.get()}</CHAT PROMPT:ROUND {cur_round:02d}>\n'
```
DonggeLiu, line 444: 3 minor notes:
1. We don't have to follow this format (e.g., `Step#`). It used to be this format because that's hardcoded by Cloud Build, now we are adding this by ourselves so we can be more flexible : )
2.  Originally `Step #` is not followed by `{cur_round}`, but by the step number is in CloudBuild, e.g.,
https://github.com/google/oss-fuzz-gen/blob/bd7a86609a561173c4bcc255fe0da1c9fd55d08c/common/cloud_builder.py#L241
We don't have to worry about this since we construct this part by ourselves.
3. With that said, I do think we can consider encoding `cur_round` here for our own needs. E.g., present the round number on the report, or other agents in the workflow.
4. If we add these new chat_history lines here (which is preferred), we will need to remove this line to avoid duplicated chat history in cloud experiments:
5. https://github.com/google/oss-fuzz-gen/blob/00bc7e25947e256c4aefb30a8077444eaba88875/common/cloud_builder.py#L422
6. I would suggest constructing the line in a function hosted by `base_agent`, because:
  * This will likely be widely used at multiple places by multiple agent.
  * It's critical to keep the line's format consistent, because we will need to parse it to generate report.
  * We can extend this to log more than llm-tool interactions. E.g., Text to present on reports or share with other agents in later workflow.

DonggeLiu (commented):
Thanks @myanvoos for addressing this issue!
I am unsure if this is ready to review yet, but I took the liberty of leaving some comments to share some thoughts.
Hope you won't mind : )

myanvoos:
Hey @DonggeLiu, thanks for the comments! And no worries, I did mean this to be a draft PR for more discussion ^^

What I've done is I made this function in `base_agent.py`
```python
def _format_chat_history_line(self, cur_round: int, prompt_or_response: str, 
                                agent_type: str, content: str) -> str:
    """Formats a chat history line."""
    return (f'Round {cur_round:02d}: <{agent_type.upper()} {prompt_or_response.upper()}:ROUND {cur_round:02d}>\n'
            f'{content}\n'
            f'</{agent_type.upper()} {prompt_or_response.upper()}:ROUND {cur_round:02d}>\n')
```

So that in `prototyper.py` we can call
```python
build_result.chat_history[self.name] += self._format_chat_history_line(
            cur_round, 'PROMPT', self.name, prompt.get())

build_result.chat_history[self.name] += self._format_chat_history_line(
            cur_round, 'RESPONSE', self.name, response)
```

And deleted the redundant line in `cloud_builder.py`. 

I included `self.name` in the line to make it easier to know which agent is at any prompt or response when reading the logs i.e. `<PROTOTYPER PROMPT:ROUND 02>`

Is this kind of what you were thinking of?

PR #992, opened 2025-04-15
run_all_experiments: harden coverage gain logic

DavidKorczynski:
In from-scratch runs keyerrors would be observed when accessing the language field. Harden this code to avoid exceptions thrown.


DavidKorczynski:
Experiment looks good: https://llm-exp.oss-fuzz.com/Result-reports/ofg-pr/2025-04-16-992-dgk-comparison/index.html

PR #991, opened 2025-04-15
Dockerfile enhancement: multistage build

AkshataABhat:
This issue aims to refactor the current Dockerfile to use a multistage build approach. The motivation for this change is to:

1. Reduce final image size by excluding unnecessary build-time dependencies (e.g., gcc, libpython3-dev, etc.)
2. Enhance maintainability by clearly separating build and runtime concerns
```
Size of the current docker image: 3.82GB
Size of the docker image after multi-stage build: 1.06GB

```

Closes #990

PR #995, opened 2025-04-16
@AkshataABhat: Dockerfile enhancement: multistage build

DonggeLiu:
Contributed by @AkshataABhat via  #991.


DonggeLiu:
@AkshataABhat, unfortunately the experiment failed:
image

AkshataABhat:
@DonggeLiu I think I've figured out the issue. The python binary doesn't seem to exist in the second stage. I will correct this.

DavidKorczynski (changes requested):
Overall I'm not a huge fan of this -- do we really need it? I feel the complexity of the set up increases using multistage builds and I don't see a significant win from splitting it up.

AkshataABhat:
> 
At the moment, there's no significant improvement except a decrease in image size, but this can help isolate build dependencies from the runtime environment in the future, when support for fine-tuned or custom models is added.

Kartikayy007:
> Overall I'm not a huge fan of this -- do we really need it? I feel the complexity of the set up increases using multistage builds and I don't see a significant win from splitting it up.

I agree to for the complexity point

PR #928, opened 2025-03-25
Coverage analysis agent

DonggeLiu:
An LLM agent that analyzes fuzz targets, fuzzing logs, project source code, coverage reports, and other related data to provide insights and suggestions for improving fuzzing coverage:
* [x] Basic implementation
* [x] fuzz targets
* [x] fuzzing log
* [x] project source code (via bash tool)
* [ ] support coverage report

PR #998, opened 2025-04-16
one-prompt: fix bug for missing coverage agent

DavidKorczynski:
Following https://github.com/google/oss-fuzz-gen/pull/928 the analysis stage tries to get a a coverage agent https://github.com/google/oss-fuzz-gen/blob/d12ae0e2a9a51dcbc1e2655288b059954a25dc48/stage/analysis_stage.py#L38 and throws a RuntimeError if it does not exist: https://github.com/google/oss-fuzz-gen/blob/d12ae0e2a9a51dcbc1e2655288b059954a25dc48/stage/base_stage.py#L55
But only the agent approach has a CoverageAnalyzer: https://github.com/google/oss-fuzz-gen/blob/d12ae0e2a9a51dcbc1e2655288b059954a25dc48/run_one_experiment.py#L256

This fixes it by enabling so an error is not thrown if there's no coverage analyzer agent in the analysis stage. Ultimately I assume one of the aims of the architecture around the pipeline/modularised agent is to support plug-and-play style (many agents, pick those you like), so we should allow different pipelines to use different agents. Arguably we should enable CoverageAnalyzer for the one-prompt pipeline as well, but will do that a bit further down the line after finishing running some experiments focused on other things.

### DIFF
--- a/llm_toolkit/prompt_builder.py
+++ b/llm_toolkit/prompt_builder.py
@@ -314,15 +314,17 @@ class DefaultTemplateBuilder(PromptBuilder):
                          instruction: str = '') -> prompts.Prompt:
     """Prepares the code-fixing prompt."""
     priming, priming_weight = self._format_fixer_priming(benchmark)
+
     if error_desc and errors:
-      problem = self._format_fixer_problem(raw_code, error_desc, errors,
-                                           priming_weight, context, instruction)
+      pass
+    elif coverage_result:
+      error_desc = coverage_result.insight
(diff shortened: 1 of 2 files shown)


DavidKorczynski:
Going through the logs, can confirm we're not running into the RuntimeError exceptions anymore. The overall experiment looks good: https://llm-exp.oss-fuzz.com/Result-reports/ofg-pr/2025-04-16-998-dgk-comparison/index.html but there are some quota issuse and some other runtime issues, which are essentially unrelated to this, so will fix in another PR.

PR #1000, opened 2025-04-16
results: add success field to base class

DavidKorczynski:
We're running into issues in the cloud build when dill deserialization fails, because https://github.com/google/oss-fuzz-gen/blob/d12ae0e2a9a51dcbc1e2655288b059954a25dc48/common/cloud_builder.py#L422 may return a Result object, which gets called from https://github.com/google/oss-fuzz-gen/blob/d12ae0e2a9a51dcbc1e2655288b059954a25dc48/stage/base_stage.py#L62-L63 which get's called from https://github.com/google/oss-fuzz-gen/blob/d12ae0e2a9a51dcbc1e2655288b059954a25dc48/stage/analysis_stage.py#L39 and the pipeline uses the success field: https://github.com/google/oss-fuzz-gen/blob/d12ae0e2a9a51dcbc1e2655288b059954a25dc48/pipeline.py#L123-L127 but the base Result class does not have a success field: https://github.com/google/oss-fuzz-gen/blob/d12ae0e2a9a51dcbc1e2655288b059954a25dc48/results.py#L24-L70

This fixes it by introducing a default (failed) success property.

### DIFF
--- a/results.py
+++ b/results.py
@@ -39,7 +39,8 @@ class Result:
                fuzz_target_source: str = '',
                build_script_source: str = '',
                author: Any = None,
-               chat_history: Optional[dict] = None) -> None:
+               chat_history: Optional[dict] = None,
+               default_success: bool = False) -> None:
     self.benchmark = benchmark
     self.trial = trial
     self.work_dirs = work_dirs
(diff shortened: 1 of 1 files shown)


AdamKorcz (approved):

DonggeLiu:
Two observations (probably unrelated to this PR):
1. The build rate is very low 10/34.
2. Many [quota issues](https://llm-exp.oss-fuzz.com/Result-reports/ofg-pr/2025-04-16-1000-dgk-comparison/sample/output-util-linux-mnt_new_table_from_dir/01.html), which likely caused 1.

PR #1002, opened 2025-04-16
build_generation: llm build script for general project

arthurscchan:
This PR adds a second llm agent on top of #979. The new llm agent considers the build generation for pure source code without build configuration files for different build system.


arthurscchan:
no longer needed.

PR #1008, opened 2025-04-22
Tests for report generation

harshtech123:
hii @DonggeLiu  i notice that there is a workflow available to generate reports and update it on cloud , but there is no end to end testing available for that process , we must have test cases that assert that current functionality works well with our expectations .

I tried to write test cases for our report after this i am aiming for writing full test suits for our whole project ! thank you so much.


DonggeLiu:
Hi @harshtech123, thanks for the PR.
Unfortunately this is a low priority for us at the moment, and it may take me a long time before I can find time to review this PR.

harshtech123:
this seems fine to me , i will be active on this anytime you wants / review so if there any changes we will adjust test cases ! thank you so much.

PR #1039, opened 2025-05-02
build_generation: Fix prompt for strict constraint and more guideline

arthurscchan:
This PR strengthens the build script generation prompt with the following improvements:

1) Enforces use of `$SRC` in include paths to ensure correct header resolution.
2) Requires use of `apt-get` to install dependencies before falling back to manual builds of missing dependencies.
3) Adds explicit handling for Android.bp (Soong) projects and AOSP build assumptions.
4) Prohibits manual compilation when a valid build system (e.g., Makefile) exists.
5) Adds support for minimal compatibility shims (e.g., for BSD macros) when needed.
6) Improves discovery strategy by including README/INSTALL parsing suggestion.
7) Tightens rules around excluding tests and use of `make -C` and flags.

### DIFF
--- a/experimental/build_generator/templates.py
+++ b/experimental/build_generator/templates.py
@@ -398,7 +398,7 @@ You must **not guess types or fabricate includes**, but minimal compatibility st
 
 - You must explicitly include **all header directories required by the project** using `-I` flags. This applies to `CFLAGS`, `CXXFLAGS`, and `CPPFLAGS`, depending on which is being used.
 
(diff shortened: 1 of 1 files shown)


DavidKorczynski (approved):

Kartikayy007:
Can we consider adding automated `#include` analysis to detect and install dependencies rather than relying on manual specification? 
what do you say @arthurscchan @DavidKorczynski

PR #1055, opened 2025-05-13
Fix error in true/false positive crash designation on the experiment result page

AmPaschal:
This PR fixes an error in the oss-fuzz-gen's experiment result page where true positive crashes are classified as false positives, and vice versa.

In report/templates/benchmark.html, crashes are designated as True vulnerabilities if `sample.result.crashes and **not** sample.result.is_semantic_error`.

However, in report/templates/index.html, the reverse occurs. Crashes were previously designated as True positives if `bug_sample.sample.result.is_semantic_error`. This PR fixes this discrepancy.

### DIFF
--- a/report/templates/index.html
+++ b/report/templates/index.html
@@ -172,7 +172,7 @@ td .signature {
             <td class="table-index">{{ loop.index }}</td>
             <td data-sort-value="{{bug_sample.benchmark.project}}">{{bug_sample.benchmark.project}}</td>
             <td data-sort-value="{{bug_sample.benchmark.id}}"><a href="sample/{{ bug_sample.benchmark.id|urlencode }}/{{ bug_sample.sample.id }}.html">{{bug_sample.benchmark.id}}-{{ bug_sample.sample.id }} </a> </td>
-            <td style="color: {{ 'red' if bug_sample.sample.result.crashes and not bug_sample.sample.result.is_semantic_error else 'green' }}" data-sort-value="{{'True positive' if bug_sample.sample.result.is_semantic_error else 'False positive'}}">{{'True positive' if bug_sample.sample.result.is_semantic_error else 'False positive'}}</td>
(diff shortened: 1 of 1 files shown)


report/templates/index.html
```diff
@@ -172,7 +172,7 @@ <h2>Crashes found by generated fuzz harnesses</h2>
             <td class="table-index">{{ loop.index }}</td>
             <td data-sort-value="{{bug_sample.benchmark.project}}">{{bug_sample.benchmark.project}}</td>
             <td data-sort-value="{{bug_sample.benchmark.id}}"><a href="sample/{{ bug_sample.benchmark.id|urlencode }}/{{ bug_sample.sample.id }}.html">{{bug_sample.benchmark.id}}-{{ bug_sample.sample.id }} </a> </td>
-            <td style="color: {{ 'red' if bug_sample.sample.result.crashes and not bug_sample.sample.result.is_semantic_error else 'green' }}" data-sort-value="{{'True positive' if bug_sample.sample.result.is_semantic_error else 'False positive'}}">{{'True positive' if bug_sample.sample.result.is_semantic_error else 'False positive'}}</td>
+            <td style="color: {{ 'red' if bug_sample.sample.result.crashes and bug_sample.sample.result.is_semantic_error else 'green' }}" data-sort-value="{{'True positive' if not bug_sample.sample.result.is_semantic_error else 'False positive'}}">{{'True positive' if not bug_sample.sample.result.is_semantic_error else 'False positive'}}</td>
```
DonggeLiu, line 175: 1. Shall we keep the color consistent with the existing design? I.e., red for true positive and green for false positive.
Essentially, true positive is more alarming because it indicates a potential vul should be reported to the project maintainers.

2. Would be clearer if we do:
```python
'False positive' if bug_sample.sample.result.is_semantic_error else 'True positive'
```

DonggeLiu (approved):
Thanks @AmPaschal for noticing and fixing this.
Some nits before merging.

report/templates/index.html
AmPaschal: Yes, these suggestions look good. I have changed the color and text accordingly.

DonggeLiu (approved):
Thanks!

oliverchang:
> Thanks @AmPaschal for noticing and fixing this. Some nits before merging.

+1 Thanks for noticing this @AmPaschal ! Great work catching this.

DavidKorczynski (approved):

PR #1065, opened 2025-05-19
Get and log the source code of the fuzz target and build script.

AmPaschal:
This commit also modifies the create_ossfuzz_project to return the path of the generated oss-fuzz project so it can read the files from the project. Finally, this commit adds the retrieved sources to the report log.

### DIFF
--- a/experiment/evaluator.py
+++ b/experiment/evaluator.py
@@ -280,7 +280,8 @@ class Evaluator:
               f'{self.benchmark.target_path}\n')
 
     if not build_script_path or os.path.getsize(build_script_path) == 0:
-      return name
+      # Return the generated project path to the caller.
+      return generated_project_path
 
     # Copy generated build script to generated_project_path
     shutil.copyfile(
@@ -294,7 +295,8 @@ class Evaluator:
     with open(os.path.join(generated_project_path, 'Dockerfile'), 'a') as f:
       f.write('\nCOPY agent-build.sh /src/build.sh\n')
 
-    return name
+    # Return the generated project path to the caller.
+    return generated_project_path
 
   def _fix_generated_fuzz_target(self, ai_binary: str,
                                  generated_oss_fuzz_project: str,
--- a/stage/execution_stage.py
+++ b/stage/execution_stage.py
@@ -60,12 +60,33 @@ class ExecutionStage(BaseStage):
                                     f'{last_result.trial:02d}.fuzz_target')
     build_script_path = os.path.join(last_result.work_dirs.fuzz_targets,
                                      f'{last_result.trial:02d}.build_script')
-    evaluator.create_ossfuzz_project(generated_oss_fuzz_project,
+    generated_project_path = evaluator.create_ossfuzz_project(generated_oss_fuzz_project,
                                      fuzz_target_path, build_script_path)
 
     status_path = os.path.join(last_result.work_dirs.status,
                                f'{last_result.trial:02}')
     os.makedirs(status_path, exist_ok=True)
+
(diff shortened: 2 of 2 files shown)


stage/execution_stage.py
```diff
@@ -60,13 +60,34 @@ def execute(self, result_history: list[Result]) -> Result:
...
-    evaluator.create_ossfuzz_project(generated_oss_fuzz_project,
+    generated_project_path = evaluator.create_ossfuzz_project(generated_oss_fuzz_project,
                                      fuzz_target_path, build_script_path)
 
     status_path = os.path.join(last_result.work_dirs.status,
                                f'{last_result.trial:02}')
     os.makedirs(status_path, exist_ok=True)
 
+    # Get the source code of the fuzz target and build script from the generated oss-fuzz projects.
+    fuzz_target_source_path = os.path.join(generated_project_path, f'{last_result.trial:02d}.fuzz_target')
```
AmPaschal, line 71: I could have just reused the existing fuzz_target_path. However, I wanted to actually log the version of the fuzz_target in the generated_oss_fuzz project (in case it differed with the version produced in the last_result.work_dirs.fuzz_targets path and stored in the current fuzz_target_path variables).

experiment/evaluator.py
```diff
@@ -280,7 +280,8 @@ def create_ossfuzz_project(self,
               f'{self.benchmark.target_path}\n')
 
     if not build_script_path or os.path.getsize(build_script_path) == 0:
-      return name
```
AmPaschal, line 283: I checked callers of this function to see if this returned `name` variable is being used. I didn't find any usage, hence believed it was safe to change it.

oliverchang (commented):
hey @AmPaschal , thanks for this change! Can you please add some context / reasoning behind this change in the PR description?

stage/execution_stage.py
```diff
@@ -60,13 +60,37 @@ def execute(self, result_history: list[Result]) -> Result:
...
+
+    if build_script_source_path and os.path.isfile(build_script_source_path):
+      with open(build_script_source_path, 'r') as f:
+        build_script_source = f.read()
+    else:
+      build_script_source = ''
+
+    # log the fuzz target and build script's source code.
+    self.logger.info('\nFuzz target source: \n%s\n', fuzz_target_source)
+    self.logger.info('\nBuild script source: \n%s\n', build_script_source)
```
DonggeLiu, line 92: Could you separate this block into individual functions, if this new block is for logging / debugging only?
For example, a function to get fuzz target source code, and log the fuzz target source code.
Another function to get build script and log the build script.

This allows us to handle different cases. E.g.,
 e.g., if the fuzz target is empty, log an error; Otherwise log a debug message.
If the build script is empty, log a debug message saying the original build script will be used, etc.
You don't have do handle these cases now, just FYI.

Thanks :)

stage/execution_stage.py
```diff
@@ -136,6 +160,11 @@ def execute(self, result_history: list[Result]) -> Result:
       else:
         run_log_content = ''
 
+      # Add fuzz_target_source and build_script_source to the report log.
+      run_log_content = (f'Fuzz target source:\n{fuzz_target_source}\n'
+                         f'Build script source:\n{build_script_source}\n'
+                         f'{run_log_content}')
```
DonggeLiu, line 166: Here we can re-invoke those function created above.

DonggeLiu (approved):

AmPaschal:
/gcbrun exp -n pamusuo -m vertex_ai_gemini-2-5-pro-chat -ag -b quick-test

PR #1073, opened 2025-05-27
Handle empty error description and errors in DefaultTemplateBuilder

### LINKED ISSUE #1052: Empty error_desc and errors lead to empty fixer_problem.txt during compilation fix
In `one_prompt_prototyper.py` at line 152, an empty string is passed to the `error_desc` parameter of `builder.build_fixer_prompt`. Then, inside the function, the following logic is executed:

```python
if error_desc and errors:
    pass
elif coverage_result:
    error_desc = coverage_result.insight
    errors = coverage_result.suggestions.splitlines()
else:
    error_desc = ''
    errors = []
```

This leads to `error_desc = ''` and `errors = []` always being empty. Consequently, at line 326 in `prompt.builder.py`, the following call:

```python
problem = self._format_fixer_problem(raw_code, error_desc, errors, priming_weight, context, instruction)
```

receives empty `error_desc` and `errors` values, which causes the `fixer_problem.txt` file, when attempting to fix compilation errors, to always have the following content:

```
Below is the error to fix:
{ERROR_SUMMARY}
<error>
{ERROR_MESSAGES}
</error>
```

and remain empty.

ch097711:
This PR fixes #1052 by ensuring that detailed build error messages are retained when generating fixer prompts.

Changes:
- Prevent errors from being cleared if error_desc is empty but errors has content.
- Default error_desc to an empty string if None.
- Default errors to an empty list if None.

This avoids the “Unexpected empty error message” warning.

### DIFF
--- a/llm_toolkit/prompt_builder.py
+++ b/llm_toolkit/prompt_builder.py
@@ -321,8 +321,8 @@ class DefaultTemplateBuilder(PromptBuilder):
       error_desc = coverage_result.insight
       errors = coverage_result.suggestions.splitlines()
     else:
-      error_desc = ''
-      errors = []
+      error_desc = error_desc or ''
+      errors = errors or []
     problem = self._format_fixer_problem(raw_code, error_desc, errors,
                                          priming_weight, context, instruction)
 


DonggeLiu:
Thanks @ch097711 !

DonggeLiu (approved):

PR #1106, opened 2025-06-11
add initial function-based prototyper

DavidKorczynski:
Adds new prototyper based on tool calling using openai's API. It does this by way of a new pipeline instead of substituing into the existing prototyper. There's still some work to be done here for it to work well, but it outlines the core needed to have harnesses generated. Submitting it now to make it easier to iterate over collectively.


DavidKorczynski:
/gcbrun exp -n dgk-z1 -m vertex_ai_gemini-1-5 -i -b comparison

PR #1114, opened 2025-06-16
one_prompt_prototyper: Fix error extraction

arthurscchan:
Sometimes the compile_log data in build_result is truncated due to excessive stdout and stderr output from the test build. This can result in missing error messages and cause the error extraction logic to fail, preventing any error messages from being passed to the LLM for code fixing.
This PR resolves the issue by also using the full stderr stored in build_result.compile_error to extract error messages, and introduces deduplication logic to avoid sending repeated messages to the LLM.


arthurscchan:
/gcbrun exp -n ascc-pr-1114-1 -m vertex_ai_gemini-1-5 -i -b comparison

PR #1120, opened 2025-06-17
stage: execution: handle binary data

DavidKorczynski:
The run logs can contain binary that can't be decoded. This is likely only a tiny bit of the logs, though, so we shouldn't fail hard on unicode decoding errors as such. This fixes it by doing a softer decoding and making sure we don't break because of unicode decode errors.


DavidKorczynski:
/gcbrun exp -n dgk-pr-1120-1 -m vertex_ai_gemini-2-5-pro-chat -b comparison

AdamKorcz (approved):

PR #1123, opened 2025-06-17
prompt: Fix error message handling

arthurscchan:
This PR fixes an incorrect check where, if either error_desc or errors is empty, the entire error message is skipped. It also improves the handling of error messages to ensure that at least one is included for the LLM to process, even when the token limit is reached.


arthurscchan:
/gcbrun exp -n ascc-pr-1123-1 -m vertex_ai_gemini-2-5-pro-chat -b comparison

PR #1128, opened 2025-06-18
Fix use of excessive storage during experiments

AmPaschal:
This PR fixes experiment failures caused by using excessive storage.
The problem was caused by temporary files being created but not getting deleted.


DonggeLiu (approved):

AmPaschal:
/gcbrun exp -n pa -m vertex_ai_gemini-2-5-pro-chat -ag

PR #1102, opened 2025-06-08
Revamp overall report UI and improve report navigation

myanvoos:
This PR overhauls the overall report UI (layout and template/sample pages) and adds a few improvements including:

- A dark mode theme for all three pages
- Navigation breadcrumbs
- Toggle-able table of content for the index page (located top right of the screen)
- Dedicated header showing aggregated results for the index and benchmark pages
- A function to create a unified JSON file out of experimental results that preserves the hierarchical relationships between projects, benchmarks, and samples

PR #1134, opened 2025-06-24
 Add text search across trials/samples

myanvoos:
This PR adds a text search functionality to the report, allowing the user to search for trials/samples containing a certain keyword. It makes use of and expands the unified data JSON structure that was introduced in #1102 by adding some additional fields to each trial/sample.

Searchable fields:
- Bug type (for example: `SEGV`, `heap-buffer-overflow`)
- Crash address
- Sanitizer (for example, `AddressSanitizer`)
- Triage
- Triager prompt

It also adds an intuitive search UI that persists across all three pages so the user can perform the search action regardless of what page they're currently on.

## Preview
image


myanvoos:
It seems like the parsing logic is not correctly extracting the crash details for them to be searchable and is relying too much on the semantic analyzer log. I'll mark this as draft for now to fix this.

myanvoos:
JOB: https://console.cloud.google.com/kubernetes/job/us-central1-c/llm-experiment/default/ofg-pr-1134-mv
REPORT: https://llm-exp.oss-fuzz.com/Result-reports/ofg-pr/2025-06-25-1134-mv-comparison/index.html
BUCKET: https://console.cloud.google.com/storage/browser/oss-fuzz-gcb-experiment-run-logs/Result-reports/ofg-pr/2025-06-25-1134-mv-comparison
BUCKET GS: `gs://oss-fuzz-gcb-experiment-run-logs/Result-reports/ofg-pr/2025-06-25-1134-mv-comparison`

DonggeLiu (approved):

DonggeLiu:
/gcbrun exp -n mv -m vertex_ai_gemini-2-5-pro-chat -ag

myanvoos:
Update: I added RegEx into the search feature, expanded the search scope to include the entire logs, added a pagination system to limit the display to showing 10 samples at a time, and added a basic scoring system to rank samples according to the number of matches across all their searchable fields.

## Preview 
image

myanvoos:
Looks like adding the full run logs onto the JSON crashed the report :(

DonggeLiu:
> Looks like adding the full run logs onto the JSON crashed the report :(

That's sad, let's support the basics only then : )

myanvoos:
> > Looks like adding the full run logs onto the JSON crashed the report :(
> 
> That's sad, let's support the basics only then : )

Maybe we can compromise somewhere? As in we don’t make the full run logs searchable, but we can still make the execution stage info (or similar) searchable. It would just be a few more lines of code to single out execution stage logs I think.

IMO one of the main issues is that the full logs can contain things like `ERROR:root:Exception <class 'google.api_core.exceptions.TooManyRequests'> (429 POST)` lines or long code blocks that might not always be relevant for searching across multiple benchmarks.

For this PR, I'll remove the full run logs but keep RegEx in as that's working fine. We can always extend this search functionality in the future

myanvoos:
/gcbrun exp -n mv -m vertex_ai_gemini-2-5-pro-chat -ag

DavidKorczynski (approved):

PR #1141, opened 2025-07-03
Optimize GDB prompt in CrashAnalyzer

maoyixie:
This pr is used to optimize the gdb prompt. The language is more refined and unified, and the structure is clearer. 

In particular, In GDB, do not use shell-style redirection like:
```run < {AFTIFACT_PATH}```
This won't work because GDB is not a shell. I have seen too many these cases, LLM model like GPT-4o also makes this mistake.

Also, avoid using:
```run -- {AFTIFACT_PATH}```
unless the fuzz driver explicitly expects that pattern. Relevant case: https://llm-exp.oss-fuzz.com/Result-reports/ofg-pr/2025-07-01-1133-pamusuo-comparison/sample/output-igraph-igraph_sparsemat_arpack_rssolve/03.html#:~:text=54%3A58%20%5BTrial%20ID%3A%2003%5D%20INFO%20%5Bbase_agent.-,chat_llm,-%3A96%5D%3A

Furthermore, while ```run {AFTIFACT_PATH}``` seems correct, it may execute the testcase multiple times, depending on how libFuzzer is configured.

✅ Instead, always use:
```run -runs=1 {AFTIFACT_PATH}```
to ensure the fuzz driver runs exactly once with the provided testcase — ideal for reproducible crash debugging.

### DIFF
--- a/prompts/tool/gdb_tool.txt
+++ b/prompts/tool/gdb_tool.txt
@@ -41,7 +41,7 @@ You can issue GDB commands one at a time and I’ll return each command’s outp
 <general rules>
 1. Do **NOT** run `gdb /out/{TARGET_NAME}` again or issue any shell commands.
 2. Do **NOT** compile or run code outside of GDB.
-3. Do **NOT** use `run < /artifact/testcase` or `run -- /artifact/testcase`. Instead, use `run -runs=1 /artifact/testcase` to execute the crash input exactly once.
+3. Do **NOT** use `run < {AFTIFACT_PATH}` or `run -- {AFTIFACT_PATH}`. Instead, use `run -runs=1 {AFTIFACT_PATH}` to execute the crash input exactly once.
 4. No Markdown code fences (```), only the XML tags defined above.
 </general rules>
 </tool>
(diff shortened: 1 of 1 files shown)


maoyixie:
/gcbrun exp -n my -m vertex_ai_gemini-2-5-pro-chat -ag

maoyixie:
There seems to be no successful build example in this experiment. Can we change the test set to conduct the experiment?
@DonggeLiu , exp report link: https://llm-exp.oss-fuzz.com/Result-reports/ofg-pr/2025-07-03-1141-my1-comparison/

maoyixie:
Here we encounter the problem you mentioned before: artifact cannot be found. Many cases. For example, in the following example: 
https://llm-exp.oss-fuzz.com/Result-reports/ofg-pr/2025-07-03-1141-my1-comparison/sample/output-quickjs-js_parsejson2/02.html

But actually the artifact is stored at the bucket: 
https://console.cloud.google.com/storage/browser/oss-fuzz-gcb-experiment-run-logs/Result-reports/ofg-pr/2025-07-03-1141-my1-comparison/results/output-quickjs-js_parsejson2/artifacts/02.fuzz_target-F0-02?pageState=(%22StorageObjectListTable%22:(%22f%22:%22%255B%255D%22))&inv=1&invt=Ab1pjA

maoyixie:
According to this case: https://llm-exp.oss-fuzz.com/Result-reports/ofg-pr/2025-07-03-1141-my1-comparison/sample/output-libfuse-af_gb_alloc_data/02.html

It can found the artifact. And the new prompt is ok. I think this pr is ok can it can be merged.

PR #1151, opened 2025-07-24
Add step parsing for each agent section

myanvoos:
This PR adds step parsing for each agent section to transform the raw content at each step into a more human-readable format, with collapsible subsections containing a text preview of the chat prompt/response (first 50 words).

## Preview
image

### DIFF
--- a/report/parse_logs.py
+++ b/report/parse_logs.py
@@ -44,77 +44,91 @@ class LogsParser:
     for i, line in enumerate(lines):
       line = line.strip()
       if line == '<bash>':
-        for j in range(i + 1, len(lines)):
-          if lines[j].strip() == '</bash>':
-            bash_content = '\n'.join(lines[i + 1:j]).strip()
-            if bash_content:
-              first_line = bash_content.split('\n')[0].strip()
-              if first_line:
-                # skip comments and placeholder text
-                if (first_line.startswith('#') or
-                    first_line.startswith('[The command') or
-                    first_line.startswith('No bash') or
-                    'No bash' in first_line or len(first_line) < 3):
-                  continue
-
-                parts = first_line.split()
-                if parts:
-                  cmd = parts[0]
-
-                  if cmd == 'grep':
-                    # Extract the search term (usually the first quoted argument)
-                    import re
-                    quoted_match = re.search(r"'([^']+)'", first_line)
-                    if quoted_match:
-                      search_term = quoted_match.group(1)
-                      command_summary = f"grep '{search_term}'"
-                    else:
-                      key_args = []
-                      for part in parts[1:]:
-                        if not part.startswith('-') and len(part) > 1:
-                          if len(part) > 20:
-                            part = part[:17] + '...'
-                          key_args.append(part)
-                          if len(key_args) >= 1:  # Limit to 1 arg for grep
-                            break
-                      command_summary = f"{cmd} {' '.join(key_args)}".strip()
-                  else:
-                    key_args = []
-                    for part in parts[1:]:
-                      if not part.startswith('-') and len(part) > 1:
-                        if len(part) > 20:
-                          part = part[:17] + '...'
-                        key_args.append(part)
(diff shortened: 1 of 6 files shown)


DonggeLiu:
Thanks, this looks great! :)

Here are a few thoughts on the UX:
1. The downward-pointing angle for all steps is a bit ambiguous. Just to confirm my understanding: are steps 1-2 collapsed, step 3 expanded (but its content hidden), and step 4 fully expanded?

2. The titles would be much more informative if they showed the tool being used, like 'Step 1 - Bash' or 'Step 5 - Conclusion'. This would make the agent's logic much easier to scan. It'd be great if this could handle multiple tools, too (e.g., 'Step 5 - Bash, Conclusion').

3. Building on that, what do you think about simplifying the UI by having only one layer of collapse? A collapsed step would show just the new informative title, and an expanded one would reveal all its content, like your 'Step 4' example.

These are just my initial impressions: Happy to hear what you think!

myanvoos:
1. Yep, your understanding is right.
2. Good point, I'll add that in. My current plan is to parse the unique tags and put that in the title i.e. if there's at least one `<bash>` line and at least one `<reason>` line, the title will show 'Step X - Bash, Reasoning`. 
3. With the new informative title, I think this is a good idea. 
4. What do you think about putting some kind of content preview in the title so that users can see at a glance the difference between, say, two steps that both have 'Bash, Reasoning'? 

So instead of
```
Step 1 - Bash, Reasoning
Step 2 - Bash, Reasoning
Step 3 - Bash, Conclusion
```

We use some variant of
```
Step 1 - Bash ('janet_fiber'), Reasoning, Bash ('typedef struct JanetFunction'), Reasoning, Bash ('JanetFunction')
Step 2 - Bash ('typedef struct JanetFunction'), Bash ('JanetFunction'), Reasoning, Bash ('janet_wrap_cfunction')
Step 3 - Bash ('janet_wrap_cfunction'), Conclusion
```

to represent the following logs:
```xml
Step 1:
<bash>
grep -rn 'janet_fiber(' /src/janet/
</bash>
...
<reason>
Find the definition of the `JanetFunction` type.
</reason>
<bash>
grep -rn 'typedef struct JanetFunction' /src/janet/
</bash>

<reason>
Examine how to create a JanetFunction.  Looking for examples of its usage.
</reason>
<bash>
grep -rn 'JanetFunction' /src/janet/ | grep -v 'typedef struct JanetFunction'
</bash>

Step 2:
<bash>
grep -rn 'typedef struct JanetFunction' /src/janet/
</bash>
...
<bash>
grep -rn 'JanetFunction' /src/janet/ | grep -v 'typedef struct JanetFunction'
</bash>
...
<reason>
It appears that creating a `JanetFunction` directly is complex.  Instead, I'll try to get a `JanetFunction` by creating a Janet function using `janet_wrap_cfunction`.
</reason>
<bash>
grep -rn 'janet_wrap_cfunction' /src/janet/
</bash>
...

Step 3:
<bash>
grep -rn 'janet_wrap_cfunction' /src/janet/
</bash>
<conclusion>
The fuzz target needs to initialize Janet, create a dummy C function, wrap it as a `Janet` value using `janet_wrap_cfunction`, create a `JanetFiber` with this value, call `janet_loop_fiber` with the fiber, and then deinitialize Janet. The build script doesn't need modification.
</conclusion>
```

myanvoos:
/gcbrun exp -n mv -m vertex_ai_gemini-2-flash-chat -ag

myanvoos:
/gcbrun exp -n mv -m vertex_ai_gemini-2-flash-chat -ag

myanvoos:
/gcbrun exp -n mv -m vertex_ai_gemini-2-flash-chat -ag

myanvoos:
/gcbrun exp -n mv -m vertex_ai_gemini-2-flash-chat -ag -b final-ui-fixes

DonggeLiu (approved):

Issue #1153, opened 2025-07-28
Error Running Crash Analyzer in OSS-Fuzz-Gen Experiments


AmPaschal:
# Summary
I have noticed the following errors when using the crash analyzer during OSS-Fuzz-Gen experiments. These errors are frequent and commonly lead to incorrect root cause analysis of the crash.

1. GDB complains that it cannot insert breakpoint because it cannot access memory at the breakpoint's address.
https://llm-exp.oss-fuzz.com/Result-reports/ofg-pr/2025-07-16-1148-pamusuo-analyzer-tests-1/sample/output-libraw-_zn6libraw17crxloaddecodeloopepvi/08.html
https://llm-exp.oss-fuzz.com/Result-reports/ofg-pr/2025-07-16-1148-pamusuo-analyzer-tests-1/sample/output-libraw-_zn6libraw17crxloaddecodeloopepvi/03.html

2. GDB complains that the artifact directory was not found.
https://llm-exp.oss-fuzz.com/Result-reports/ofg-pr/2025-07-16-1148-pamusuo-analyzer-tests-1/sample/output-astc-encoder-_z20symbolic_to_physicalrk21block_size_descriptorrk25symbolic_compressed_blockph/03.html
https://llm-exp.oss-fuzz.com/Result-reports/ofg-pr/2025-07-16-1148-pamusuo-analyzer-tests-1/sample/output-libsndfile-sf_open/10.html (crash in cycle 1)

3. The LLM hallucinates GDB interactions and uses this hallucinated interaction to derive its conclusion.
https://llm-exp.oss-fuzz.com/Result-reports/ofg-pr/2025-07-16-1148-pamusuo-analyzer-tests-1/sample/output-mosh-_zn8terminal8emulator6resizeemm/10.html
https://llm-exp.oss-fuzz.com/Result-reports/ofg-pr/2025-07-16-1148-pamusuo-analyzer-tests-1/sample/output-quickjs-js_parsejson2/07.html

# Why these issues are important 

Agents in OSS-Fuzz-Gen, specifically the Context Analyzer and the Enhancer, rely on the root cause analysis from the crash analyzer to either determine the feasibility of the crash or modify the fuzz driver to prevent similar crashes. Hence, when the crash analyzer is wrong, the context analyzer is usually also wrong (and reports false crashes as being feasible or vice versa) or the enhancer fails to fix the crash in the next cycle.

# Root Causes

[...]

PR #1154, opened 2025-07-28
Fix Reported Bugs in the Crash Analyzer

AmPaschal:
This PR addresses the three errors reported in [Issue 1153](https://github.com/google/oss-fuzz-gen/issues/1153).

## Error 1: GDB complains that it cannot insert breakpoint because it cannot access memory at the breakpoint's address.
To address this error, this PR modifies the oss-fuzz-checkout.prepare_project_image function to use the oss-fuzz project created by the crash analyzer using the Evaluator.create_ossfuzz_project_with_gdb function.

In addition, this PR also disables caching by default. This is because when a cached project image is used, OSS-Fuzz-Gen edit the Dockerfile in the created oss-fuzz project and removes the commands modifying environment variables.
A cleaner solution would be to also modify the compile flags in the cached image but I couldn't get this to work within the short time I worked on this.

## Error 2: GDB complains that the artifact directory was not found.
Using the oss-fuzz project created using the Evaluator.create_ossfuzz_project_with_gdb function partially addresses this error. 
I also modified the tutorial of the GDB tool so that it references the correct path of the artifact in the project container used by the GDB tool.

## Error 3: The LLM hallucinates GDB interactions and uses this hallucinated interaction to derive its conclusion.
Since I cannot prevent the LLM from hallucinating, I addressed this error by adding two validations to the LLM response. 
First, I added a check that the LLM response does not contain both gdb/bash commands and the LLM's conclusion. The reasoning is that the LLM should not be issuing tool commands and providing a conclusion at the same time. 
In the future, I expect that the use of function tools should also address this problem.

I also added a second validation that ensures the gdb tool is used at least once before the LLM produces a response. I'm not sure if this can prevent hallucination, but from my experience, once the LLM uses the GDB tool the first time, it continues using the GDB tool.

Finally, this PR also fixes a bug in the Crash Analyzer where the ProjectContainer tool is not terminated before the Crash Analyzer exits and adds necessary files for testing the Crash Analyzer directly.

## Evaluation Instruction
The Crash Analyzer (before and after the changes provided by this PR) can be tested directly using the command below (change the oss-fuzz directory path):
```
python3 -m agent_tests.agent_test -y benchmark-sets/comparison/mosh.yaml -f _ZN8Terminal11Framebuffer6resizeEii -p CrashAnalyzer -pf agent_tests/prompt_files/crash-analyzer-mosh-01.txt -afp ./agent_tests/2025-07-16-1148-pamusuo-analyzer-tests-1/ -of [path/to/oss-fuzz] > result-test-01.txt 2>&1
```
## Expectation
Without the contributed changes, the LLM response will exhibit one of the three errors described in [Issue 1153](https://github.com/google/oss-fuzz-gen/issues/1153).
With the contributed changes, the LLM should use the GDB tool at least once, and the GDB tool invocation should not fail.

### DIFF
--- a/agent/crash_analyzer.py
+++ b/agent/crash_analyzer.py
@@ -134,16 +134,19 @@ class CrashAnalyzer(BaseAgent):
     # If there's a conclusion tag and a tool usage tag, then there's an error
     prompt = prompt_builder.CrashAnalyzerTemplateBuilder(self.llm,
                                                          None).build([])
-    if self._parse_tag(response, 'conclusion') and (self._parse_tag(response, 'gdb') or self._parse_tag(response, 'bash')):
+    if self._parse_tag(response, 'conclusion') and (self._parse_tag(
+        response, 'gdb') or self._parse_tag(response, 'bash')):
       extra_note = 'NOTE: You cannot provide both tool commands and conclusion in the same response.'
       return self._container_handle_invalid_tool_usage(
-        [self.gdb_tool, self.bash_tool], cur_round, response, prompt, extra_note)
+          [self.gdb_tool, self.bash_tool], cur_round, response, prompt,
+          extra_note)
 
     if self._parse_tag(response, 'conclusion'):
       if not self.gdb_tool_used:
         extra_note = 'NOTE: You MUST use the provided GDB tool to analyze the crash before providing a conclusion.'
         return self._container_handle_invalid_tool_usage(
-          [self.gdb_tool, self.bash_tool], cur_round, response, prompt, extra_note)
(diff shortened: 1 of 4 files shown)


AmPaschal:
/gcbrun exp -n pamusuo -m vertex_ai_gemini-2-5-pro-chat -ag -b analyzer-tests-1

DonggeLiu:
@maoyixie This is the PR : )

maoyixie:
@AmPaschal I apologize for the delayed feedback. Thank you for your detailed analysis of the root cause and the corresponding fix. I checked it out overall, no problem.

DonggeLiu (approved):

PR #1189, opened 2025-11-17
Fix docker version issue

AmPaschal:
This PR should fix the docker client version issue that occur when cloud build instances try to pull docker images.


DonggeLiu (approved):
Thanks!

